PR #3, opened 2017-02-27
Make the Dockerfile build glide-aware.

### LINKED ISSUE #2: Docker image build broken for hostpath example.
It fails with go-wrapper download errors, because the Dockerfile is not glide-aware.

dlorenc:
This should fix #2

### DIFF
--- a/docs/demo/hostpath-provisioner/Dockerfile
+++ b/docs/demo/hostpath-provisioner/Dockerfile
@@ -16,5 +16,5 @@ FROM golang:1.7.4
 COPY . /go/src/app
 WORKDIR /go/src/app
 RUN go get github.com/Masterminds/glide
-RUN glide install
+RUN glide install -v
 RUN go-wrapper install
--- a/docs/demo/hostpath-provisioner/Dockerfile
+++ b/docs/demo/hostpath-provisioner/Dockerfile
@@ -12,4 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM golang:1.7.4-onbuild
+FROM golang:1.7.4
+COPY . /go/src/app
+WORKDIR /go/src/app
+RUN go get github.com/Masterminds/glide
+RUN glide install
+RUN go-wrapper install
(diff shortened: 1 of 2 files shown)


dlorenc:
It might also be better to compile this statically and use a Makefile so the image can be built FROM SCRATCH. I can handle that too if you'd like.

docs/demo/hostpath-provisioner/Dockerfile
```diff
@@ -12,4 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM golang:1.7.4-onbuild
+FROM golang:1.7.4
+COPY . /go/src/app
+WORKDIR /go/src/app
+RUN go get github.com/Masterminds/glide
+RUN glide install
```
wongma7: does it work without '-v'? to remove nested vendor dirs

wongma7:
also can you update this https://github.com/kubernetes-incubator/external-storage/tree/master/docs/demo/hostpath-provisioner#building-and-running-our-hostpath-dynamic-provisioner so the Dockerfile block looks the same. 

I think FROM SCRATCH could be better BUT I just don't want the hostpath-provisioner binary thing (which is huge) lying around. But to be fair I think the containerized build is an even-huger Docker image soo...if you want to handle a FROM SCRATCH build as well I will accept it.

docs/demo/hostpath-provisioner/Dockerfile
dlorenc: It worked with and without it. I added the -v.

dlorenc:
> also can you update this https://github.com/kubernetes-incubator/external-storage/tree/master/docs/demo/hostpath-provisioner#building-and-running-our-hostpath-dynamic-provisioner so the Dockerfile block looks the same.
> 
Done.
> I think FROM SCRATCH could be better BUT I just don't want the hostpath-provisioner binary thing (which is huge) lying around. But to be fair I think the containerized build is an even-huger Docker image soo...if you want to handle a FROM SCRATCH build as well I will accept it.
Cool. I can add that in another PR.

wongma7:
ok Lgtm

PR #8, opened 2017-03-01
Add a Makefile for compiling a static demo container.

dlorenc:
The container is now 50mb instead of hundreds :)

To build: `make image`
```shell
$ docker images
hostpath-provisioner                    latest              sha256:2c323        About a minute ago   50.76 MB
```


wongma7:
thanks, this is great. can you please add the license boilerplate to the Makefile then run repo-infra/verify/verify-boilerplate.sh and that's all

wongma7:
merging, I'll get around to the boilerplate ;)

PR #12, opened 2017-03-03
Use shared top-level vendor directory

wongma7:
deleted nfs/trash.yaml and nfs/vendor

Then at top-level directory I did the following:
* glide init
* manually choose a new version of mitchellh/mapstructure because somebody uses an old version but e2e needs a new version
* glide install -v
* ~~glide-vc~~ glide-vc --use-lock-file


wongma7:
this **shouldn't** affect users of the library which is the main concern. Later today I'll test that by using hostpath-provisioner which should simulate a user experience

PR #16, opened 2017-03-06
Add efs provisioner

wongma7:


wongma7:
1 of the builds failed bc of dnf install hanging, merging

PR #17, opened 2017-03-06
add cephfs provisioner

rootfs:


wongma7:
@rootfs theres a top-level vendor now, I think it will "just work" if you try to build without your own vendor directory. since you dont depend on anything special, everything should be there

rootfs:
@wongma7 cool, rebased and removed vendor

cephfs/pod.yaml
```diff
@@ -0,0 +1,16 @@
+kind: Pod
+apiVersion: v1
+metadata:
+  name: hostpath-provisioner
```
wongma7: think this file should be deleted !

rootfs: done

Issue #4, opened 2017-02-27
Intermediary Status Hooks for External Provisioners Needed


sandaymin123:
Not all External Provisioners return immediately. Some provisoners *may* take several minutes or more to complete processing.  It is important that while a provisioner is processing the request, it can provide "intermediary status details" back to the volume controller  — by updating PVC status

A related issue is this:
https://github.com/kubernetes/kubernetes/issues/42121
@npramod  @xingzhou

xingzhou:
Let me take some investigation first.

wongma7:
For a short-term "fix" we can add a simple "External provisioner xyz has started provisioning for the claim." type of event to the provision controller right before it calls Provision.

That way, the user will see the "cannot find provisioner "ubiquity/flex", expecting that a volume for the claim is provisioned either manually or via external software" event and then immediately after that see "External provisioner xyz has started provisioning for the claim." which clarifies that it's the "via external software" case.

Eventually if provisioning fails, the claim will receive a ProvisioningFailed event and if it succeeds a ProvisioningSucceeds.

xingzhou:
@wongma7, thanks for the suggestions. I agree, I'm working on the new event code right now
to @sandaymin123, what do you think if we keep original event message so that probably user may see the following events in a PVC description:
```1. cannot find provisioner "ubiquity/flex", expecting that a volume for the claim is provisioned either manually or via external software```
```2. External provisioner xyz has started provisioning for the claim.```

xingzhou:
A sample PoC in my env, @sandaymin123 and @wongma7, FYI:
<img width="1279" alt="2017-03-06 4 48 35" src="https://cloud.githubusercontent.com/assets/5344440/23602906/af7357d0-028d-11e7-9516-308312341769.png">

sandaymin123:
Is it possible to just to have . External provisioner xyz has started provisioning for the claim If controller knows its working with an external provisioner..otherwise show the other message for other in-tree provisioiners?

xingzhou:
@sandaymin123, I think for PV controller, it is not aware whether or not a specific external storage provisioner is there, so in terms of event, it should be reported by external provisioner not the PV controller, is that right?

npramod:
@xingzhou @sandaymin123 In external provisioner - this is the extent of message change that is possible. If we want to modify the first message/event -- it will need to be done in PV_Controller in main k8s repo in addition to this change. @wongma7 thoughts?

wongma7:
[...]

PR #27, opened 2017-03-15
Added one more event for PVC when provisioning volume.

xingzhou:
Added one more event for PVC from external storage
provisioner when provisioning the volume.

fixed part of #4


wongma7:
lgtm

PR #32, opened 2017-03-16
relocate cephfs to ceph

rootfs:


wongma7:
there is now 1.py thats intentional right?

Also can you rename (here or later) kubernetes.io/cephfs to something-else/cephfs, somebody brought up yesterday that using kubernetes.io can make Delete fail

rootfs:
thanks, cleaned up now.

wongma7:
lgtm, don't forget to rename the provisioner from kubernetes.io later! Gluster is using gluster.org, idk what we can use here for ceph

rootfs:
yes, the name is renamed to `ceph`

PR #33, opened 2017-03-16
setgid on efs and nfs

wongma7:
EFS needs this. NFS not so much but it's nice to have.


wongma7:
note: EFS doesn't support setgid but will in the "near future"

PR #44, opened 2017-03-29
Flex provisioner example.

childsb:
Move the flex provisioner example from https://github.com/childsb/flex-provision


wongma7:
it won't build with the shared vendor dir after we merge it because they reshuffled client-go and apimachinery repos. Need to rebase on latest external-storage master and fix the errors. I can fix it after we merge

flex/deploy/docker/flex-provision.sh
```diff
@@ -0,0 +1,119 @@
+#!/bin/sh
+
```
wongma7: script is complaining this bash file needs boilerplate

flex/pkg/volume/provision_test.go
```diff
@@ -0,0 +1,639 @@
+/*
```
wongma7: this file should be deleted I think it has a lot of nfs provisioner specific things still

childsb:
I made the two updates you requested.  As far as the build.....

It works from me building clean--

```bash
mkdir -p /opt/go/src/github.com/kubernetes-incubator/
git clone git@github.com:childsb/external-storage.git
cd /opt/go/src/github.com/kubernetes-incubator/external-storage/flex
make
```

Where are you building that fails because of the shared library?

PR #50, opened 2017-03-30
Add scc to openshift example

andrewklau:
openshift needed the  `hostmount-anyuid` scc to start the pod


wongma7:
Thanks, lgtm. Will merge after cla sign :)

Issue #58, opened 2017-04-05
CephFS,Gluster Block: statefulness


wongma7:
@rootfs @humblec we have to implement a way for CephFS & Gluster Block provisioners to maintain state (across restarts) https://github.com/kubernetes-incubator/external-storage/blob/master/ceph/cephfs/cephfs-provisioner.go#L61 <- this old method won't survive restarts since it's generated at runtime. First you have to decide if your provisioner needs to maintain state at all? Like, whose PVs should it be allowed to Delete? If it restarts can it still Delete the PVs it Provisioned before it restarted, or will those PVs be left dangling?

To see what I am talking about, please see https://github.com/kubernetes-incubator/external-storage/tree/master/docs#running-multiple-provisioners-and-giving-provisioners-identities. In not so many words, I am basically talking about the same thing as the "unique name" in the CSI proposal https://docs.google.com/document/d/1JMNVNP-ZHz8cGlnqckOnpJmHF-DNY7IYP-Di7iuVhQI/edit#heading=h.6r2wd16r05z7 which allows a plugin to associate a given "storage space" with itself in case of failure. In the same way, our external provisioners have to be able to associate themselves with given "storage spaces."

In our case, Kubernetes won't decide a "unique name" for us, so we have to decide if we need one, and then if so generate & maintain it *ourselves.*

rootfs:
Yes that makes senses. I see you use an identify file, it is a good solution. I'll use a command line option to pass the identity.

humblec:
@wongma7 Indeed make sense. Thanks for the pointer 👍 . I will get it done for gluster block provisioner.

PR #59, opened 2017-04-05
use cmd option to pass provisioner identity

rootfs:


rootfs:
fix #58

ceph/cephfs/README.md
```diff
@@ -16,16 +16,16 @@ docker build -t cephfs-provisioner .
...
 ```bash
-ceph auth get client.admin 2>&1 |grep "key = " |a^C '{print  $3'} |xargs echo -n > /tmp/secret
+ceph auth get client.admin 2>&1 |grep "key = " |awk '{print  $3'} |xargs echo -n > /tmp/secret
 kubectl create secret generic ceph-secret-admin --from-file=/tmp/secret --namespace=kube-system
 ```
 
 * Start CephFS provisioner
 
-Assume kubeconfig is at `/root/.kube`:
+The following example uses `cephfs-provisioner-1` as the identity for the instance and assumes kubeconfig is at `/root/.kube`. If there are multiple provisioners, each should have a different identity.
```
wongma7: it's also needed if the pod gets restarted, otherwise the 'new' pod wont delete the 'old' pod's storage to be safe since the identities mismatch. But maybe that's okay, the user just needs to know that he is then responsible for deleting the 'old' pod's storage. Does ceph fs provisioner need to be stateful at all? Would somebody need to run more than one? 

(I am struggling still with how to communicate this stuff to users :) )

wongma7: e.g. the efs one is stateless because any "state" is based on what is on the efs share

rootfs: updated wording.

wongma7:
lgtm!

PR #73, opened 2017-04-13
Set umask to 0 to make sure the pvc directories have proper permissions

luksa:
Depending on the umask, the PVC directories permissions weren't always 777 and were not writable by certain containers.


wongma7:
lgtm

PR #75, opened 2017-04-17
Add `failedDeletionRetryThreshold` for external-storage controller.

### LINKED ISSUE #63: Use the "failedRetryThreshold" parameter in delete PVC flow
In the external-storage incubator code I see the `failedRetryThreshold` is used only for provision flow. 
Currently I see if delete PVC fails then it keeps retrying forever. If we have `failedRetryThreshold` apply for delete as well then we can control the retries.

xingzhou:
Instead of infinite volume deletion retry, this new option provides
the user the ability to configure the controller only to try limited
times when encounter failure during volume deletion.

fixed #63 

cc @wongma7, added some code to implement a failedDeletionRetryThreshold option, welcome your comments

### DIFF
--- a/lib/controller/controller.go
+++ b/lib/controller/controller.go
@@ -164,6 +164,14 @@ type ProvisionController struct {
 	// Map of failed claims
 	failedClaimsStats      map[types.UID]int
 	failedClaimsStatsMutex *sync.Mutex
+
+	// Threshold for max number of retries on failure of the deletion of provisioner. Non positive number means controller will retry the deletion infinitely.
+	failedDeletionRetryThreshold int
+
+	// Map of deletion failed volumes
+	failedDeletionVolumesStats map[types.UID]int
+
+	failedDeletionVolumesMutex *sync.Mutex
 }
 
 // NewProvisionController creates a new provision controller
@@ -212,6 +220,8 @@ func NewProvisionController(
 		failedClaimsStats:             make(map[types.UID]int),
 		failedRetryThreshold:          failedRetryThreshold,
 		failedClaimsStatsMutex:        &sync.Mutex{},
+		failedDeletionVolumesStats:    make(map[types.UID]int),
+		failedDeletionVolumesMutex:    &sync.Mutex{},
 	}
 
 	controller.claimSource = &cache.ListWatch{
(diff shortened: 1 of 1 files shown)


lib/controller/controller.go
```diff
@@ -297,6 +307,13 @@ func (ctrl *ProvisionController) Run(stopCh <-chan struct{}) {
 	<-stopCh
 }
 
+// Set the value of failedDeletionRetryThreshold
```
wongma7: linter says comment must begin with `SetFailedDeletionRetryThreshold`

BTW, do we want a `SetFailedProvisionRetryThreshold` too?

lib/controller/controller.go
```diff
@@ -164,6 +164,14 @@ type ProvisionController struct {
 	// Map of failed claims
 	failedClaimsStats      map[types.UID]int
```
wongma7: We should rename these variables to be consistent, don't need to do it now though.

lib/controller/controller.go
```diff
@@ -566,6 +595,27 @@ func (ctrl *ProvisionController) updateStats(claim *v1.PersistentVolumeClaim, er
 	}
 }
 
+func (ctrl *ProvisionController) updateDeletionVolumesStats(volume *v1.PersistentVolume, err error) {
```
wongma7: we can probably combine this with updateStats and make it take more parameters since thye look almost the same. But again, don't need to do it now/here.

lib/controller/controller.go
wongma7: ah okay, failedRetryThreshold for provision is an argument to NewProvisonController, so never mind about SetFailedProvisionRetryThreshold. IMO we should make failedDeletionRetryThreshold an arg too right after failedRetryThreshold. I am just a bit wary about how I keep adding args to NewProvisionController and breaking it :)

I think the best thing to do would be make NewProvisionController take a struct of parameters as an argument (just like upstream PV controller) instead of like 20, and increasing, different parameters. And document each of those parameters. I will do that soon and I guess it will be a breaking change.

xingzhou: yes, how about a new struct named `ControllerOptions`.

xingzhou: one more thing for failedDeletionRetryThresdhold, I suggest we keep the default behavior, which means infinitely retry here. So that provisioner can select their own choice to retry limited times or infinitely(periodically or exponentialBackoff)

lib/controller/controller.go
xingzhou: agree, I can do that later in an independent patch

lib/controller/controller.go
xingzhou: agree, I can refine it in a later PR

lib/controller/controller.go
xingzhou: Let me refine the comments and re-submit the patch

xingzhou:
@wongma7, have rebased and updated the patch according to your comments, PTAL, thanks! Besides, have some other ideas want to discuss with you, what's your slack id, please?

wongma7:
Lgtm. My slack is mawong I'm in #sig-storage

Issue #77, opened 2017-04-18
NewProvisionController needs refactoring


wongma7:
The parameters could be documented better, and we are adding new ones relatively often. @xingzhou I have started a little work on implementing this functional options pattern, wdyt? https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis . We are not dealing with a complex API here, just one function, but anything to make it more pleasant to use is good :).

xingzhou:
Thanks @wongma7 for creating the ticket. Something in my mind, for discussion:
1. Shall we keep the old ```NewProvisionController``` API for a while? As a external used project, I believe there should be quite some users started to use this controller in their own provisioner, if we can keep the old ```NewProvisionerController``` method, this may provide some time for the users to change their code.

2. For those options we want to pass into the construction of controller, do we support to change those options during controller's runtime? Because for some retry count related options, if we can dynamically change them during the controller runtime, I think it would provide some benefits for user's provisioner code.

Once the new constructor is ready, I think we can inform the community in the mailing list about the changes, and I can help provide the patches for local path/NFS provisioners to adopt the new constructor.

wongma7:
1. I am leaning toward doing this: release 2.1.0 soon with the deletion retry threshold changes and mark NewProvisionController deprecated. But then I don't think we need to wait long, we could bite the bullet and release 3.0.0 shortly afterward where the only change is the new, less ugly, NewProvisionController. Better to break it sooner than later IMO, then hopefully we can stay on 3.x.x for a longer while, truly respect semver, and not annoy consumers so much with the periodic changes to NewProvisionController's signature.
2. Yes, we could, I guess we have to decide which are changeable and which not. Thresholds being changeable make sense to me

Apologies in advance to anybody using the lib and annoyed by breaking changes to the API -- which is yes, only one function, but still annoying I'm sure :).

wongma7:
btw, I think I'll wait for client-go 3.0.0 before releasing our own 3.0.0. Client-go broke their API again, giving me another excuse to break ours in the next month or so.

PR #80, opened 2017-04-19
WIP: Use functional options pattern for NewProvisionController

wongma7:
fixes https://github.com/kubernetes-incubator/external-storage/issues/77

WIP, started writing this before the deletion threshold stuff was merged

### DIFF
--- a/lib/controller/controller.go
+++ b/lib/controller/controller.go
@@ -62,42 +62,10 @@ const annDynamicallyProvisioned = "pv.kubernetes.io/provisioned-by"
 
 const annStorageProvisioner = "volume.beta.kubernetes.io/storage-provisioner"
 
-// Number of retries when we create a PV object for a provisioned volume.
-const createProvisionedPVRetryCount = 5
-
-// Interval between retries when we create a PV object for a provisioned volume.
-const createProvisionedPVInterval = 10 * time.Second
-
-const (
-	// DefaultLeaseDuration is a suggested controller LeaseDuration:
-	// LeaseDuration is the duration that non-leader candidates will
-	// wait to force acquire leadership. This is measured against time of
-	// last observed ack.
-	DefaultLeaseDuration = 15 * time.Second
-	// DefaultRenewDeadline is a suggested controller RenewDeadline:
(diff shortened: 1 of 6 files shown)


wongma7:
@xingzhou please review if you'd like. The first commit is the actual changes to NewProvisionController that will affect you & other users. Second commit is just converting the provisioners in this repo to it.

Now, invocations will look something like this
```go
ctrl := controller.NewProvisionController(
  client,
  provisionerName,
  provisioner,
  serverGitVersion,
  controller.ResyncPeriod(60 * time.Second),
  controller.ExponentialBackOffOnError(false),
)
```
And I have set defaults for many parameters, since before it was mandatory to set them and now theya re optional.

lib/controller/controller.go
```diff
@@ -129,67 +97,131 @@ type ProvisionController struct {
...
+}
 
-	// Map of failed claims
-	failedClaimsStats      map[types.UID]int
-	failedClaimsStatsMutex *sync.Mutex
+// ExponentialBackOffOnError determines whether to exponentially back off from
+// failures of Provision and Delete. Defaults to true.
+func ExponentialBackOffOnError(exponentialBackOffOnError bool) func(*ProvisionController) error {
+	return func(c *ProvisionController) error {
+		c.runningOperations = goroutinemap.NewGoRoutineMap(exponentialBackOffOnError)
```
xingzhou: @wongma7, do you think it is good that we explicitly add a set of `SetXXX` method in the controller identifying that these attributes can be changed/updated during controller runtime. And for those can not modified ones, if the provisioner called these option methods again in controller runtime, the option method can detect and throw errors.

wongma7: yes, I'll add SetFailedProvisionThreshold and SetFailedDeleteThreshold. The other attributes however, I don't think can be changed during runtime. And I suppose I could error the options methods: say, after controller Run is called, a flag is set, and if that flag == true all the 'options' functions throw errors.

wongma7: It's so verbose I am considering not throwing errors if called in controller runtime at all. It should be understood that these are functional options and not something to call during runtime IMO, I will think baout it some more.

PR #105, opened 2017-05-12
Enable Chap Authentication in gluster block

humblec:


humblec:
@wongma7 PTAL :)

gluster/block/glusterblock-provisioner.go
```diff
@@ -167,6 +198,9 @@ func (p *glusterBlockProvisioner) Provision(options controller.VolumeOptions) (*
 					Lun:          0,
 					FSType:       "ext4",
 					ReadOnly:     false,
+					//chapAuthSession: true,
+					//SecretRef:       secretRef,
```
wongma7: is this ready for merge? Why create a secret if it will not be used

humblec: @wongma7 here is the story :). This api change happened with https://github.com/kubernetes/kubernetes/pull/43396 . However afaict, the kube libraries here are not updated to that.  Can we update the same ? or any thoughts?

wongma7: client-go 0ec10d4351f32832ab60d381d15e6d9e1f858bd3 and apimachinery 28d5d926c7c976d5885f31ace03d4a9b3ad8abd2 is what we need.

This means we can't use the client-go release-3.0 branch and so we can't use client-go 3.0.0 when that comes out, we'll have to track for a while. Not sure how I feel about it.

humblec:
@wongma7 Please review. :)

wongma7:
lgtm

Issue #101, opened 2017-05-11
defaultFailedProvisionThreshold   looks to be  really low.


humblec:
This threshold value has been set to ` 5` at present in the controller.  There may be scenarios where user still create the required artifacts after firing the pvc.  for eg# creating a secret or recreating a storageclass..etc.  As there are no more attempts made for the `claim` after this failure,  it will be bad user experience  for a user/developer. Can I increase the this value to more attempts or more time waiting ?

humblec:
@wongma7 please share your thought. I will send a patch accordingly.

wongma7:
sure, 5 is too low for exponentialbackoff to have any effect. 10 or something would be better

PR #106, opened 2017-05-12
At present, defaultFailedProvisionThreshold value in the controller

humblec:
is really low. There may be scenarios where user still create the
required artifacts after firing the pvc.
for eg# creating a secret or recreating a storageclass..etc.
As there are no more attempts made for the claim after this failure,
it will be bad user experience for a user/developer.
This patch lifts this threshold.
Issue#https://github.com/kubernetes-incubator/external-storage/issues/101


humblec:
@wongma7 PTAL.

wongma7:
lgtm

PR #110, opened 2017-05-16
Local storage static provisioner

msau42:
This provisioner is still in development and not functional yet.  It is pending API and volume plugin changes (PRs kubernetes/kubernetes#44640 and kubernetes/kubernetes#44897).  But the unit tests are passing and I did deploy it on a cluster to make sure it starts up and runs.

This PR includes:
- Basic controller
- PV cache and informer
- Volume discovery at a fixed directory
- PV deletion
- Deployment files
- Project documentation


local-volume/provisioner/pkg/deleter/deleter.go
```diff
@@ -0,0 +1,80 @@
...
+
+type deleter struct {
+	*types.RuntimeConfig
+}
+
+func NewDeleter(config *types.RuntimeConfig) Deleter {
+	return &deleter{RuntimeConfig: config}
+}
+
+func (d *deleter) DeletePVs() {
```
ddysher: Also, evict PV that doesn't belong to us? e.g. user manually creates a PV (malicious user maybe) with correct provisioned-by annotation, but random host path.

local-volume/provisioner/README.md
```diff
@@ -0,0 +1,53 @@
+# local-volume-provisioner
+
+NOTE: provisioner is still in development and is not functional yet
+
+local-volume-provisioner is an out-of-tree static provisioner for the Local volume plugin, which is a 1.7 alpha feature.
+
+It runs on each node in the cluster and monitors specified directories to look for new local mount points.  It then statically creates a Local PersistentVolume for each mount point.  It also monitors when the PersistentVolumes have been deleted, and will clean up the volume.
```
msau42: Yeah, I mean released, not deleted.  If it's deleted manually by the user, then we won't do any cleanup.

If the admin wants to delete the underlying storage, then I think the best way to do it would be:
1. Stop the pod that is using the volume
2. Admin removes the volume
3. Delete the PVC
4. Daemonset will try to cleanup, and it fails since the volume is no longer there.  But will still delete the API object.

local-volume/provisioner/cmd/main.go
```diff
@@ -0,0 +1,69 @@
...
+		HostDir:      "/mnt/disks",
+		MountDir:     "/local-disks",
+		DiscoveryMap: createDiscoveryMap(),
+	})
+}
+
+func createDiscoveryMap() map[string]string {
+	m := make(map[string]string)
+	// Default setting
+	m["local-storage"] = ""
```
msau42: This is for my ease of development on GCE environment until user configurable directories are added.

local-volume/provisioner/pkg/cache/populator.go
```diff
@@ -0,0 +1,119 @@
...
+	kcache "k8s.io/client-go/tools/cache"
+)
+
+type Populator struct {
+	client *kubernetes.Clientset
+	cache  *VolumeCache
+	name   string
+}
+
+func NewPopulator(client *kubernetes.Clientset, cache *VolumeCache, name string) *Populator {
```
msau42: Oops... yeah I also need to figure out a better interface for this.  I want to pass in RuntimeConfig, but it would create a circular dependency because it's in the cache package.

local-volume/provisioner/pkg/deleter/deleter.go
msau42: If an admin removes a mount point, I would prefer that they go through the standard workflow of PV deletion instead of adding a new method.

You need admin privileges to create PVs, so I don't think we need to handle the case of a malicious admin.

local-volume/provisioner/pkg/discovery/discovery.go
```diff
@@ -0,0 +1,116 @@
...
+		glog.Errorf("Error reading directory: %v\n", err)
+		return
+	}
+
+	for _, file := range files {
+		// Check if PV already exists for it
+		pvName := generatePVName(file, d.NodeName, class)
+		if !d.Cache.PVExists(pvName) {
+			// If not, create PV
+			// TODO: Validate that this path is file-based, and detect capacity
```
msau42: It can be a mount point or a directory.

local-volume/provisioner/pkg/discovery/discovery.go
```diff
@@ -0,0 +1,116 @@
...
+	fullPath := filepath.Join(d.MountDir, relativePath)
+	glog.Infof("Discovering volumes at mount path %q for storage class %q", fullPath, class)
+
+	files, err := d.VolUtil.ReadDir(fullPath)
+	if err != nil {
+		glog.Errorf("Error reading directory: %v\n", err)
+		return
+	}
+
+	for _, file := range files {
```
msau42: I can validate that the file is a directory.

local-volume/provisioner/pkg/discovery/discovery.go
```diff
@@ -0,0 +1,116 @@
...
+			},
+			PersistentVolumeSource: v1.PersistentVolumeSource{
+			/** TODO: api not merged yet
+			LocalVolume: &v1.LocalVolumeSource{
+				NodeName: config.NodeName,
+				Fs: v1.LocalFsVolume{Path: outsidePath},
+			},*/
+			},
+			AccessModes: []v1.PersistentVolumeAccessMode{
+				v1.ReadWriteOnce,
```
msau42: Yeah the current accessmodes is misleading and I'm not sure how to apply it to local volumes.  Technically, the local volume is only attachable to one node.  But because of the volume stickiness, pods using the same local volume will land on the same node and can share it without any problem.  

I need to look at the code that uses access mode to see how it's being used, and how that impacts local volume.

local-volume/provisioner/pkg/types/types.go
```diff
@@ -0,0 +1,50 @@
...
+const (
+	AnnProvisionedBy = "pv.kubernetes.io/provisioned-by"
+)
+
+type UserConfig struct {
+	// Name of the node this is running on
+	// TODO: Need to resolve nodename vs hostname
+	NodeName string
+	// The hostpath directory
+	HostDir string
```
msau42: I'm not sure I follow, can you give an example?  The DiscoveryMap is already a config of relative paths, so on the host you could have something like:

/mnt/local-volumes/class1
/mnt/local-volumes/class2

then your hostpath =  "/mnt/local-volumes", discoverymap = { "sc1": "class1", "sc2": "class2"}

local-volume/provisioner/pkg/discovery/discovery.go
wongma7: access mode is only used for binding PV and PVCs. Yes it may just be a bit misleading to the user if it's RWO but technically nothing is stopping them from using it as a "RWM" and using it from multiple pods

wongma7:
overall lgtm, ddysher has pointed out stuff (populator) and most everything else is being tracked as TODO already.

Feel free to add build & test invocations to travis (it takes forever at the moment since it builds everything in this repo 1 by 1, but I am working on splitting it up)

local-volume/provisioner/pkg/discovery/discovery.go
```diff
@@ -0,0 +1,116 @@
...
+func (d *Discoverer) createPV(file, relativePath, class string) {
+	pvName := generatePVName(file, d.NodeName, class)
+	outsidePath := filepath.Join(d.HostDir, relativePath, file)
+
+	glog.Infof("Found new volume at host path %v, creating LocalVolume PV %q\n", outsidePath, pvName)
+	pvSpec := &v1.PersistentVolume{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: pvName,
+			Annotations: map[string]string{
+				common.AnnProvisionedBy: d.Name,
```
msau42: Need to add topology constraints annotation with hostname label from the node.

local-volume/provisioner/README.md
ddysher: thanks, can you also update the doc.  The deletion flow basically works, but is not quite intuitive since admin will have to find the PVC bound to PV, and delete the PVC.  I think this is the right model, but just want to make sure this is clear to users.

local-volume/provisioner/pkg/discovery/discovery.go
ddysher: yeah, using RWO doesn't allow sharing volume (thus fsGroup and selinux problem is a fairly low priority issue :) ? 

What is the main concern for not using RWM here?  to be consistent with API semantics with other volume plugins?

local-volume/provisioner/pkg/types/types.go
ddysher: my point is we can support multiple hostdir, e.g.

/mnt/local-volumes/class1
/mnt/local-volumes/class2

/data/local-volumes/class3
/data/local-volumes/class4

so we have discoverymap = { "sc1": {hostpath: "/mnt/local-volumes/", relativePaths: []string{"class3"}}

just a rough idea, i think this can be deferred to dynamic setting.

local-volume/provisioner/README.md
msau42: I thought a little more about this situation.  I'm not sure if we want to automatically delete the PV object if the cleanup fails.  Because if the cleanup failed for some reason not due to path missing, then if we deleted the PV object, we would then automatically recycle a partially cleaned up PV that can be used by some other application.

So I think in the case of where cleanup fails, we should throw an error on the PV, and require admin manual intervention to cleanup.

wongma7: Agree, don't delete PV just leave it in failed state for admin to triage later.

This is basically like the existing Recycle reclaim policy. Not that you should necessarily be taking design cues from a deprecated feature, but it helps define what the existing flow for managing pvs from an admin perspective is like

local-volume/provisioner/pkg/discovery/discovery.go
msau42: I talked to Saad about how AccessModes work today.  Today, AccessModes is only used to pair PVCs with PVs with the same access modes.  The system does not have any special behavior of enforcing attaches or mounts based on access modes.  Today, it is the underlying storage (like gce pd) that does the "enforcing" of access mode (via attach error). There is a PR out that is looking to enforce attaching based on access modes in Kubernetes itself instead of the underlying storage.

However, there is no plan to support enforcing mounts based on access modes.  The reason for this is because users may have been relying on the current behavior of no mounting enforcement, and we should not break them by adding more strict enforcement.  If people really need mount enforcement, then we would have to provide it through a new field.  Another question is why do you need mount time enforcement though?  You will only share a volume if you specify the same PVC.  And PVCs are namespaced, so that means you are in control of all the pods sharing a PVC, so the enforcement would only be protecting you from yourself.

Considering all of this, AccessMode has no meaning for local volumes, because AccessMode only refers to attaching, and local volumes are already attached.  If we want to enforce some mounting access modes, then we need a new field in the PVC/PV.  It's a big user misconception that access modes refers to mounting, and maybe we need to update the documentation to clarify that.

local-volume/provisioner/pkg/types/types.go
msau42: Ok thanks for the clarification. I like your suggestion, let's handle it when we add in the configurable settings.

local-volume/provisioner/pkg/discovery/discovery.go
wongma7: +1. Makes so much more sense to think of accessmodes in terms of attachable vols only

RWO is the least bad description of local volumes 

Better documentation would be great. Confusion will be unavoidable because accessmodes is a first class spec field and it's just natural to expect it to *do* something other than act as a souped up label. (In its defence, PV label selectors did not exist yet)

local-volume/provisioner/README.md
ddysher: Yes, I thought the same.  I saw there is a TODO in deleter to surface the error.

msau42:
I added some more commits to address review comments.

local-volume/provisioner/pkg/discovery/discovery.go
ddysher: for local volume, i like the idea of having a new field in PVC/PV for mounting access mode. i think we lack the ability now to control how a local volume can be used, i.e. i want this volume to be shared across multiple pods, but this to be a dedicated local volume.

Using RWO is an unfortunate fact, let's make it clear in the documentation.  Ideally, i think this field should be empty since we don't depend on it eventually (scheduler integration doesn't require it?), but for alpha release, this is not optional, otherwise binding won't happen.

msau42: I think mounting access mode would need to be a generic field, for all volumes, not just local.  But again, why do we need to control the sharing of a local volume from within the same namespace?  As a user, you should be in full control of all the pods in your namespace, and there's nobody to protect from except yourself.

ddysher: yeah, mounting access mode should be generic, local volume is just an example in the last comment.

and I agree user should be in full control, but what's not clear to me right now is how they can express it in current api?  are you suggesting that if user want's to share a volume, then use the same claim in pods; otherwise, use a single claim in a single pod?

msau42: Yeah, if you want to share between multiple pods, then all those pods use the same PVC.  If you don't want to share, then they all have to have different PVCs.  And we don't have to worry about random pods using the same PVC because PVC is namespaced.

ddysher: cool, thx.  There is no need to handle it on pv/pvc layer.

wattsteve: Interesting discussion. ReadWriteOnce is about allowing only a Single Writer. We've never explicitly stated it, but this means a Non-Shared FileSystem on a Block Device (or when we have support for Raw Block it can mean that too).  ReadWriteMany is Multiple concurrent Writers, and this means Shared FileSystem.  The fact that we're articulating writers is part of the legacy that we've always historically associated a FileSystem with a PV. I would argue that if someone is putting EXT 3/4 or XFS on a Local Disk, then it would always be ReadWriteOnce. Are you anticipating a different Local Disk use case?

msau42: The number of writers is with respect to the node, not to the pods.  So with that definition, I agree, local disk is just RWO.

ddysher:
Generally LGTM, we can revisit this when API is finalized.

embarrassing...  how can I join the community?  @msau42 @wongma7 
> @msau42: GitHub didn't allow me to assign the following users: ddysher.

msau42:
This is the criteria to be added as a member: https://github.com/kubernetes/community/blob/master/community-membership.md.  I don't know if it automatically does it, or someone needs to add you manually.  If you meet the member criteria, let me know and I can get someone to add you.

ddysher:
Thanks, I'll take a look.

wongma7:
I am not a kubernetes org member, kubernetes-incubator is distinct. I think for kubernetes-incubator we just find somebody with admin privileges and have them add us to the org as external-storage collaborators

local-volume/provisioner/pkg/discovery/discovery.go
humblec: Yes, its abt number of nodes. However to mention, PVC can contain list/array of access modes.

local-volume/provisioner/pkg/discovery/discovery.go
humblec: @msau42 you can also add more annotations for its unique identity.

ddysher:
Thanks for the heads up; not a priority right now, i'll ping folks if this is necessary.
> I am not a kubernetes org member, kubernetes-incubator is distinct. I think for kubernetes-incubator we just find somebody with admin privileges and have them add us to the org as external-storage collaborators

local-volume/provisioner/pkg/discovery/discovery.go
```diff
@@ -0,0 +1,116 @@
...
+			/** TODO: api not merged yet
+			LocalVolume: &v1.LocalVolumeSource{
+				NodeName: config.NodeName,
+				Fs: v1.LocalFsVolume{Path: outsidePath},
+			},*/
+			},
+			AccessModes: []v1.PersistentVolumeAccessMode{
+				v1.ReadWriteOnce,
+			},
+			StorageClassName: "local-fast",
```
jsafrane: this should not be hardcoded

local-volume/provisioner/pkg/deleter/deleter.go
```diff
@@ -0,0 +1,80 @@
...
+	for _, pv := range d.Cache.ListPVs() {
+		if pv.Status.Phase == v1.VolumeReleased {
+			name := pv.Name
+			glog.Infof("Deleting PV: %v\n", name)
+
+			// Cleanup volume
+			err := d.cleanupPV(pv)
+			if err != nil {
+				// TODO: Log event on PV
+				glog.Errorf("Error reading directory: %v\n", err.Error())
```
jsafrane: "Error *clearing* directory"?

local-volume/provisioner/pkg/discovery/discovery.go
jsafrane: oops, this has been already fixed, sorry.

local-volume/provisioner/pkg/deleter/deleter.go
jsafrane: this has been already fixed, sorry.

jsafrane:
lgtm after a very quck review

local-volume/provisioner/pkg/discovery/discovery.go
msau42: Gluster can expose its volumes as RWX, so an application using a gluster volume can request rwx.  The gluster pod itself can specify using local PVs as rwo, but that detail is not exposed to the application.

local-volume/provisioner/pkg/controller/controller.go
```diff
@@ -0,0 +1,58 @@
...
+	glog.Info("Initializing volume cache\n")
+
+	runtimeConfig := &types.RuntimeConfig{
+		UserConfig: config,
+		Cache:      cache.NewVolumeCache(),
+		VolUtil:    util.NewVolumeUtil(),
+		APIUtil:    util.NewAPIUtil(client),
+		Client:     client,
+		// TODO: make this unique based on node name?
+		Name: "local-volume-provisioner",
```
msau42: Thanks for the suggestion! I will add node UUID to my next change.

local-volume/provisioner/pkg/deleter/deleter.go
```diff
@@ -0,0 +1,77 @@
...
+
+			// Remove API object
+			err = d.APIUtil.DeletePV(name)
+			if err != nil {
+				// TODO: Log event on PV
+				glog.Errorf("Error deleting PV %q: %v", name, err.Error())
+				continue
+			}
+
+			d.Cache.DeletePV(name)
```
msau42: The end effect is still the same in that the PV Is no longer in the cache.  But you do bring up some race conditions to watch out for, if the user deletes the PV in the middle of this cleanup.  Right now the delete and discovery are synchronous, but if we make it async in the future, it's something to watch out for.

local-volume/provisioner/pkg/discovery/discovery.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+	if !isDir {
+		return fmt.Errorf("Path is not a directory")
+	}
+	return nil
+}
+
+// TODO: maybe a better way would be to hash the 3 fields
+func generatePVName(file, node, class string) string {
+	return fmt.Sprintf("%v-%v-%v", class, node, file)
```
msau42: I agree.  I'm thinking of hashing the 3 fields to generate the id.  I need to be able to consistently regenerate the same ID so that the discovery can detect if it already exists.

Then the pv name could just be "local-pv-<id>", and doesn't contain redundant info about the node, storageclass and file name, which are all already in the details.

wongma7:
Last call for reviews, will merge shortly.

humblec:
lgtm after a very quck review

msau42:
Thanks guys!  I'll incorporate all the comments here into the next PR.  I just want to get a base change in so that others can start working on it too.

PR #119, opened 2017-05-19
Add RoundUpSize() to the Controller.

humblec:


lib/controller/controller.go
```diff
@@ -1085,3 +1085,12 @@ func (ctrl *ProvisionController) getStorageClassFields(name string) (string, map
 func claimToClaimKey(claim *v1.PersistentVolumeClaim) string {
 	return fmt.Sprintf("%s/%s", claim.Namespace, claim.Name)
 }
+
```
wongma7: IMO we should make a util folder and put it in there like lib/util/util.go. Then we can copy over what we need from upstream pkg/volume/util.go (until/unless we refactor upstream to include those in client-go)

humblec: @wongma7 I agree. Do you want to wait for util functions to create one. Otherwise I can do it with this PR . Please let me know what do you think ?

wongma7: with this PR, it's okay to have only RoundUpSize for now

humblec: @wongma7 Sure.. On it !

humblec: @wongma7 Updated :)

gluster/block/glusterblock-provisioner.go
```diff
@@ -257,14 +256,11 @@ func (p *glusterBlockProvisioner) createSecret(nameSpace string, secretName stri
...
-		volSize := p.options.PVC.Spec.Resources.Requests[v1.ResourceName(v1.ResourceStorage)]
-		volSizeBytes := volSize.Value()
-		sz := int(volume.RoundUpSize(volSizeBytes, 1024*1024*1024))
-	*/
-
-	volSizeBytes := "1073741824"
+
+	volSize := p.options.PVC.Spec.Resources.Requests[v1.ResourceName(v1.ResourceStorage)]
+	volSizeBytes := volSize.Value()
+	volszInt := int(controller.RoundUpSize(volSizeBytes, 1024*1024*1024))
```
wongma7: this should be util.RoundUpSize now (import util)

humblec: yeah.. was correcting it.

humblec: Done.

humblec:
@wongma7 PTAL :)

PR #120, opened 2017-05-20
Fix a regression with volume size calculation.

humblec:


humblec:
@wongma7 PTAL .:)

PR #122, opened 2017-05-22
Add local volume controller alpha workaround

ddysher:
This controller basically acts as a proof-of-concept for how we can solve the problem outlined in README.  I can add more tests and docs if we think this controller actually solves it.

NOTE this currently won't build as I'm having a hard time managing vendor (missing k8s.io/client-go/listers); since it is essentially shared with multiple sub-project.  @wongma7 any suggestions on updating vendors?

I've tested locally in a new repo (to bypass vendor problem)

@msau42

### DIFF
--- a/local-volume/alpha-controller/Makefile
+++ b/local-volume/alpha-controller/Makefile
@@ -1,0 +1,35 @@
+# Copyright 2017 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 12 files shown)


ddysher:
@msau42 also, how much effort do we want to put into this, I think this is pretty short-lived code?

wongma7:
@ddysher I think the reason `listers` is missing is because we run "glide-vc --use-lock-file" to keep the repo size down, documented briefly here.https://github.com/kubernetes-incubator/external-storage/blob/master/docs/README.md#contributing.

So please try running `glide install -v` then `glide-vc --use-lock-file` again to pick up the new packages being used by your code.

Sorry for the troubles :)

ddysher:
@wongma7 thx for the tips!

I'll update if folks think this controller is necessary for alpha release.  There are many higher priority tasks.

discordianfish:
Can this be closed? Sounds like local-volume was added somewhere else..?

msau42:
Correct this is no longer needed now that Kubernetes scheduler is integrated with PV binding

PR #128, opened 2017-05-23
Keep different dictionaries for gluster block mode and heketi mode.

humblec:
Parse opmode args to these dictionaries accordingly.

### DIFF
--- a/gluster/block/cmd/glusterblock-provisioner/glusterblock-provisioner.go
+++ b/gluster/block/cmd/glusterblock-provisioner/glusterblock-provisioner.go
@@ -396,6 +396,7 @@ func parseClassParameters(params map[string]string, kubeclient kubernetes.Interf
 	authEnabled := true
 	chapAuthEnabled := true
 	parseOpmode := ""
+	blkmodeArgs := ""
 	haCount := 3
 
 	for k, v := range params {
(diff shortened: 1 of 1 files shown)


gluster/block/cmd/glusterblock-provisioner/glusterblock-provisioner.go
```diff
@@ -432,65 +435,72 @@ func parseClassParameters(params map[string]string, kubeclient kubernetes.Interf
...
+				if err != nil {
+					return nil, err
+				}
+			} else {
+				return nil, fmt.Errorf("StorageClass for provisioner %q must have secretNamespace and secretName either both set or both empty", "glusterblock")
+
 			}
 		} else {
-			return nil, fmt.Errorf("StorageClass for provisioner %q must have secretNamespace and secretName either both set or both empty", "glusterblock")
+			cfg.secretValue = cfg.userKey
```
wongma7: what if userKey is blank?

IMO it would be clearer if it did all the validation of params beforehand in a different function instead of waiting until they are needed.

wongma7:
lgtm. docs will need update, which I assume is coming later ;)

humblec:
Sure.. Before sending next PRs, I will send README update to an extent. :)

PR #135, opened 2017-05-24
Detect local volume capacity

ddysher:
Detect local volume capacity, a couple of implementation notes:
- to correctly detect host devices, daemonset is changed to run in privileged mode
- after investigating cadvisor, i think we can just detect capacity using statfs, instead of collecting all information from device name, device number, mountpoint, etc; we don't need that much data
- statfs returns 'total, free, available'.  We use 'total - free + available' as pv [capacity](https://github.com/kubernetes-incubator/external-storage/pull/135#discussion_r119220717)
- if user just create regular directory, then the available space for the mount device is returned, in which case the behavior is similar to hostpath

Future work
- zfs and devicemapper might require [special handling](https://github.com/google/cadvisor/blob/86daf2b4db38364f86aca6af38b28f4844885872/fs/fs.go#L342)
- hostpath mount option should be [rshared](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/propagation.md)

@msau42 @jsafrane @humblec @wongma7

### DIFF
--- a/local-volume/provisioner/pkg/util/volume_util.go
+++ b/local-volume/provisioner/pkg/util/volume_util.go
@@ -21,6 +21,8 @@ import (
 	"os"
 	"path/filepath"
 
+	"golang.org/x/sys/unix"
+
 	"github.com/golang/glog"
 )
 
@@ -34,6 +36,9 @@ type VolumeUtil interface {
 
 	// Delete all the contents under the given path, but not the path itself
 	DeleteContents(fullPath string) error
+
+	// Get available capacity for fs on full path
+	GetFsAvailableByte(fullPath string) (uint64, error)
 }
 
 var _ VolumeUtil = &volumeUtil{}
@@ -99,6 +104,18 @@ func (u *volumeUtil) DeleteContents(fullPath string) error {
 	return nil
 }
 
+// GetFsAvailableByte returns available capacity in byte about a mounted filesystem.
+// fullPath is the pathname of any file within the mounted filesystem. Capacity
+// returned here is total capacity available to non-root users, and does not include
+// fs reserved space.
+func (u *volumeUtil) GetFsAvailableByte(fullPath string) (uint64, error) {
+	var s unix.Statfs_t
+	if err := unix.Statfs(fullPath, &s); err != nil {
+		return 0, err
+	}
+	return uint64(s.Frsize) * (s.Blocks - s.Bfree + s.Bavail), nil
+}
+
 var _ VolumeUtil = &FakeVolumeUtil{}
 
 // FakeVolumeUtil is a stub interface for unit testing
@@ -114,7 +131,8 @@ type FakeFile struct {
 	Name     string
 	IsNotDir bool
 	// Expected hash value of the PV name
-	Hash uint32
+	Hash     uint32
+	Capacity uint64
 }
 
 // NewFakeVolumeUtil returns a VolumeUtil object for use in unit testing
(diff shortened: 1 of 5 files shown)


local-volume/provisioner/pkg/util/volume_util.go
```diff
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"syscall"
```
msau42: I think this pkg is deprecated, and we're supposed to use https://godoc.org/golang.org/x/sys/unix instead.

local-volume/provisioner/pkg/util/volume_util.go
```diff
@@ -94,6 +98,16 @@ func (u *volumeUtil) DeleteContents(fullPath string) error {
...
 }
 
+// GetFsAvailable returns available capacity about a mounted filesystem. fullPath
+// is the pathname of any file within the mounted filesystem.
+func (u *volumeUtil) GetFsAvailable(fullPath string) (uint64, error) {
+	var s syscall.Statfs_t
+	if err := syscall.Statfs(fullPath, &s); err != nil {
+		return 0, err
+	}
+	return uint64(s.Frsize) * s.Bavail, nil
```
msau42: What's the difference between frsize and bsize?

Do we want to use total blocks instead of available?  I'm thinking of a situation where they might want pre-populated data in the volume. 

Also in the shared directory case, we could get inconsistent capacities depending on the order that pods fill them up (although the capacity is already going to be inaccurate in that case).

ddysher: I found this [article](http://docs.oracle.com/cd/E19253-01/817-5093/fsfilesysappx-10/index.html) which explains this pretty well (the concept applies to most fs).

What is not quite clear is how this was handled in all sorts of filesystem; but it looks like the values are usually the same (the option to create different fragment size is actually removes the option from mke2fs, though man page still has the option).

Using Frsize here is mainly based on cadvisor https://github.com/google/cadvisor/blob/aee77359083ccef523d0299316e831f1aed712bd/fs/fs.go#L530

Also, df uses frsize as welll (fall back to bsize if it is not available): https://github.com/coreutils/gnulib/blob/2a0c08e48834ae056964922ee6d21ccd223d1d80/lib/fsusage.c#L131

ddysher: @msau42 I'm ok with using total blocks for reasons listed in ur comment (pre-populated data and consistent capacity).  My concern is that the total block is generally not the capacity user can actually use, e.g. for ext4, there's 5% system reserved space.

msau42: Hm is there a field that can distinguish between system reserved, and user used?

ddysher: There is no such a field that i know of.  Looking further into df, I think what we can do is to use 'used + available'.

related logic from df:
```c
  bv->total = fsu->fsu_blocks;
  bv->available_to_root = fsu->fsu_bfree;

  bv->used = bv->total - fsu->fsu_bfree;

  uintmax_t nonroot_total = v->used + v->available;
```

so simply put, we can return `s.Blocks - s.Bfree + s.Bavail`.  This will return total capacity for non root user, removed system reserved altogether.

WDYT

ddysher:
@msau42 comment addressed

ddysher:
Also, to correctly detect capacity, we need mount propagation for hostpath
https://github.com/kubernetes/community/blob/master/contributors/design-proposals/propagation.md

local-volume/provisioner/pkg/util/volume_util.go
```diff
@@ -98,14 +99,14 @@ func (u *volumeUtil) DeleteContents(fullPath string) error {
 	return nil
 }
 
-// GetFsAvailable returns available capacity about a mounted filesystem. fullPath
-// is the pathname of any file within the mounted filesystem.
-func (u *volumeUtil) GetFsAvailable(fullPath string) (uint64, error) {
-	var s syscall.Statfs_t
-	if err := syscall.Statfs(fullPath, &s); err != nil {
+// GetFsAvailableByte returns available capacity in byte about a mounted filesystem.
+// fullPath is the pathname of any file within the mounted filesystem.
```
msau42: Can you also add in this comment that this does not include fs reserved space?

msau42:
Remind me to add the limitation about mount propagation in the docs.

msau42:
@jsafrane do you know the status for the mount propagation?

ddysher:
@msau42 talked to @jsafrane in another issue, this misses 1.7

https://github.com/kubernetes/community/pull/659

ddysher:
@msau42 rebase to HEAD, PTAL

PR #141, opened 2017-05-30
 Manually patch in local storage API

wongma7:
note to self tomrrow: may need to update travis to use go1.8


wongma7:
versioned helpers (v1.GetPersistentVolumeClaimClass) are gone for whatever reason in latest client-go.

wongma7:
merging, will fix tests after

PR #144, opened 2017-06-01
Make more readable error strings.

humblec:

### DIFF
--- a/gluster/block/cmd/glusterblock-provisioner/glusterblock-provisioner.go
+++ b/gluster/block/cmd/glusterblock-provisioner/glusterblock-provisioner.go
@@ -270,7 +270,7 @@ func (p *glusterBlockProvisioner) createVolume(volSizeInt int, blockVol string) 
 	sizeStr := strconv.Itoa(volSizeInt)
 	haCountStr := strconv.Itoa(p.provConfig.haCount)
 
-	glog.V(2).Infof("glusterfs: create block volume of size: %d  and configuration %+v", volSizeInt, p.provConfig)
+	glog.V(2).Infof("glusterblock: create block volume of size: %d  and configuration %+v", volSizeInt, p.provConfig)
 
 	// Possible opModes are gluster-block and heketi:
 	switch p.provConfig.opMode {
@@ -304,7 +304,7 @@ func (p *glusterBlockProvisioner) createVolume(volSizeInt int, blockVol string) 
 			json.Unmarshal([]byte(out), blockRes)
 
 			if blockRes.User == "" || blockRes.AuthKey == "" {
-				return nil, fmt.Errorf("gluster-block: missing CHAP - invalid volume creation ")
+				return nil, fmt.Errorf("glusterblock: missing CHAP - invalid volume creation ")
 			}
 
 		}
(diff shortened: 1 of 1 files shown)


humblec:
@wongma7 please review. I am yet to check the travis failure is spurious.

gluster/block/cmd/glusterblock-provisioner/glusterblock-provisioner.go
```diff
@@ -363,27 +367,31 @@ func (p *glusterBlockProvisioner) Delete(volume *v1.PersistentVolume) error {
...
 	case "heketi":
-		glog.V(1).Infof("heketik: Deleteing Volume %v", delBlockVolName)
+		glog.V(1).Infof("glusterblock: opmode[heketi]: deleting Volume %v", delBlockVolName)
 	default:
-		glog.Errorf("Unknown OpMode, failed to delete volume %v", delBlockVolName)
+		glog.Errorf("glusterblock: Unknown OpMode, failed to delete volume %v", delBlockVolName)
 
 	}
 
+	deleteSecErr := p.client.Core().Secrets(p.volConfig.BlockSecretNs).Delete(p.volConfig.BlockSecret, nil)
```
wongma7: Won't other PVs rely on the secret? I thought it was one secret per Storage Class

gluster/block/cmd/glusterblock-provisioner/glusterblock-provisioner.go
```diff
@@ -245,7 +247,9 @@ func (p *glusterBlockProvisioner) createSecretRef(nameSpace string, secretName s
 		Type: chapType,
 	}
 
+	p.volConfig.BlockSecret = secretName
 	secretRef := &v1.LocalObjectReference{}
+	p.volConfig.BlockSecretNs = nameSpace
 	if secret != nil {
 		_, err = p.client.Core().Secrets(nameSpace).Create(secret)
```
wongma7: What if the secret already exists? Isn't there a very high chance of that if Provisioning multiple PVs? I was under the impression they all use the same secret (storage class immutable)

wongma7: Correction: I mean if provisioning pvs for multiple pvcs *in the same namespace*. Same idea though.

gluster/block/cmd/glusterblock-provisioner/glusterblock-provisioner.go
wongma7: one secret per Storage Class per namespace *

wongma7:
the travis failure is nfs's fault, my 'everything-else' suite passed.

gluster/block/cmd/glusterblock-provisioner/glusterblock-provisioner.go
humblec: @wongma7 its per PV and it holds the user credentials of that block volume. At time of PV deletion it should go away.

gluster/block/cmd/glusterblock-provisioner/glusterblock-provisioner.go
humblec: @wongma7  Actually secrets are created with a unique identifier, that said, the `user` get a unique ID from gluster block.  so I believe we are safe here.

wongma7:
Secret name is not unique though if multiple pvs can have the same user. Is
that possible?

humblec:
@wongma7 the user field is filled with a value called 'GBID' which is unique identifier generated by the gluster block internally.

wongma7:
Ok, I must have confused cfg.user (class param) with vol.user (from response).

Lgtm

PR #146, opened 2017-06-02
Functional local storage provisioner

msau42:
This PR makes the local storage provisioner functional, and also addresses the comments from the last review.  Changes include:
* Using the new local volume API
* Using the new storage node affinity annotation
* Make each provisioner use the node UUID in its name
* Generate FNV hash as the PV name
* Only have informer update the cache

### DIFF
--- a/local-volume/provisioner/TODO.md
+++ b/local-volume/provisioner/TODO.md
@@ -14,3 +14,5 @@
 ## P2
 * Partitioning, formatting, and mount extensions (needs mount propagation)
 * Block device support (needs API and volume plugin changes too)
+* Refactor to just use informer's cache (and need to stub out API calls for unit
+  testing)
--- a/local-volume/provisioner/TODO.md
+++ b/local-volume/provisioner/TODO.md
@@ -1,19 +1,16 @@
 # TODO
 
 ## P0
-* Update with Local volume API (msau42)
-* Give each provisioner a unique name (msau42)
 * Detect capacity of mount points (ddysher)
 * Deploy to public image repo (msau42)
 * E2E tests
 
 ## P1
+* Can we filter the PV watch to only certain fields?
 * Investigate nodename vs hostname issue (msau42)
-* Investigate better PV naming scheme - hashing?
 * PV events on deletion failure
 * Configmap for user parameters (ddysher)
 
 ## P2
 * Partitioning, formatting, and mount extensions (needs mount propagation)
 * Block device support (needs API and volume plugin changes too)
-* Independent sync loops for discovery and deleter.
--- a/local-volume/provisioner/deployment/kubernetes/admin_account.yaml
+++ b/local-volume/provisioner/deployment/kubernetes/admin_account.yaml
@@ -3,7 +3,7 @@ kind: ServiceAccount
 metadata:
   name: local-storage-admin
 ---
-apiVersion: rbac.authorization.k8s.io/v1alpha1
+apiVersion: rbac.authorization.k8s.io/v1beta1
 kind: ClusterRoleBinding
 metadata:
   name: local-storage-provisioner-pv-binding
@@ -17,7 +17,7 @@ roleRef:
   name: system:persistent-volume-provisioner
   apiGroup: rbac.authorization.k8s.io
 ---
-apiVersion: rbac.authorization.k8s.io/v1alpha1
+apiVersion: rbac.authorization.k8s.io/v1beta1
 kind: ClusterRoleBinding
 metadata:
   name: local-storage-provisioner-node-binding
--- a/local-volume/provisioner/deployment/kubernetes/admin_account.yaml
+++ b/local-volume/provisioner/deployment/kubernetes/admin_account.yaml
@@ -6,7 +6,7 @@ metadata:
 apiVersion: rbac.authorization.k8s.io/v1alpha1
 kind: ClusterRoleBinding
 metadata:
-  name: local-storage-provisioner
+  name: local-storage-provisioner-pv-binding
   namespace: default
 subjects:
 - kind: ServiceAccount
@@ -15,4 +15,18 @@ subjects:
 roleRef:
   kind: ClusterRole
   name: system:persistent-volume-provisioner
+  apiGroup: rbac.authorization.k8s.io
+---
+apiVersion: rbac.authorization.k8s.io/v1alpha1
+kind: ClusterRoleBinding
+metadata:
+  name: local-storage-provisioner-node-binding
+  namespace: default
+subjects:
+- kind: ServiceAccount
+  name: local-storage-admin
+  namespace: default
+roleRef:
+  kind: ClusterRole
+  name: system:node
   apiGroup: rbac.authorization.k8s.io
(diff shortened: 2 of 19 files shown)


local-volume/provisioner/deployment/kubernetes/admin_account.yaml
```diff
@@ -16,3 +16,17 @@ roleRef:
   kind: ClusterRole
   name: system:persistent-volume-provisioner
   apiGroup: rbac.authorization.k8s.io
+---
+apiVersion: rbac.authorization.k8s.io/v1alpha1
```
ddysher: should we use v1beta1?

local-volume/provisioner/pkg/deleter/deleter.go
```diff
@@ -17,28 +17,35 @@ limitations under the License.
...
 	"k8s.io/client-go/pkg/api/v1"
 )
 
-type Deleter interface {
-	// Cleanup and delete all its owned PVs that have been released
-	DeletePVs()
-}
+const (
+	pollInterval = 1 * time.Second
+	// TODO: is this too fast?
```
ddysher: If `waitForDeleteTimeout` passed and pv still exists, then we'll just retry deleting, so this doesn't matter too much?

local-volume/provisioner/pkg/discovery/discovery.go
```diff
@@ -80,49 +119,36 @@ func (d *Discoverer) validateFile(fullPath string) error {
...
+		Name:            pvName,
+		HostPath:        outsidePath,
+		StorageClass:    class,
+		ProvisionerName: d.Name,
+		AffinityAnn:     d.nodeAffinityAnn,
+	})
+
+	// Add to cache first to handle race condition between discoverer and informer updating the same pv in the cache
+	// This way the ordering is consistent. First discoverer adds to cache, then informer will update the pv object
+	d.Cache.AddPV(pvSpec)
```
ddysher: Can u elaborate more on this race?  My understanding is that if pv is not added here, then it will be added when we create the pv in kubernetes and updated via informer; but if it takes too long for informer to update the cache, then we'll end up create the pv again.

local-volume/provisioner/pkg/deleter/deleter.go
msau42: I think because deletion and creation occur serially right now, this is fine.  There is no risk of the informer updating the cache in the middle of the second deletion attempt , and the creation happening in the middle.

Maybe in that case, I should not even wait for it to be deleted from the cache and let it delete twice if it takes too long the first time.  The main controller already sleeps for a few seconds between loops.

local-volume/provisioner/pkg/discovery/discovery.go
msau42: Yeah, the idea was to immediately add it to the cache after creating, so that we don't try to create the PV twice.  But I suppose there is no harm in trying to create twice, since we don't do any sort of formatting or destructive action here.  I can remove the cache add here and just let the informer do it.

local-volume/provisioner/pkg/common/common.go
```diff
@@ -14,22 +14,27 @@ See the License for the specific language governing permissions and
...
+
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/pkg/api/v1"
 )
 
 const (
 	AnnProvisionedBy = "pv.kubernetes.io/provisioned-by"
+	// TODO: is this the correct key?
```
ddysher: Yes, I agree with ur understanding, and internally, we always use `--hostname-override` to have consistent hostname seen by kubernetes.

This is unfortunate but do people already use hostname for node affinity?  If so, then we can just conform to that.

local-volume/provisioner/pkg/controller/controller.go
```diff
@@ -17,36 +17,41 @@ limitations under the License.
...
-	runtimeConfig := &types.RuntimeConfig{
+	runtimeConfig := &common.RuntimeConfig{
 		UserConfig: config,
 		Cache:      cache.NewVolumeCache(),
 		VolUtil:    util.NewVolumeUtil(),
 		APIUtil:    util.NewAPIUtil(client),
 		Client:     client,
-		// TODO: make this unique based on node name?
-		Name: "local-volume-provisioner",
+		Name:       fmt.Sprintf("local-volume-provisioner-%v", config.Node.UID),
```
ddysher: Can u create a node with the same name; wouldn't kubernetes complain name already exist?

If mount protection is only hostname, then at least using hostname here is not a degradation wrt protection.  Using UID is less consistent and less convenient for users if user want to inspect provisioner.  but I don't have a strong opinion.

local-volume/provisioner/pkg/discovery/discovery.go
ddysher: ok, removing this sounds good.

local-volume/provisioner/pkg/deleter/deleter.go
ddysher: The problem sounds similar to creating the volume twice.  Since controller will actively sync the state, we'll eventually reach the consistent point, so waiting doesn't seem to be strictly required?

local-volume/provisioner/pkg/controller/controller.go
msau42: I believe it should check if node name is the same when registering.  So the only case I can think of is if you delete the node, and then create a new node with the same name.  Here is one scenario:

1. On old node, have PV-a using path1
2. Delete old node (without deleting PV-a)
3. Create new node with same name, create PV-b using path1 and start using it
4. Delete PV-a, now provisioner will delete contents from the new node's path1, which is PV-b.

I think it would be better to protect data for as many cases as we can, even if we cannot protect from all situations.  Maybe for debugging/troubleshooting, the provisioner name can include the human readable name and the uid?

I will also document this limitation in the alpha user doc.

local-volume/provisioner/TODO.md
```diff
@@ -1,19 +1,16 @@
...
 
 ## P0
-* Update with Local volume API (msau42)
-* Give each provisioner a unique name (msau42)
 * Detect capacity of mount points (ddysher)
 * Deploy to public image repo (msau42)
 * E2E tests
 
 ## P1
+* Can we filter the PV watch to only certain fields?
```
msau42: Ok thanks, from my initial search, it seems like the use of fieldSelector is very limited (not generic).  I will look into this as an enhancement for later.

local-volume/provisioner/pkg/controller/controller.go
ddysher: ok, we can both name and uid

msau42:
@jsafrane @wongma7 @ddysher @humblec PTAL when you get the chance, especially the discussion about using node UID and the situation where you could recreate a node (with different volumes) with the same name.

wongma7:
lgtm. +1 to "Remove cache updates from discoverer and deleter", that simplifies things a lot.

jsafrane:
nothing caught my eyes after a quick review, lgtm

wongma7:
Will merge shortly, last call for reviews...

ddysher:
lgtm

PR #154, opened 2017-06-10
Add test to verify storageclass

ddysher:
Based on https://github.com/kubernetes-incubator/external-storage/pull/135.  Review the last commit will suffice.

### DIFF
--- a/local-volume/provisioner/pkg/util/volume_util.go
+++ b/local-volume/provisioner/pkg/util/volume_util.go
@@ -21,6 +21,8 @@ import (
 	"os"
 	"path/filepath"
 
+	"golang.org/x/sys/unix"
+
 	"github.com/golang/glog"
 )
 
@@ -34,6 +36,9 @@ type VolumeUtil interface {
 
 	// Delete all the contents under the given path, but not the path itself
 	DeleteContents(fullPath string) error
+
+	// Get available capacity for fs on full path
+	GetFsAvailableByte(fullPath string) (uint64, error)
 }
 
 var _ VolumeUtil = &volumeUtil{}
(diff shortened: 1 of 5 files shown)


ddysher:
@msau42

local-volume/provisioner/pkg/util/volume_util.go
```diff
@@ -99,6 +104,18 @@ func (u *volumeUtil) DeleteContents(fullPath string) error {
 	return nil
 }
 
+// GetFsAvailableByte returns available capacity in byte about a mounted filesystem.
+// fullPath is the pathname of any file within the mounted filesystem. Capacity
+// returned here is total capacity available to non-root users, and does not include
+// fs reserved space.
+func (u *volumeUtil) GetFsAvailableByte(fullPath string) (uint64, error) {
```
ianchakeres: How is this AvailableByte different from available bytes inside kubernetes/pkg/volume/util/fs.go:33 FsInfo?

ddysher: Thanks for pointing this out.  I didn't know there is such a util in kubernetes :)  You can take a look at discussion here: https://github.com/kubernetes-incubator/external-storage/pull/135#discussion_r119220717   Basically, available from that package returns all available spaces block wise, but there are certain amount of blocks can't be used by normal user. AvailableByte removes those.

ianchakeres: @ddysher @msau42 would it make sense to modify the volume util package to match this definition of availablebytes?

msau42: Looking at FsInfo, I think it returns all the info that you need here.  You could do (usage + available) to get the equivalent.

ddysher: yeah, `usage + available` is basically the idea.

@ianchakeres available in fs.go means remaining bytes available to be consumed, while available here means capacity that can be used by normal users. so we can't change the definition of available in fs.go.  It may worth taking a look if returning a new field for availablebytes is necessary though, but i don't know if there is use case. Even if there is, it can be calculated by just using usage+available. 

But you raise another IMPORTANT problem, that we should change to a better name......

PR #157, opened 2017-06-12
local-volume README updates

jsafrane:
- fix a typo in daemonset file name
- explicitly write that there is nothing to configure at this point
- add an example how to test with `hack/local-up-cluster.sh` that everybody loves


msau42:
Thanks for adding this!

PR #162, opened 2017-06-13
Enable repo verify-go-src & local-volume build+test in CI

wongma7:
At the moment, NFS build+integration+e2e tests takes about 12 minutes, so should keep all "suites" under that time. The NFS build can definitely be shortened (in the future) though.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ go:
 services: docker
 
 env:
-# - TEST_SUITE=local-volume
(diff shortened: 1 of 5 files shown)


wongma7:
@msau42 made 3 tiny comments edits in local-volume to stop golint warnings. repo-wide verification (make verify) is now enabled--let's try it out, if it gets too oppressive I have no issue dialing it down...

PR #166, opened 2017-06-14
Fix gofmt warning

ddysher:
@wongma7  PTAL


ddysher:
`./repo-infra/verify/go-tools/verify-gofmt.sh` is unhappy about this verbose style for struct array.

PR #168, opened 2017-06-14
Update nfs-provisioner to NFS Ganesha 2.5

wongma7:
~Includes from what I can tell many bugfixes, lets us not have to build ganesha from source in Dockerfile (yay!). Only cost is that I can't set some CMAKE variables for where to store the NFSv4 recovery directories, so I just do some symlinking instead.~

~reliance on fedora:rawhide should be temporary; I will wait for ganesha 2.5 to show up in fedora 26 (and the docker fedora:26 image to show up) before releasing nfs-provisioner v1.0.9~


ekozan:
any news on this one ? :D

wongma7:
Unfortunately it's not stable, the server randomly stops working on restart sometimes. Will try to debug but it's not high priority

clkao:
[latest 2.5.x](https://github.com/nfs-ganesha/nfs-ganesha/commits/V2.5-stable) seems to have lots of fixes. Perhaps it's worth reevaluating.

Is there any particular set of tests that people should run to validate storage providers?

wongma7:
will investigate this someday. for now, remember nfs-provisioner is not recommended for production use, maybe just to test dynamic provisioning on your cluster!

PR #170, opened 2017-06-14
Deploy local storage provisioner to public repo

msau42:
* Updated the Makefile to push to the public repo in quay.io
* Updated the TODO and user guide
* Fix example pvc spec

### DIFF
--- a/local-volume/provisioner/Makefile
+++ b/local-volume/provisioner/Makefile
@@ -1,4 +1,4 @@
-# Copyright 2017 The Kubernetes Authors.
+# Copyright 2016 The Kubernetes Authors.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


msau42:
/assign @wongma7 

cc @ddysher

local-volume/provisioner/Makefile
```diff
@@ -37,18 +28,26 @@ else
 endif
```
wongma7: we can remove this block, it won't work anymore with every provisioner getting its own set of tags. Here's what I do now: https://github.com/kubernetes-incubator/external-storage/blob/master/gluster/block/Makefile#L18

local-volume/provisioner/Makefile
```diff
@@ -37,18 +28,26 @@ else
...
+.PHONY: all build
+
+container: build quick-container
+.PHONY: container
 
-container: build
+quick-container:
 	cp local-volume-provisioner deployment/docker
-	docker build --pull -t $(IMAGE):$(VERSION) deployment/docker
+	docker build -t $(IMAGE):$(VERSION) deployment/docker
```
wongma7: can add a 'docker tag' to ensure :latest is always up-to-date regardless of $(VERSION) https://github.com/kubernetes-incubator/external-storage/blob/master/gluster/block/Makefile#L46

(copied this idea from service-catalog : ))

local-volume/provisioner/Makefile
```diff
@@ -37,18 +28,26 @@ else
...
+.PHONY: quick-container
 
 push: container
-	gcloud docker -- push $(IMAGE):$(VERSION)
+	docker push $(IMAGE):$(VERSION)
```
wongma7: ditto, push $(IMAGE):latest too every time there's new version

msau42:
Changed Makefile to look more more like gluster's.

PR #171, opened 2017-06-16
add cephfs provisioner rbac rules

rootfs:
@wongma7 @humblec @liggitt

### DIFF
--- a/ceph/cephfs/README.md
+++ b/ceph/cephfs/README.md
@@ -27,6 +27,11 @@ The following example uses `cephfs-provisioner-1` as the identity for the instan
 ```bash
 docker run -ti -v /root/.kube:/kube -v /var/run/kubernetes:/var/run/kubernetes --privileged --net=host  cephfs-provisioner /usr/local/bin/cephfs-provisioner -master=http://127.0.0.1:8080 -kubeconfig=/kube/config -id=cephfs-provisioner-1
 ```
+Alternatively, start a deployment:
+
+```bash
+kubectl create -f deployment.yaml
+```
 
 * Create a CephFS Storage Class
 
--- a/ceph/cephfs/auth/clusterrole.yaml
+++ b/ceph/cephfs/auth/clusterrole.yaml
@@ -1,0 +1,20 @@
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1alpha1
+metadata:
+  name: cephfs-provisioner-runner
+rules:
+  - apiGroups: [""]
+    resources: ["persistentvolumes"]
+    verbs: ["get", "list", "watch", "create", "delete"]
+  - apiGroups: [""]
+    resources: ["persistentvolumeclaims"]
+    verbs: ["get", "list", "watch", "update"]
+  - apiGroups: ["storage.k8s.io"]
+    resources: ["storageclasses"]
+    verbs: ["get", "list", "watch"]
+  - apiGroups: [""]
+    resources: ["events"]
+    verbs: ["watch", "create", "update", "patch"]
+  - apiGroups: [""]
+    resources: ["secrets"]
+    verbs: ["get", "create", "delete"]
--- a/ceph/cephfs/auth/openshift-clusterrole.yaml
+++ b/ceph/cephfs/auth/openshift-clusterrole.yaml
@@ -1,0 +1,20 @@
+kind: ClusterRole
+apiVersion: v1
+metadata:
+  name: cephfs-provisioner-runner
+rules:
+  - apiGroups: [""]
+    resources: ["persistentvolumes"]
+    verbs: ["get", "list", "watch", "create", "delete"]
+  - apiGroups: [""]
+    resources: ["persistentvolumeclaims"]
+    verbs: ["get", "list", "watch", "update"]
+  - apiGroups: ["storage.k8s.io"]
+    resources: ["storageclasses"]
+    verbs: ["get", "list", "watch"]
+  - apiGroups: [""]
+    resources: ["events"]
+    verbs: ["list", "watch", "create", "update", "patch"]
+  - apiGroups: [""]
+    resources: ["secrets"]
+    verbs: ["get", "create", "delete"]
--- a/ceph/cephfs/auth/clusterrolebinding.yaml
+++ b/ceph/cephfs/auth/clusterrolebinding.yaml
@@ -1,0 +1,12 @@
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1alpha1
+metadata:
+  name: run-cephfs-provisioner
+subjects:
+  - kind: ServiceAccount
+    name: cephfs-provisioner
+    namespace: default
+roleRef:
+  kind: ClusterRole
+  name: cephfs-provisioner-runner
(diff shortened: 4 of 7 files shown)


wongma7:
cephfs is lacking deployment yaml examples; the README assumes one is doing `docker run --privileged`, etc. IMO we should add those first

ceph/cephfs/auth/clusterrole.yaml
```diff
@@ -0,0 +1,20 @@
...
+    verbs: ["get", "list", "watch", "update"]
+  - apiGroups: ["storage.k8s.io"]
+    resources: ["storageclasses"]
+    verbs: ["get", "list", "watch"]
+  - apiGroups: [""]
+    resources: ["events"]
+    verbs: ["watch", "create", "update", "patch"]
+  - apiGroups: [""]
+    resources: ["secrets"]
+    verbs: ["get", "create", "delete"]
```
liggitt: I would _not_ expect global read/write access to secrets for a provisioner. I would much rather see this be namespaced, the provisioner given read/write access within a specific namespace, and the namespace containing the FS secrets be part of the information propagated into the PV

wongma7: ping @humblec let's make sure to handle this consistently

ceph/cephfs/auth/openshift-clusterrole.yaml
```diff
@@ -0,0 +1,20 @@
...
+    verbs: ["get", "list", "watch", "update"]
+  - apiGroups: ["storage.k8s.io"]
+    resources: ["storageclasses"]
+    verbs: ["get", "list", "watch"]
+  - apiGroups: [""]
+    resources: ["events"]
+    verbs: ["list", "watch", "create", "update", "patch"]
+  - apiGroups: [""]
+    resources: ["secrets"]
+    verbs: ["get", "create", "delete"]
```
liggitt: same here, I'd expect read/write access to be within a namespace, not global

rootfs: What's the magic keyword for that?

wongma7: namespaced role instead of clusterrole I think is meant

ceph/cephfs/auth/clusterrole.yaml
liggitt: that requires PVs to be able to reference secrets by namespace

ceph/cephfs/README.md
```diff
@@ -48,6 +48,12 @@ kubectl create -f claim.yaml
 kubectl create -f test-pod.yaml
 ```
 
+* Deploy
```
wongma7: it should be this *or* "Start CephFS provisioner" above I think. Else a user going step by step will do both and have 2 instances running & be very confused.

ceph/cephfs/auth/openshift-clusterrole.yaml
rootfs: It is a catch 22 to me, we have to know which namespace for the role

liggitt: create a role and rolebinding object without a namespace, you pass the namespace in when creating them with `kubectl create -f ... --namespace secret-namespace`

rootfs: @liggitt this is hard to do: the provisioner only gets the namespace when a pvc arrives and creates a user secret for mounter by then

liggitt: why does the secret holding the credentials for the PV go in the PVC namespace?

rootfs: the secret is part of the pv
https://github.com/kubernetes/kubernetes/blob/master/pkg/api/types.go#L972

and it is expected to be in the same namespace as the pod/pvc
https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/cephfs/cephfs.go#L103

liggitt: > the secret is part of the pv
https://github.com/kubernetes/kubernetes/blob/master/pkg/api/types.go#L972
> 
> and it is expected to be in the same namespace as the pod/pvc
https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/cephfs/cephfs.go#L103

I'm saying that data model doesn't make sense. It exposes details of how the storage backing the PVC to the user in the PVC namespace. At best, that means they can disrupt the operation by deleting the secret. At worst, it means admin credentials are exposed to namespaced users.

rootfs: > I'm saying that data model doesn't make sense. It exposes details of how the storage backing the PVC to the user in the PVC namespace. At best, that means they can disrupt the operation by deleting the secret. At worst, it means admin credentials are exposed to namespaced users.

The admin credentials are never exposed to PV or PVC.  What is in the PV is the unprivileged user's credential.

liggitt: It still doesn't make sense to give the user that credential... only the node needs it to mount/unmount, not the PVC-creating user.

rootfs: we needed a time traveling machine :)

liggitt: well, new provisioners like storageos and provisioners like azure that *are* copying admin credentials into namespaces are forcing revisiting the issue, which is good. Fixing the data model so that provisioners can place secrets (either admin or per-volume) into a separate namespace, reference the secret name/namespace in the PV, and only have their provisioner able to read/write secrets in that namespace seems like a good path forward.

rootfs: There is an issue with that model as well. If the secret is in a dedicated namespace, users in any namespaces can use the PVs. 

Then, if the PV's reclaim policy is Retain and status is unbound, data created on the PV are exposed.

wongma7:
PV that is Retained can't be reused/re-bound to another PVC. It just sits
there for admin to clean up.

rootfs:
+1 that @wongma7

rootfs:
Following up, what's the use case for users in namespaces other than the PVC creator's to access the PV? If the PV is gone once it is unbound?

liggitt:
> what's the use case for users in namespaces other than the PVC creator's to access the PV

so that an external provisioner can be limited to that namespace for reading/writing secrets, and so that the PVC user doesn't have access to the credentials for the underlying PV

rootfs:
> so that an external provisioner can be limited to that namespace for reading/writing secrets, and so that the PVC user doesn't have access to the credentials for the underlying PV

ok, that's fair.

rootfs:
before the going-forward path is here, what's the verdict for the existing model?

liggitt:
> before the going-forward path is here, what's the verdict for the existing model?

I can't speak for all clusters, but openshift would not enable dynamic provisioning as long as it is pushing secrets into end-users' namespaces, and I wouldn't expect provisioners to have global secret write permissions in the default RBAC policy

rootfs:
ok, let's revisit this issue once we have the new model in place.

PR #173, opened 2017-06-17
Update hostpath-provisioner to use new NewProvisionController

avagin:
Signed-off-by: Andrei Vagin <avagin@virtuozzo.com>


avagin:
Oops, I have not noticed that hostpath-provisioner works with the fixed version of external-storage, which is declared in glide.yaml.

PR #181, opened 2017-06-20
provision Cephfs PVs in designated volume

rootfs:
@wongma7 @humblec

### DIFF
--- a/ceph/cephfs/README.md
+++ b/ceph/cephfs/README.md
@@ -48,6 +48,17 @@ kubectl create -f claim.yaml
 kubectl create -f test-pod.yaml
 ```
 
+* Provision for claims in specific namespace
+
+If only provisioning for specific namespaces, start the provisioner with `-namespace=target_namespace`, e.g
+```bash
+docker run -ti -v /root/.kube:/kube -v /var/run/kubernetes:/var/run/kubernetes --privileged --net=host  cephfs-provisioner /usr/local/bin/cephfs-provisioner -master=http://127.0.0.1:8080 -kubeconfig=/kube/config -id=cephfs-provisioner-1 -namespace=myns
+```
+
+And test the claim:
+```bash
+kubectl create -f claim-ns.yaml
+```
 
 # Known limitations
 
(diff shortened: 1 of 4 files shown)


ceph/cephfs/README.md
```diff
@@ -48,6 +48,17 @@ kubectl create -f claim.yaml
 kubectl create -f test-pod.yaml
 ```
 
+* Provision for claims in specific namespaces
+
+If only provisioning for specific namespaces, start the provisioner with `-namespace=target_namespace`, e.g
```
humblec: can we accept more than one here ? if not 's/specific namespaces/specific namespace'

ceph/cephfs/cephfs-provisioner.go
```diff
@@ -40,6 +40,7 @@ const (
 	provisionCmd     = "/usr/local/bin/cephfs_provisioner"
 	provisionerIDAnn = "cephFSProvisionerIdentity"
 	cephShareAnn     = "cephShare"
+	defaultNamespace = v1.NamespaceAll
```
humblec: Instead of setting 'defaultNamespace' to "v1.NamespaceAll" , Isnt it good to set "default"  namespace as the defaultNamespace ?

ceph/cephfs/README.md
rootfs: no, provisioner can watch one namespace a time.

ceph/cephfs/cephfs-provisioner.go
rootfs: defaultNamespace is to be compatible with current behavior

ceph/cephfs/README.md
humblec: yep, then "specific namespaces" should be "specific namespace" as mentioned in the above comment :)

ceph/cephfs/cephfs-provisioner.go
humblec: k.SGTM.

wongma7:
Can we make it a storage class parameter instead? And allow for multiple namespace to be specified per storage class (comma separated string)? We can still List & Watch all PVCs but later filter them to only the namespaces we want.

I say this because it shoudl be configurable during runtime not just at start. And I'm afraid restricting the provisioner to a single namespace will make it much too limiting to use.

rootfs:
sgtm

humblec:
Yes, this is almost in the same line of what we discussed. It would be good to accept more namespaces in SC and making it run time option. Thanks all!

Issue #178, opened 2017-06-20
glide fails to build dependencies 


raffaelespazzoli:
I'm getting the following error if I run `glide up -v` on this repo:
```
[rspazzol@rspazzol external-storage]$ glide up -v
[INFO]	Downloading dependencies. Please wait...
[INFO]	--> Fetching updates for github.com/dgrijalva/jwt-go.
[INFO]	--> Fetching updates for github.com/aws/aws-sdk-go.
[INFO]	--> Fetching updates for github.com/docker/docker.
[INFO]	--> Fetching updates for k8s.io/apimachinery.
[INFO]	--> Fetching updates for github.com/heketi/heketi.
[INFO]	--> Fetching updates for k8s.io/kubernetes.
[INFO]	--> Fetching updates for github.com/guelfey/go.dbus.
[INFO]	--> Fetching updates for github.com/lpabon/godbc.
[INFO]	--> Fetching updates for k8s.io/client-go.
[INFO]	--> Fetching updates for github.com/mitchellh/mapstructure.
[INFO]	--> Fetching updates for github.com/golang/glog.
[INFO]	--> Setting version for github.com/mitchellh/mapstructure to db1efb5.
[INFO]	--> Setting version for github.com/docker/docker to v1.13.1.
[INFO]	--> Setting version for github.com/lpabon/godbc to 9577782.
[INFO]	--> Setting version for github.com/heketi/heketi to 7bb2c5b.
[INFO]	--> Setting version for github.com/aws/aws-sdk-go to v1.7.3.
[INFO]	--> Setting version for k8s.io/client-go to 450baa5d60f8d6a251c7682cb6f86e939b750b2d.
[INFO]	--> Setting version for k8s.io/apimachinery to 2de00c78cb6d6127fb51b9531c1b3def1cbcac8c.
[INFO]	--> Setting version for k8s.io/kubernetes to v1.6.0.
[INFO]	Resolving imports
[INFO]	--> Fetching updates for github.com/go-ini/ini.
[INFO]	--> Fetching updates for golang.org/x/sys.
[INFO]	Found Godeps.json file in /home/rspazzol/.glide/cache/src/https-k8s.io-apimachinery
[INFO]	--> Parsing Godeps metadata...
[INFO]	--> Fetching updates for github.com/go-openapi/spec.
[INFO]	--> Setting version for github.com/go-openapi/spec to 6aced65f8501fe1217321abf0749d354824ba2ff.
[...]

PR #182, opened 2017-06-20
Update client-go and fix glide

wongma7:
fix https://github.com/kubernetes-incubator/external-storage/issues/178


wongma7:
@rootfs @msau42 fyi helpers are now under "github.com/kubernetes-incubator/external-storage/lib/helper" if you need them. Because they don't exist anywhere else (besides k8s/k8s at the moment). Also had to import "k8s.io/kubernetes/pkg/kubelet/apis" 'cause `LabelHostname` was moved there.

PR #184, opened 2017-06-20
Fix auto push to quay for local volume

wongma7:


wongma7:
merging, none of this code is touched by travis.

PR #185, opened 2017-06-21
Fix auto push to quay for local volume for realsies

wongma7:
REGISTRY has a trailing slash for no good reason

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,4 +19,4 @@ deploy:
     script: ./deploy.sh
     on:
       tags: true
-      condition: $TEST_SUITE = everything-else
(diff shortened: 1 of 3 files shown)


ddysher:
@wongma7 this causes problems.

If one creates image using
```
make container VERSION=dev
```

then the output will be

quay.io/external_storagelocal-volume-provisioner:latest

wongma7:
Shoot, sorry, fixing in
https://github.com/kubernetes-incubator/external-storage/pull/189 . Now
auto build works but not manual

Review is important! :o

ddysher:
Feel free to ping me if needed :)

PR #188, opened 2017-06-21
Functional configurable provisioner (bootstrapper)

ddysher:
This PR makes configurable provisioner functional, including:

- Pass configs to bootstrapper via command line flag
- Automatically generates mountDir
- Update local-volume, bootstrapper and provisioner README

There are two refactoring commits, so it's better to review commit by commit.

@msau42


local-volume/bootstrapper/deployment/kubernetes/bootstrapper.yaml
```diff
@@ -1,21 +1,21 @@
...
   labels:
-    app: local-volume-bootstrapper
+    app: local-volume-provisioner-bootstrap
 spec:
   restartPolicy: Never
   serviceAccountName: local-storage-bootstrapper
   containers:
   - name: bootstrapper
-    image: local-volume-bootstrapper:dev
+    image: quay.io/external_storage/local-volume-provisioner-bootstrap
```
msau42: There should be a tag like ":latest"

local-volume/bootstrapper/deployment/kubernetes/bootstrapper.yaml
```diff
@@ -1,21 +1,21 @@
...
     env:
     - name: MY_NAMESPACE
       valueFrom:
         fieldRef:
           fieldPath: metadata.namespace
     args:
-    - --image="quay.io/external_storage/local-volume-provisioner"
+    - --image=quay.io/external_storage/local-volume-provisioner
```
msau42: Should this include the version tag too?

local-volume/bootstrapper/bootstrapper.go
```diff
@@ -67,13 +71,48 @@ func setupClient() *kubernetes.Clientset {
...
+	h.Write([]byte(mount.MountDir))
 	return fmt.Sprintf("mount-%x", h.Sum32())
 }
 
+// generateMountDir generates mount directory path in container, rule is to trim '/' prefix and
+// change "/" to "~", according to kubernetes convention, then concatenate with default root path.
+// E.g.
+//   "/mnt/ssds" -> "/mnt/local-storage/mnt~ssds"
+func generateMountDir(mount *common.MountConfig) string {
+	return path.Join(*mountRoot, strings.Replace(strings.TrimPrefix(mount.HostDir, "/"), "/", "~", -1))
```
msau42: Should we handle the corner case of the directory being too long?

local-volume/bootstrapper/bootstrapper.go
```diff
@@ -67,13 +71,48 @@ func setupClient() *kubernetes.Clientset {
...
+	var err error
+	configMap, err := client.CoreV1().ConfigMaps(namespace).Get(name, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+	configMap.Data, err = common.VolumeConfigToConfigMapData(config)
+	if err != nil {
+		return err
+	}
+	_, err = client.CoreV1().ConfigMaps(namespace).Update(configMap)
```
msau42: I'm slightly weary of updating a user-created config map.  Maybe it would be better to generate a new configmap?

local-volume/README.md
```diff
@@ -1,14 +1,17 @@
...
 ## Overview
+
 Local persistent volumes allows users to access local storage through the
 standard PVC interface in a simple and portable way.  The PV contains node
 affinity information that the system uses to schedule pods to the correct
 nodes.
 
-An external static provisioner is available to help simplify local storage
-management once the local volumes are configured.
+An external static provisioner and a related bootstrapper is available to
```
msau42: "are available"

local-volume/README.md
```diff
@@ -80,27 +89,33 @@ done
...
-### Running the external static provisioner
-This is optional, only for automated creation and cleanup of local volumes.
-See `provisioner/` for details and sample configuration files.
+### Step 2: Create local persistent volumes
 
-1. Create an admin account with persistentvolume provisioner and node system privileges.
+#### Option 1: Bootstrapping the external static provisioner
+
+This is optional, only for automated creation and cleanup of local volumes. See
+`bootstrapper/` and `provisioner/` for details and sample configuration files.
```
msau42: Make these links to the README

local-volume/README.md
```diff
@@ -110,29 +125,30 @@ kind: PersistentVolume
...
-                "nodeSelectorTerms": [
-                    { "matchExpressions": [
-                        { "key": "kubernetes.io/hostname",
-                          "operator": "In",
-                          "values": ["my-node"]
-                        }
-                    ]}
-                 ]}
-              }`,
+    "volume.alpha.kubernetes.io/node-affinity": `{
```
msau42: Docs review pointed this out to me, it should be a single quote instead of a backtick

local-volume/bootstrapper/README.md
```diff
@@ -1,9 +1,70 @@
...
-case of bootstrapper is to make provisioner configurable. It configures provisioner
-based on user config, creates appropriate service account, role bindings, then starts
-provisioner. It will exit as soon as provisioner is successfully created.
+local-volume-provisioner-bootstrap is used to bootstrap provisioner. The main use case of bootstrapper is to make
+provisioner configurable. Below is a detailed flow of how the bootstrap process works:
+
+- looks for a configmap passed in via flag `-volume-config`; otherwise, a default configmap named `local-volume-default-config` will be created
+- reads and validates the configmap, then auto-generates missing configurations, see below
+- looks for a service account passed in via flag `-serviceaccount`; otherwise, a default service account named `local-storage-admin` will be created
+- creates two cluster role bindings for the service account, i.e. `system:persistent-volume-provisioner` and `system:node`.
```
msau42: I think system:node actually has more privileges than we need.  Maybe as a future enhancement, we can look into using a different role with fewer permissions.

Also, are there any roles needed for the boostrapper to update/create the configmap?

local-volume/bootstrapper/deployment/kubernetes/bootstrapper.yaml
ddysher: updated to latest.

What about versioning. For simplicity, we can use same version number to bootstrap as well?, i.e.
```
quay.io/external_storage/local-volume-provisioner-bootstrap:1.0.0
```

ddysher: I don't think I have permission to push the image, can you push one once it's ready?

local-volume/bootstrapper/deployment/kubernetes/bootstrapper.yaml
ddysher: I've updated to version 1.0.0

local-volume/bootstrapper/README.md
ddysher: Yes, bootstrapper actually needs more privileges.  For update/create configmap, I believe edit will be fine, but creating service account that binds to system:node and pv-provisioner requires higher privilege.  Right now, we just give bootstrapper cluster-admin role (see admin-account.yaml).  

Agree we can look into it and narrow the permissions for both provisioner and bootstrapper.

local-volume/bootstrapper/bootstrapper.go
ddysher: If we handle this, which part should we truncate or should we create ids?  There is also a corner case where it conflicts with other path if we truncate..

This is not common so i think we can ignore it for now

local-volume/bootstrapper/bootstrapper.go
ddysher: The configmap is also passed to provisioner, it seems natural to update it and use in provisioner.  I agree it's not 100% secure; but right now, the config only has settings for volumes so I think it has little to no impact.  How about we revisit this when we have more settings in this configmap?

ddysher:
@msau42 Updated.  There's some suggestions that i think we can leave as future improvement.

local-volume/bootstrapper/README.md
```diff
@@ -1,9 +1,85 @@
...
+
+To see all options, compile bootstrapper and use `-h` option, below is a curated
+list of important options:
+
+- `-image`: Name of local volume provisioner image (default "quay.io/external_storage/local-volume-provisioner")
+- `-volume-config`: Name of the local volume configuration configmap. The configmap
+  must reside in the same namespace with bootstrapper. (default "local-volume-default-config")
+- `-serviceaccount`: Name of the service accout for local volume provisioner
+  (default "local-storage-admin")
+- `-mount-root`: Container root directory of volume mount path for discoverying
```
msau42: Does this need to be user configurable?

local-volume/bootstrapper/bootstrapper.go
```diff
@@ -35,19 +37,28 @@ import (
 )
 
 const (
-	defaultImage         = "local-volume-provisioner:dev"
-	defaultConfigMapName = "local-volume-config"
+	defaultImageName          = "quay.io/external_storage/local-volume-provisioner"
```
msau42: We also need the version tag here?

msau42:
Can you also update the TODO with the future items

local-volume/bootstrapper/README.md
ddysher: Not quite necessary, but i'm not sure if people want to change it... Since it's already there, I'll just remove the description from README.

local-volume/bootstrapper/bootstrapper.go
ddysher: Good catch, added.

ddysher:
@msau42 Updated, PTAL

PR #189, opened 2017-06-21
Add trailing slash to default REGISTRY

wongma7:
arguably the rest should not have trailing slash but okay


ddysher:
LGTM

nit: I think it is more natural to set registry to a value without the trailing slash, i.e. `quay.io/repo` instead of `quay.io/repo/`.

PR #192, opened 2017-06-23
Fixed feature flag in README for PersistentLocalVolumes

ianchakeres:
Fixed feature flag in README for PersistentLocalVolumes
Added instructions for executing PersistentLocalVolumes e2e tests

### DIFF
--- a/local-volume/README.md
+++ b/local-volume/README.md
@@ -48,7 +48,7 @@ Future features:
 ### Bringing up a cluster with local disks
 #### GCE
 ``` console
-KUBE_FEATURE_GATES=PersistentLocalVolumes NODE_LOCAL_SSDS=<n> kube-up.sh
+KUBE_FEATURE_GATES="PersistentLocalVolumes=true" NODE_LOCAL_SSDS=<n> cluster/kube-up.sh
 ```
 #### GKE (not available until 1.7)
 ``` console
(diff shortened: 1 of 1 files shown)


ianchakeres:
@msau42

local-volume/README.md
```diff
@@ -83,6 +83,13 @@ $ ALLOW_PRIVILEGED=true LOG_LEVEL=5 FEATURE_GATES=PersistentLocalVolumes=true ha
 3. Continue with [Running the external static provisioner](#running-the-external-static-provisioner)
    below.
 
+
+#### GCE E2E Testing
+``` console
+KUBE_FEATURE_GATES="PersistentLocalVolumes=true" NODE_LOCAL_SSDS=1 go run hack/e2e.go -- -v --up
```
msau42: The e2e should be able to run against any env, not just gce, so I would remove this line since it is already mentioned above.

ianchakeres: @msau42 - Does NODE_LOCAL_SSDs work against any env?

msau42: No, it only works on GCE.  But the details of bringing up a cluster with local disks for each environment are already covered up above.  So I think this entire line can be removed, and just keep the 2nd line that actually runs the tests.

ianchakeres: ok.

msau42:
/lgtm

Thanks!

PR #195, opened 2017-06-24
Make proper error messages in return paths.

humblec:


humblec:
@wongma7 Please review.

wongma7:
this breaks heketi (temporarily) right? heketiBlockVolRes is never filled in

humblec:
@wongma7 heketi mode is not operational. I can fill the essential members as I do with `default` hardcoded values. It does not make much sense though.  But I will be filling heketi response soon.

wongma7:
okay yeah that makes sense. lgtm!

humblec:
@wongma7  Thanks!

PR #196, opened 2017-06-26
Add RBAC examples for nfs-client

claytononeill:
Adds example configuration files for clusters that have RBAC enabled
(1.6) or OpenShift users.

This is mostly a cut and paste job from the EFS example.


wongma7:
lgtm, thank you!

PR #197, opened 2017-06-26
Snapshot: Add proposal document

tsmetana:
This is the first step in adding the external snapshot controller and provisioner. The document should reflect the current or near-future state of the prototype that currently lives at https://github.com/rootfs/snapshot.

@wongma7 PTAL
cc: @childsb @rootfs

### DIFF
--- a/snapshot/volume-snapshotting-proposal.md
+++ b/snapshot/volume-snapshotting-proposal.md
@@ -1,0 +1,517 @@
+Kubernetes Snapshotting Proposal
+================================
+
+**Authors:** [Cindy Wang](https://github.com/ciwang), [Jing Xu](https://github.com/jinxu97),[Tomas Smetana](https://github.com/tsmetana), [Huamin Chen ](https://github.com/rootfs)
+
+## Background
+
+Many storage systems (GCE PD, Amazon EBS, etc.) provide the ability to create "snapshots" of a persistent volumes to protect against data loss. Snapshots can be used in place of a traditional backup system to back up and restore primary and critical data. Snapshots allow for quick data backup (for example, it takes a fraction of a second to create a GCE PD snapshot) and offer fast recovery time objectives (RTOs) and recovery point objectives (RPOs).
+
+Typical existing backup solutions offer on demand or scheduled snapshots.
+
+An application developer using a storage may want to create a snapshot before an update or other major event. Kubernetes does not currently offer a standardized snapshot API for creating, listing, deleting, and restoring snapshots on an arbitrary volume.
+
+Existing solutions for scheduled snapshotting include [cron jobs](https://forums.aws.amazon.com/message.jspa?messageID=570265) and [external storage drivers](http://rancher.com/introducing-convoy-a-docker-volume-driver-for-backup-and-recovery-of-persistent-data/). Some cloud storage volumes can be configured to take automatic snapshots, but this is specified on the volumes themselves.
+
+## Objectives
+
+For the first version of snapshotting support in Kubernetes, only on-demand snapshots will be supported. Features listed in the roadmap for future versions are also nongoals.
+The initial version is created as an external controller and provisioner (i.e. two separate binaries).
+
+* Goal 1: Enable *on-demand* snapshots of Kubernetes persistent volumes by application developers.
+
+    * Nongoal: Enable *automatic* periodic snapshotting for direct volumes in pods.
+
+* Goal 2: Expose standardized snapshotting operations to create, list and delete snapshots in Kubernetes REST API.
+
+* Goal 3: Implement snapshotting interface for Amazon EBS.
+
+* Goal 4: Implement snapshotting interface for GCE PDs.
+
+### Feature Roadmap
+
+Major features, planned for the first version:
+
+* On demand snapshots
+
+    * API to create new snapshots
+
+    * API to list snapshots available to the user
+
+    * API to delete existing snapshots
+
+    * API to create a new persistent volume with persistent volume claim from a snapshot
+
+### Future Features
+
+The features that are not planned for the first version of the API bus should be considered in future versions:
+
+* Creating snapshots
+
+    * Scheduled and periodic snapshots
+
+    * Aplication initiated on-demand snapshot creation
+
+    * Support snapshot per PVC, pod or StatefulSet
+
+    * Support snapshots for non-cloud storage volumes (plugins might require actions to be triggered from the node)
+
+    * Support application-consistent snapshots (coordinate distributed snapshots across multiple volumes)
+
+    * Enable to create a pod/statefulsets with snapshots
+
+* List snapshots
+
+    * Enable to get the list of all snapshots for a specified persistent volume
+
+    * Enable to get the list of all snapshots for a pod/StatefulSet
+
+* Delete snapshots
+
+    * Enable to automatic garbage collect older snapshots when storage is limited
+
+* Quota management
+
+    * Enable to set quota for limiting how many snapshots could be taken and saved
+
+    * When quota is exceeded, delete the oldest snapshots automatically
+
+## Requirements
+
+### Performance
+
+* Time SLA from issuing a snapshot to completion:
+
+* The period we are interested is the time between the scheduled snapshot time, the time the snapshot is cut (it is safe to write to the snapshotted volume again) and the time the snapshot is finished uploading to its storage location.
+
+* This should be on the order of a few minutes.
+
+### Reliability
+
+* Data corruption
+
+    * Though it is generally recommended to stop application writes before executing the snapshot command, we will not do this automatically for several reasons:
+
+        * GCE and Amazon can create snapshots while the application is running.
+
+        * Stopping application writes cannot be done from the master and varies by application, so doing so will introduce unnecessary complexity and permission issues in the code.
+
+        * Most file systems and server applications are (and should be) able to restore inconsistent snapshots the same way as a disk that underwent an unclean shutdown.
+
+    * The data consistency would be best-effort only: e.g., call fsfreeze prior the snapshot on filesystems that support it.
+
+    * There are several proposed solutions that would enable the users to specify the action to perform prior/after the
+    snapshots: e.g. use pod annotations.
+
+* Snapshot failure
+
+    * Case: Failure during external process, such as during API call or upload
+
+        * Log error, do not attempt to retry
+
+    * Case: Failure within Kubernetes, such as controller restarts
+
+        *  If the master restarts in the middle of a snapshot operation, then the controller will find a snapshot request in pending state and should be able to successfully finish the operation.
+
+## Solution Overview
+
+There are a few uniqueness related to snapshots:
+
+* Both users and admins might create snapshots. Users should only get access to the snapshots belong to their namespaces. For this aspect, snapshot objects should be in user namespace. Admins might to choose expose the snapshots they created to some users who have access to those volumes.
+
+* After snapshots are taken, users might use them to create new volumes or restore the existing volumes back the time when the snapshot is taken.
+
+* There are use cases that data from snapshots taken from one namespace need to be accessible by users in another namespace.
+
+* For security purpose, if a snapshot object is created by a user, kubernetes should prevent other users duplicating this object in a different namespace if they happen to get the snapshot name.
+
+* There might be some existing snapshots taken by admins/users and they want to use those snapshots through kubernetes API interface.
+
+* **Create:**
+
+    1. The user creates a `VolumeSnapshot` referencing a persistent volume claim bound to a persistent volume
+
+    2. The controller fulfils the `VolumeSnapshot` by creating a snapshot using the volume plugins.
+
+    3. A new object `VolumeSnapshotData` is created to represent the actual snapshot binding the `VolumeSnapshot` with
+       the on-disk snapshot.
+
+* **List:**
+
+    1. The user is able to list all the `VolumeSnapshot` objects in the namespace
+
+* **Delete:**
+
+    1. The user deletes the `VolumeSnapshot`
+
+    2. The controller removes the on-disk snapshot. Note: snapshots have no notion of "reclaim policy" - there is
+       no way to recover the deleted snapshot.
+
+    3. The controller removes the `VolumeSnapshotData` object.
+
+* **Promote snapshot to PV:**
+
+    1. The user creates a persistent volume claim referencing the snapshot object in the annotation. The PVC must
+       belong to a `StorageClass` using the external volume snapshot provisioner. Note: the special annotation might get replaced
+       by a dedicated attribute of the `PersistentVolumeClaim` in the future.
+
+    2. The provisioner will use the `VolumeSnapshotData` object to create a persistent volume using the corresponding
+       volume snapshot plugin.
+
+    3. The PVC is bound to the newly created PV containing the data from the snapshot.
+
+
+The snapshot operation is a no-op for volume plugins that do not support snapshots via an API call (i.e. non-cloud storage).
+
+
+### API
+
+* The `VolumeSnapshot` object
+
+```
+// The volume snapshot object accessible to the user. Upon succesful creation of the actual
+// snapshot by the volume provider it is bound to the corresponding VolumeSnapshotData through
+// the VolumeSnapshotSpec
+type VolumeSnapshot struct {
+	metav1.TypeMeta `json:",inline"`
+	Metadata        metav1.ObjectMeta `json:"metadata"`
+
+	// Spec represents the desired state of the snapshot
+	// +optional
+	Spec VolumeSnapshotSpec `json:"spec" protobuf:"bytes,2,opt,name=spec"`
+
+	// Status represents the latest observer state of the snapshot
+	// +optional
+	Status VolumeSnapshotStatus `json:"status" protobuf:"bytes,3,opt,name=status"`
+}
+
+type VolumeSnapshotList struct {
+	metav1.TypeMeta `json:",inline"`
+	Metadata        metav1.ListMeta  `json:"metadata"`
+	Items           []VolumeSnapshot `json:"items"`
+}
+
+// The description of the volume snapshot
+type VolumeSnapshotSpec struct {
+	// Source represents the source of the volume snapshot
+	VolumeSnapshotSource `json:",inline" protobuf:"bytes,1,opt,name=volumeSnapshotSource"`
+	// VolumeSnapshotDataName binds the VolumeSnapshot object with the VolumeSnapshotData
+	// +optional
+	VolumeSnapshotDataName string `json:"volumeSnapshotDataName" protobuf:"bytes,2,opt,name=volumeSnapshotDataName"`
+}
+
+// The source of the volume snapshot
+type VolumeSnapshotSource struct {
+	// PersistentVolumeClaimName is the name of the PVC being snapshotted
+	// +optional
+	PersistentVolumeClaimName string `json:"persistentVolumeClaimName" protobuf:"bytes,1,opt,name=persistentVolumeClaimName"`
+
+	// ... other source types might be added in the future
+}
+
+type VolumeSnapshotStatus struct {
+	// The time the snapshot was successfully created
+	// +optional
+	CreationTimestamp metav1.Time `json:"creationTimestamp" protobuf:"bytes,1,opt,name=creationTimestamp"`
+
+	// Representes the lates available observations about the volume snapshot
+	Conditions []VolumeSnapshotCondition `json:"conditions" protobuf:"bytes,2,rep,name=conditions"`
+}
+
+type VolumeSnapshotConditionType string
+
+// These are valid conditions of a volume snapshot.
+const (
+	// VolumeSnapshotReady is added when the snapshot has been successfully created and is ready to be used.
+	VolumeSnapshotConditionReady VolumeSnapshotConditionType = "Ready"
+)
+
+// VolumeSnapshot Condition describes the state of a volume snapshot  at a certain point.
+type VolumeSnapshotCondition struct {
+	// Type of replication controller condition.
+	Type VolumeSnapshotConditionType `json:"type" protobuf:"bytes,1,opt,name=type,casttype=VolumeSnapshotConditionType"`
+	// Status of the condition, one of True, False, Unknown.
+	Status core_v1.ConditionStatus `json:"status" protobuf:"bytes,2,opt,name=status,casttype=ConditionStatus"`
+	// The last time the condition transitioned from one status to another.
+	// +optional
+	LastTransitionTime metav1.Time `json:"lastTransitionTime" protobuf:"bytes,3,opt,name=lastTransitionTime"`
+	// The reason for the condition's last transition.
+	// +optional
+	Reason string `json:"reason" protobuf:"bytes,4,opt,name=reason"`
+	// A human readable message indicating details about the transition.
+	// +optional
+	Message string `json:"message" protobuf:"bytes,5,opt,name=message"`
+}
+```
+
+* The `VolumeSnapshotData` object
+
+```
+// +genclient=true
+// +nonNamespaced=true
+
+// VolumeSnapshotData represents the actual "on-disk" snapshot object
+type VolumeSnapshotData struct {
+	metav1.TypeMeta `json:",inline"`
+	// +optional
+	Metadata metav1.ObjectMeta `json:"metadata"`
+
+	// Spec represents the desired state of the snapshot
+	// +optional
+	Spec VolumeSnapshotDataSpec `json:"spec" protobuf:"bytes,2,opt,name=spec"`
+
+	// Status represents the latest observed state of the snapshot
+	// +optional
+	Status VolumeSnapshotDataStatus `json:"status" protobuf:"bytes,3,opt,name=status"`
+}
+
+// The desired state of the volume snapshot
+type VolumeSnapshotDataSpec struct {
+	// Source represents the location and type of the volume snapshot
+	VolumeSnapshotDataSource `json:",inline" protobuf:"bytes,1,opt,name=volumeSnapshotDataSource"`
+
+	// VolumeSnapshotRef is part of bi-directional binding between VolumeSnapshot
+	// and VolumeSnapshotData
+	// +optional
+	VolumeSnapshotRef *core_v1.ObjectReference `json:"volumeSnapshotRef" protobuf:"bytes,2,opt,name=volumeSnapshotRef"`
+
+	// PersistentVolumeRef represents the PersistentVolume that the snapshot has been
+	// taken from
+	// +optional
+	PersistentVolumeRef *core_v1.ObjectReference `json:"persistentVolumeRef" protobuf:"bytes,3,opt,name=persistentVolumeRef"`
+}
+
+// Represents the actual location and type of the snapshot. Only one of its members may be specified.
+type VolumeSnapshotDataSource struct {
+	// HostPath represents a directory on the host.
+	// Provisioned by a developer or tester.
+	// This is useful for single-node development and testing only!
+	// On-host storage is not supported in any way and WILL NOT WORK in a multi-node cluster.
+	// More info: https://kubernetes.io/docs/concepts/storage/volumes#hostpath
+	// +optional
+	HostPath *HostPathVolumeSnapshotSource `json:"hostPath,omitempty"`
+	// AWSElasticBlockStore represents an AWS Disk resource that is attached to a
+	// kubelet's host machine and then exposed to the pod.
+	// More info: https://kubernetes.io/docs/concepts/storage/volumes#awselasticblockstore
+	// +optional
+	AWSElasticBlockStore *AWSElasticBlockStoreVolumeSnapshotSource `json:"awsElasticBlockStore,omitempty"`
+
+    // ... etc. for other snapshot types
+}
+```
+
+An example of the `VolumeSnapshotDataSource` for Amazon EBS snapshots:
+
+```
+// AWS EBS volume snapshot source
+type AWSElasticBlockStoreVolumeSnapshotSource struct {
+	// Unique id of the persistent disk snapshot resource. Used to identify the disk snapshot in AWS
+	SnapshotID string `json:"snapshotId"`
+}
+```
+
+### Event loop in controller
+
+The volume snapshot controller maintains two data structures (ActualStateOfWorld and DesiredStateOfWorld) and
+periodically reconciles the two. The data structures are being update by the API sever event handlers.
+
+* If a new `VolumeSnapshot` is added, the *add* handler adds it to the DesiredStateOfWorld (DSW)
+
+* If a `VolumeSnapshot` is deleted, the *delete* handler removes it from the DSW.
+
+* Reconciliation loop in the controller
+
+    * For every `VolumeSnapshot` in the ActualStateOfWorld (ASW) find the corresponding `VolumeSnapshot` in the DSW.
+    If such a snapshot does not exist, start a snapshot deletion operation:
+
+        * Determine the correct volume snapshot plugin to use from the `VolumeSnapshotData` referenced by the
+        `VolumeSnapshot`
+
+        * Create a delete operation: only one such operation is allowed to exist for the given `VolumeSnapshot` and
+        `VolumeSnapshotData` pair
+
+        * The operation is an asynchronous function using the volume plugin to delete the actual snapshot in the back-end.
+
+        * When the plugin finishes deleting the snapshot, delete the `VolumeSnapshotData` referencing it and remove the
+        `VolumeSnapshot` reference from the ASW
+
+    * For every `VolumeSnapshot` in the DSW find the corresponding `VolumeSnapshot` in the ASW. If such a snapshot
+    does not exist, start a snapshot creation operation:
+
+        * Determine the correct volume snapshot plugin to use from the `VolumeSnapshotData` referenced by the
+        `VolumeSnapshot`
+
+        * Create a volume snapshot creation operation: only one such operation is allowed to exist for the given
+        `VolumeSnapshot` and `VolumeSnapshotData` pair.
+
+        * The operation is an asynchronous function using the volume plugin to create the actual snapshot in the back-end.
+
+        * When the plugin finishes creating the snapshot a new `VolumeSnapshotData` is created holding a reference to
+        the actual volume snapshot.
+
+    * For every snapshot present in the ASW and DSW find its `VolumeSnapshotData` and verify the bi-directional
+    binding is correct: if not, update the `VolumeSnapshotData` reference.
+
+
+### Create Snapshot Logic
+
+To create a snapshot:
+
+* Acquire operation lock for volume so that only one one snapshot creation operation is running for the specified volume
+
+    * Abort if there is already a pending operation (reconciliation loop will retry, if needed).
+
+* Spawn a new thread:
+
+    * Execute the volume-specific logic to create a snapshot of the persistent volume referenced by the PVC.
+
+    * For any errors, log the error, send it as an event on the corresponding `VolumeSnapshot`, and terminate the thread (the main controller will retry as needed).
+
+    * Once a snapshot is created successfully:
+
+        * Make a call to the API server to add the new snapshot ID/timestamp to the `VolumeSnapshotData` API object, update its
+        status.
+
+### Snapshot to PV promotion logic
+
+For the `PersistentVolumeClaim` provisioned by the snapshot external provisioner:
+
+* Check the `PeristentVolumeClaim` annotation and get the `VolumeSnapshot` name.
+  
+* Retrieve the `VolumeSnapshot` object from the API server:
+  
+    * Verify both the `VolumeSnapshot` and `PersistentVolumeClaim` belong to the same namespace.
+    
+    * Verify the `VolumeSnapshotData` referenced by the `VolumeSnapshot` exists and that its `VolumeSnapshot` reference points "back" to the given `VolumeSnapshot` object.
+    
+* Find the correct plugin to use to create a `PersistentVolume` bound to the given PVC.
+
+## Example Use Case
+
+### Alice wants to backup her MySQL database data
+
+Alice is a DB admin who runs a MySQL database and needs to backup the data on a remote server prior to the database
+upgrade. She has a short maintenance window dedicated to the operation that allows her to pause the dabase only for
+a short while. Alice will therefore stop the database, create a snapshot of the data, re-start the database and after
+that start time-consuming network transfer to the backup server.
+
+The database is running in a pod with the data stored on a persistent volume:
+```yaml
+apiVersion: v1
+kind: Pod
+metadata:
+  name: mysql
+  labels:
+    name: mysql
+spec:
+  containers:
+    - resources:
+        limits :
+          cpu: 0.5
+      image: openshift/mysql-55-centos7
+      name: mysql
+      env:
+        - name: MYSQL_ROOT_PASSWORD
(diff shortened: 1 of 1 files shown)


snapshot/volume-snapshotting-proposal.md
```diff
@@ -0,0 +1,412 @@
...
+
+* Creating snapshots
+
+    * Scheduled and periodic snapshots
+
+    * Aplication initiated on-demand snapshot creation
+
+    * Support snapshot per PVC, pod or StatefulSet
+
+    * Support snapshots for non-cloud storage volumes (i.e. plugins that require actions to be triggered from the node)
```
humblec: May be this can be better worded. That said,  non-cloud storage volumes are not clear to me. Thinking specificically about GlusterFS, it dont require any actions to be triggered from the kubelet.

snapshot/volume-snapshotting-proposal.md
```diff
@@ -0,0 +1,412 @@
...
+
+    * When quota is exceeded, delete the oldest snapshots automatically
+
+## Requirements
+
+### Performance
+
+* Time SLA from issuing a snapshot to completion:
+
+* The period we are interested is the time between the scheduled snapshot time and the time the snapshot is finishes uploading to its storage location.
```
humblec: May be a typo here , "is finishes".

snapshot/volume-snapshotting-proposal.md
```diff
@@ -0,0 +1,412 @@
...
+    * Though it is generally recommended to stop application writes before executing the snapshot command, we will not do this automatically for several reasons:
+
+        * GCE and Amazon can create snapshots while the application is running.
+
+        * Stopping application writes cannot be done from the master and varies by application, so doing so will introduce unnecessary complexity and permission issues in the code.
+
+        * Most file systems and server applications are (and should be) able to restore inconsistent snapshots the same way as a disk that underwent an unclean shutdown.
+
+    * The data consistency would be best-effort only: e.g., call fsfreeze prior the snapshot on filesystems that support it.
+
```
humblec: IMO, data consistency when performing snapshot should be left to the storage provider. As  a second thought, the volume plugins could report/declare a  field/attribute like `online`  or `offline` ( their snapshot capability) in one of the spec. So that the controller can be better taught to perform snapshot. This may come handy in future when we expand more use cases.

snapshot/volume-snapshotting-proposal.md
```diff
@@ -0,0 +1,412 @@
...
+        *  If the master restarts in the middle of a snapshot operation, then the controller will find a snapshot request in pending state and should be able to successfully finish the operation.
+
+## Solution Overview
+
+There are a few uniqueness related to snapshots:
+
+* Both users and admins might create snapshots. Users should only get access to the snapshots belong to their namespaces. For this aspect, snapshot objects should be in user namespace. Admins might to choose expose the snapshots they created to some users who have access to those volumes.
+
+* After snapshots are taken, users might use them to create new volumes or restore the existing volumes back the time when the snapshot is taken.
```
humblec: Does this mean, user2  can restore the snapshot created by user1  in same namespace? Will there be any security issue ? Just as a thought, do  we need to tag the snapshot to user UUID for better security ?

snapshot/volume-snapshotting-proposal.md
```diff
@@ -0,0 +1,412 @@
...
+
+* There are use cases that data from snapshots taken from one namespace need to be accessible by users in another namespace.
+
+* For security purpose, if a snapshot object is created by a user, kubernetes should prevent other users duplicating this object in a different namespace if they happen to get the snapshot name.
+
+* There might be some existing snapshots taken by admins/users and they want to use those snapshots through kubernetes API interface.
+
+* **Create:**
+
+    1. The user creates a `VolumeSnapshot` refrencing a persistent volume claim bound to a persistent volume
```
humblec: s/refrencing/referencing

snapshot/volume-snapshotting-proposal.md
tsmetana: Don't know. I would start by copying what we do with PV/PVC...

snapshot/volume-snapshotting-proposal.md
tsmetana: Will fix.

rootfs:
In the use case below, one `VolumeSnapshot` could contain an array of  `VolumeSnapshotData`. All the `VolumeSnapshotData` are identical except that they are in different zones. 

Alice has a MySQL instance running in Zone East. She makes a snapshot of MySQL database in Zone East and copies the snapshot to Zone West. If Zone East is down, a new instance in Zone West will be spawned and the snapshot in Zone West is promoted to a PV.

In such use case, one `VolumeSnapshot` has two copies of `VolumeSnapshotData` in two Zones.

snapshot/volume-snapshotting-proposal.md
```diff
@@ -0,0 +1,421 @@
...
+// the VolumeSnapshotSpec
+type VolumeSnapshot struct {
+	metav1.TypeMeta
+	Metadata        metav1.ObjectMeta
+
+	// Spec represents the desired state of the snapshot
+	Spec VolumeSnapshotSpec
+
+	// Status represents the latest observer state of the snapshot
+	Status VolumeSnapshotStatus
```
jsafrane: VolumeSnapshotStatus is not described anywhere. How can user see if the snapshot has been taken?

snapshot/volume-snapshotting-proposal.md
```diff
@@ -0,0 +1,421 @@
...
+// VolumeSnapshotData represents the actual "on-disk" snapshot object
+type VolumeSnapshotData struct {
+	metav1.TypeMeta
+	Metadata metav1.ObjectMeta
+
+	// Spec represents the desired state of the snapshot
+	Spec VolumeSnapshotDataSpec
+
+	// Status represents the latest observed state of the snapshot
+	Status VolumeSnapshotDataStatus
```
jsafrane: VolumeSnapshotDataStatus is not described anywhere

snapshot/volume-snapshotting-proposal.md
```diff
@@ -0,0 +1,421 @@
...
+	// and VolumeSnapshotData
+	VolumeSnapshotRef *core_v1.ObjectReference
+
+	// PersistentVolumeRef represents the PersistentVolume that the snapshot has been
+	// taken from
+	PersistentVolumeRef *core_v1.ObjectReference
+}
+
+// Represents the actual location and type of the snapshot. Only one of its members may be specified.
+type VolumeSnapshotDataSource struct {
```
jsafrane: At least some example for AWS and GCE would be nice here.

snapshot/volume-snapshotting-proposal.md
```diff
@@ -0,0 +1,421 @@
...
+	Status VolumeSnapshotStatus
+}
+
+// The desired state of the volume snapshot
+type VolumeSnapshotSpec struct {
+	// PersistentVolumeClaimName is the name of the PVC being snapshotted
+	PersistentVolumeClaimName string
+
+	// SnapshotDataName binds the VolumeSnapshot object with the VolumeSnapshotData
+	SnapshotDataName string
```
jsafrane: Add `// omitempty` and some note that it can't be changed once it's set. I am not sure about CRD, but validation in API server should insist on this.

snapshot/volume-snapshotting-proposal.md
```diff
@@ -0,0 +1,421 @@
...
+}
+
+// The desired state of the volume snapshot
+type VolumeSnapshotDataSpec struct {
+	// Source represents the location and type of the volume snapshot
+	VolumeSnapshotDataSource
+
+	// VolumeSnapshotRef is part of bi-directional binding between VolumeSnapshot
+	// and VolumeSnapshotData
+	VolumeSnapshotRef *core_v1.ObjectReference
```
jsafrane: Again, add // omitempty and some note that it can't be changed once it's set.

snapshot/volume-snapshotting-proposal.md
```diff
@@ -0,0 +1,421 @@
...
+
+	// PersistentVolumeRef represents the PersistentVolume that the snapshot has been
+	// taken from
+	PersistentVolumeRef *core_v1.ObjectReference
```
jsafrane: `// omitempty` for snapshots taken outside of Kubernetes?

snapshot/volume-snapshotting-proposal.md
```diff
@@ -0,0 +1,421 @@
...
+
+### API
+
+* The `VolumeSnapshot` object
+
+```
+// The volume snapshot object accessible to the user. Upon succesful creation of the actual
+// snapshot by the volume provider it is bound to the corresponding VolumeSnapshotData through
+// the VolumeSnapshotSpec
+type VolumeSnapshot struct {
```
jsafrane: Should VolumeSnapshot / VolumeSnapshotData have a Capacity? How we're going to handle quota of snapshots?

snapshot/volume-snapshotting-proposal.md
```diff
@@ -0,0 +1,421 @@
...
+* Event loop in controller
+
+The volume snapshot controller maintains two data structures (ActualStateOfWorld and DesiredStateOfWorld) and
+periodically reconciles the two. The data structures are being update by the API sever event handlers.
+
+* If a new `VolumeSnapshot` is added, the *add* handler adds it to the DesiredStateOfWorld (DSW)
+
+* If a `VolumeSnapshot` is deleted, the *delete* handler removes it from the DSW.
+
+* Reconciliation loop in the controller
```
jsafrane: This loop does not cover restoring a snapshot into a PV/PVC. Some link to "default external provisioner workflow / library" would be nice.

For security reasons the provisioner must check that 1) the referenced VolumeSnapshot is indeed in the same namespace as the PVC and 2) that referenced VolumeSnapshotData point back to the right VolumeSnapshot (and not to a different VolumeSnapshot in a different namespace). Attacker must not be able to steal data by crafting a VolumeSnapshot + PVC in his namespace pointing to a VolumeSnapshotData belonging to a different namespace.

snapshot/volume-snapshotting-proposal.md
```diff
@@ -0,0 +1,421 @@
...
+
+    * Aplication initiated on-demand snapshot creation
+
+    * Support snapshot per PVC, pod or StatefulSet
+
+    * Support snapshots for non-cloud storage volumes (i.e. plugins that require actions to be triggered from the node)
+
+    * Support application-consistent snapshots (coordinate distributed snapshots across multiple volumes)
+
+    * Enable to create a pod/statefulsets with snapshots
```
jsafrane: Create a *new* pod/statefulsets from *existing* snapshots? Or create a snapshot of a pod/stateful set? That's already covered above in "Support snapshot per PVC, pod or StatefulSet"

snapshot/volume-snapshotting-proposal.md
```diff
@@ -0,0 +1,421 @@
...
+    3. A new object `VolumeSnapshotData` is created to represent the actual snapshot binding the `VolumeSnapshot` with
+       the on-disk snapshot.
+
+* **List:**
+
+    1. The user is able to list all the `VolumeSnapshot` objects in the namespace
+
+* **Delete:**
+
+    1. The user deletes the `VolumeSnapshot`
```
jsafrane: I would emphasize somewhere that we don't implement any ReclaimPolicy, i.e. all snapshots are deleted when user removes `VolumeSnapshot`.

snapshot/volume-snapshotting-proposal.md
```diff
@@ -0,0 +1,421 @@
...
+
+    2. The controller removes the on-disk snapshot.
+
+    3. The controller removes the `VolumeSnapshotData` object.
+
+* **Promote snapshot to PV:**
+
+    1. The user creates a persistent volume claim referencing the snapshot object in the annotation. The PVC must
```
jsafrane: Is this long term solution? IMO we would like to use regular storage class, this external volume snapshot provisioner is just a temporary solution, right?

jsafrane: Also, annotation on PVC is used only during alpha, we should design complete API with real fields first and then  implement these fields as alpha annotations during development.

snapshot/volume-snapshotting-proposal.md
```diff
@@ -0,0 +1,421 @@
...
+
+    * For every snapshot present in the ASW and DSW find its `VolumeSnapshotData` and verify the bi-directional
+    binding is correct: if not, update the `VolumeSnapshotData` reference.
+
+
+### Create Snapshot Logic
+
+To create a snapshot:
+
+* Acquire operation lock for volume so that no other attach or detach operations can be started for volume.
```
jsafrane: How exactly are you going to lock attach/detach controller?

snapshot/volume-snapshotting-proposal.md
```diff
@@ -0,0 +1,421 @@
...
+
+    * Abort if there is already a pending operation for the specified volume (main loop will retry, if needed).
+
+* Spawn a new thread:
+
+    * Execute the volume-specific logic to create a snapshot of the persistent volume referenced by the PVC.
+
+    * For any errors, log the error, and terminate the thread (the main controller will retry as needed).
```
jsafrane: + send the errors as events on corresponding `VolumeSnapshot` object so user can see what's going on in `kubectl describe volumesnapshot`

snapshot/volume-snapshotting-proposal.md
tsmetana: Yes. Will add this.

snapshot/volume-snapshotting-proposal.md
tsmetana: We don't handle the quotas anyhow yet. Which is why there is no Capacity attribute anywhere.

snapshot/volume-snapshotting-proposal.md
tsmetana: This is only for the external provisioner. I hope once snapshots will become internal (or at least internally supported) by Kubernetes the PVC annotations would go away. And I actually don't thing we ever considered much more than the VolumeSnapshot reference in the PVC so there is not much to design here. What did I miss?

snapshot/volume-snapshotting-proposal.md
tsmetana: OK.

snapshot/volume-snapshotting-proposal.md
tsmetana: Looking at the discussions in the [Google doc](https://docs.google.com/document/d/1SyViEo1okCJRwrM2LtE-DM6K6G-O46tYCBNCjNaS0Lg/edit#heading=h.z4bjntixeo48) it's really the former. Which is perhaps something that doesn't need to have any special support in the snapshots themselves. I'd leave it here until we know for sure and to demonstrate we do know about this use-case.

snapshot/volume-snapshotting-proposal.md
tsmetana: OK.

snapshot/volume-snapshotting-proposal.md
tsmetana: Will add it.

snapshot/volume-snapshotting-proposal.md
tsmetana: The controller does not promote snapshots to PV at all: it's the provisioner. I'll try to describe the part in more detail too. As for the second part: good point, I'll add it too.

snapshot/volume-snapshotting-proposal.md
tsmetana: I won't. This is a historical artifact...

snapshot/volume-snapshotting-proposal.md
```diff
@@ -0,0 +1,509 @@
...
+
+* Goal 4: Implement snapshotting interface for GCE PDs.
+
+### Feature Roadmap
+
+Major features, planned for the first version:
+
+* On demand snapshots
+
+    * API to create new snapshots
```
jingxu97: should we also a little bit detail about create  such as the snapshot source is PVC?

snapshot/volume-snapshotting-proposal.md
jingxu97: maybe remove i.e., just add a comment about some plugins might require actions to be triggered from the node

snapshot/volume-snapshotting-proposal.md
```diff
@@ -0,0 +1,509 @@
...
+
+    * API to list snapshots available to the user
+
+    * API to delete existing snapshots
+
+    * API to create a persistent volume from a snapshot
```
jingxu97: maybe mention to create a new persistent volume (and its corresponding claim)?

snapshot/volume-snapshotting-proposal.md
jingxu97: Yes, what I mean in the doc is the former. But this should not belong to creating snapshots category. This is restore from snapshots. For first version, we only support create new volumes from snapshots. The future version might consider directly restore pods from the snapshot. Does it make sense?

snapshot/volume-snapshotting-proposal.md
```diff
@@ -0,0 +1,509 @@
...
+
+## Requirements
+
+### Performance
+
+* Time SLA from issuing a snapshot to completion:
+
+* The period we are interested is the time between the scheduled snapshot time and the time the snapshot is finished uploading to its storage location.
+
+* This should be on the order of a few minutes.
```
jingxu97: I think it is very possible to go up to hours.

snapshot/volume-snapshotting-proposal.md
```diff
@@ -0,0 +1,509 @@
...
+
+    * When quota is exceeded, delete the oldest snapshots automatically
+
+## Requirements
+
+### Performance
+
+* Time SLA from issuing a snapshot to completion:
+
+* The period we are interested is the time between the scheduled snapshot time and the time the snapshot is finished uploading to its storage location.
```
jingxu97: Also I feel the time when snapshot is cut (i.e., when it is safe to write data on disk) is also critical. If user needs to pause their application before taking snapshot, he can know when to resume the application.

snapshot/volume-snapshotting-proposal.md
jingxu97: I don't think normally admin sets up in a way that two different users are in the same namespace? In the case both users have the right permission in the same namespace, that means they can perform the operations such as restoring snapshots etc because the system is set up in this way..

snapshot/volume-snapshotting-proposal.md
```diff
@@ -0,0 +1,509 @@
...
+
+    * Case: Failure within Kubernetes, such as controller restarts
+
+        *  If the master restarts in the middle of a snapshot operation, then the controller will find a snapshot request in pending state and should be able to successfully finish the operation.
+
+## Solution Overview
+
+There are a few uniqueness related to snapshots:
+
+* Both users and admins might create snapshots. Users should only get access to the snapshots belong to their namespaces. For this aspect, snapshot objects should be in user namespace. Admins might to choose expose the snapshots they created to some users who have access to those volumes.
```
jingxu97: Do we have a clean idea how to " expose the snapshots they created to some users who have access to those volumes"?

snapshot/volume-snapshotting-proposal.md
jingxu97: As @jsafrane this part is temporary solution. Maybe we should clarify that and put what we plan to do too, e.g., adding fromVolumeSnapshot field..

snapshot/volume-snapshotting-proposal.md
jingxu97: Current storage quota supports the total number of pvc in a namespace or storageclass. Similarly we could first support the total number of snapshots allowed. 
For capacity management, we don't have much support yet. Will consider it in future.

snapshot/volume-snapshotting-proposal.md
```diff
@@ -0,0 +1,509 @@
...
+	Status VolumeSnapshotStatus `json:"status" protobuf:"bytes,3,opt,name=status"`
+}
+
+type VolumeSnapshotList struct {
+	metav1.TypeMeta `json:",inline"`
+	Metadata        metav1.ListMeta  `json:"metadata"`
+	Items           []VolumeSnapshot `json:"items"`
+}
+
+// The desired state of the volume snapshot
```
jingxu97: The description of a volume snapshot?

snapshot/volume-snapshotting-proposal.md
```diff
@@ -0,0 +1,509 @@
...
+type VolumeSnapshotSpec struct {
+	// PersistentVolumeClaimName is the name of the PVC being snapshotted
+	// +optional
+	PersistentVolumeClaimName string `json:"persistentVolumeClaimName" protobuf:"bytes,1,opt,name=persistentVolumeClaimName"`
```
jingxu97: I thought we might plan to have different sources for taking the snapshot in the future. So should we put something like the doc https://docs.google.com/document/d/1SyViEo1okCJRwrM2LtE-DM6K6G-O46tYCBNCjNaS0Lg 

// Source represents the source of the snapshot
Source VolumeSnapshotSource

type VolumeSnapshotSource struct {
	PersistentVolumeClaimName *string  // For user-created snapshots in the same namespace
....

snapshot/volume-snapshotting-proposal.md
```diff
@@ -0,0 +1,509 @@
...
+
+	// SnapshotDataName binds the VolumeSnapshot object with the VolumeSnapshotData
+	// +optional
+	SnapshotDataName string `json:"snapshotDataName" protobuf:"bytes,2,opt,name=snapshotDataName"`
```
jingxu97: maybe VolumeSnapshotDataName to avoid confution?

snapshot/volume-snapshotting-proposal.md
```diff
@@ -0,0 +1,509 @@
...
+}
+
+type VolumeSnapshotStatus struct {
+	// The time the snapshot was successfully created
```
jingxu97: During snapshot creation, there are several time point. For example, in gce pd,
CREATING means we are talking to the Nanny/PD to tell them to prepare to take a snapshot.
UPLOADING means that the snapshot has been cut, it's safe to write new data.
READY means the snapshot is usable and can be restored to any disk thereafter.
Is the time here means snapshot is ready to use?

snapshot/volume-snapshotting-proposal.md
```diff
@@ -0,0 +1,509 @@
...
+	// Representes the lates available observations about the volume snapshot
+	Conditions []VolumeSnapshotCondition `json:"conditions" protobuf:"bytes,2,rep,name=conditions"`
+}
+
+type VolumeSnapshotConditionType string
+
+// These are valid conditions of a volume snapshot.
+const (
+	// VolumeSnapshotReady is added when the snapshot has been successfully created and is ready to be used.
+	VolumeSnapshotConditionReady VolumeSnapshotConditionType = "Ready"
```
jingxu97: I am thinking whether to add uploading phase so that user can have a idea when is the time to continue to write data.

snapshot/volume-snapshotting-proposal.md
```diff
@@ -0,0 +1,509 @@
...
+	// Spec represents the desired state of the snapshot
+	// +optional
+	Spec VolumeSnapshotDataSpec `json:"spec" protobuf:"bytes,2,opt,name=spec"`
+
+	// Status represents the latest observed state of the snapshot
+	// +optional
+	Status VolumeSnapshotDataStatus `json:"status" protobuf:"bytes,3,opt,name=status"`
+}
+
+// The desired state of the volume snapshot
```
jingxu97: desired state --> description

snapshot/volume-snapshotting-proposal.md
```diff
@@ -0,0 +1,509 @@
...
+	// +optional
+	HostPath *HostPathVolumeSnapshotSource `json:"hostPath,omitempty"`
+	// AWSElasticBlockStore represents an AWS Disk resource that is attached to a
+	// kubelet's host machine and then exposed to the pod.
+	// More info: https://kubernetes.io/docs/concepts/storage/volumes#awselasticblockstore
+	// +optional
+	AWSElasticBlockStore *AWSElasticBlockStoreVolumeSnapshotSource `json:"awsElasticBlockStore,omitempty"`
+
+    // ... etc. for other snapshot types
+}
```
jingxu97: Could you also add the following

// GCEPersistentDiskSnapshotSource represents an GCE PD snapshot resource
--
  | // +optional
  | GCEPersistentDiskSnapshot *GCEPersistentDiskSnapshotSource `json:"gcePersistentDisk,omitempty"`

snapshot/volume-snapshotting-proposal.md
jingxu97: Do we also have the loop to periodically check the status of the snapshot?

snapshot/volume-snapshotting-proposal.md
tsmetana: Yes, however deeper in the document: this section is meant to be a high-level overview only.

snapshot/volume-snapshotting-proposal.md
tsmetana: OK. Will do.

snapshot/volume-snapshotting-proposal.md
tsmetana: Good point. I will try to mention this too.

snapshot/volume-snapshotting-proposal.md
tsmetana: Not really. I have sketched [an idea in the snapshotter code](https://github.com/rootfs/snapshot/blob/master/pkg/controller/snapshotter/snapshotter.go#L239) but I'm not sure it would work and haven't quite looked in the details yet.

snapshot/volume-snapshotting-proposal.md
tsmetana: I admit I got lost here... We're snapshotting volumes. How do I restore a pod from a volume snapshot?

snapshot/volume-snapshotting-proposal.md
tsmetana: We have `VolumeSnapshot` and `VolumeSnapshotData`. The `SnapshotRequest` did not quite follow the Kubernetes API concept of being "intent-based": see https://kubernetes.io/docs/concepts/overview/working-with-objects/kubernetes-objects/ Thus we let the user to create directly a `VolumeSnapshot` object to expresses the desired state.

snapshot/volume-snapshotting-proposal.md
tsmetana: Clearly I was just being lazy here... `VolumeSnapshotDataName` is definitely more consistent with the rest of the API.

snapshot/volume-snapshotting-proposal.md
tsmetana: The snapshot status and its changes needs a lot of work definitely. We need some set of conditions that would make sense for all the snapshot backends. We need some set of states/conditions that are universal enough for all the backends. Currently the implementation looks like this:
We have three conditions: "Ready", "Pending" and "Error". It is possible for the object to exist with empty status condition array meaning the snapshot is being created, once the on-disk snapshot exists, the "Pending" condition is added -- we have the snapshot, but it is not usable yet, when everything finishes OK, the "Pending" condition is replaced with "Ready". "Error" is a generic condition signalizing something went wrong.

snapshot/volume-snapshotting-proposal.md
tsmetana: Yes, this might be useful. It roughly corresponds to the "Pending" state we use now.

snapshot/volume-snapshotting-proposal.md
tsmetana: Copied from PersistentVolumes I think... But yes, it's rather a description since it changes.

snapshot/volume-snapshotting-proposal.md
```diff
@@ -0,0 +1,509 @@
...
+
+	// VolumeSnapshotRef is part of bi-directional binding between VolumeSnapshot
+	// and VolumeSnapshotData
+	// +optional
+	VolumeSnapshotRef *core_v1.ObjectReference `json:"volumeSnapshotRef" protobuf:"bytes,2,opt,name=volumeSnapshotRef"`
+
+	// PersistentVolumeRef represents the PersistentVolume that the snapshot has been
+	// taken from
+	// +optional
+	PersistentVolumeRef *core_v1.ObjectReference `json:"persistentVolumeRef" protobuf:"bytes,3,opt,name=persistentVolumeRef"`
```
tsmetana: I'm sure I asked this somewhere before... But it still bothers me: Is the PV reference really useful? The PV might be long gone even though its snapshot exists. What's worse there might exist a completely different PV with the same name which sounds like a recipe for all kinds of confusion.

snapshot/volume-snapshotting-proposal.md
tsmetana: AWS is there just as an example (see the comment): it is the best-known thing among the users so I picked that one.

snapshot/volume-snapshotting-proposal.md
tsmetana: There is some rudimentary state check in the reconciler too. It checks whether the objects present in both states of world has all the required attributes set correctly and fixes it eventually. It's there primarily to correct eventual failed API server operations.

tsmetana:
I made some changes to the document. Thanks everyone for the comments.

snapshot/volume-snapshotting-proposal.md
jingxu97: I am not sure to use empty status condition to represent a status (e.g., snapshot request is being created?) because of confusion. I assume normally when an object is created, it has some status?

snapshot/volume-snapshotting-proposal.md
jingxu97: any thoughts about it?

snapshot/volume-snapshotting-proposal.md
jingxu97: The description of a volume snapshot? I am trying to following other API object such as pod, pvc...

snapshot/volume-snapshotting-proposal.md
```diff
@@ -0,0 +1,510 @@
...
+    * The data consistency would be best-effort only: e.g., call fsfreeze prior the snapshot on filesystems that support it.
+
+    * There are several proposed solutions that would enable the users to specify the action to perform prior/after the
+    snapshots: e.g. use pod annotations.
+
+* Snapshot failure
+
+    * Case: Failure during external process, such as during API call or upload
+
+        * Log error, retry until success (indefinitely)
```
jingxu97: In last snapshot sync up meeting, we agree to not retry. Could you please update the doc?

snapshot/volume-snapshotting-proposal.md
jingxu97: I feel using a VolumeSnapshotSource would be much easier to extend unless we are very sure PersistentVolumeClaim is the only source. 
For extension, maybe we can add source to represent some existing snapshots for the snapshots sharing across namespace use case.

snapshot/volume-snapshotting-proposal.md
jingxu97: I am thinking to use the following conditions, CREATED, READY, FAILED with the following meanings

- CREATED(no) READY(no) FAILED(no): the snapshot request is pending or just initiated
- CREATED(yes) READY(no) FAILED(no): the snapshot request is received and issue, and the snapshot is cut (after this timepoint, user can resume their applications)
- CREATED(yes) READY(yes) FAILED(no): the snapshot is created and finishes uploading the data, it is ready to use.
- CREATED(yes/no) READY(no) FAILED(yes): the snapshot is not ready to use. It might failed before created or after during uploading.

snapshot/volume-snapshotting-proposal.md
rootfs: In the prototype, PV spec is not really used other than validating supported types.

tsmetana: Hm... Didn't we patch this already? Since even that sounds wrong (for the same reasons I mention above)...

snapshot/volume-snapshotting-proposal.md
```diff
@@ -0,0 +1,517 @@
...
+    * The data consistency would be best-effort only: e.g., call fsfreeze prior the snapshot on filesystems that support it.
+
+    * There are several proposed solutions that would enable the users to specify the action to perform prior/after the
+    snapshots: e.g. use pod annotations.
+
+* Snapshot failure
+
+    * Case: Failure during external process, such as during API call or upload
+
+        * Log error, do not attempt to retry
```
enisoc: I wasn't in the design meeting. Can you add more context on why you decided not to retry? Is it to avoid accidentally spamming snapshot creations? If so, do you have a plan for preventing such spamming if the controller crash-loops at a step after it sends the external creation request?

snapshot/volume-snapshotting-proposal.md
jingxu97: @enisoc, I think your proposal of conditions is good. I forgot there is an Unknown condition which could be used to represent initial state. Thanks!

snapshot/volume-snapshotting-proposal.md
jingxu97: Yes, we prefer not to retry because normally snapshot creation failed for a good reason, retry will most likely fail again. Also user might find a snapshot is finally taken long after the request is sent. 
When controller crash-loops, the system might try to take a snapshot again. But I think if we make sure the snapshot name is the same, the volume provider should return an error indicating the snapshot is already taken.

snapshot/volume-snapshotting-proposal.md
```diff
@@ -0,0 +1,517 @@
...
+
+
+### API
+
+* The `VolumeSnapshot` object
+
+```
+// The volume snapshot object accessible to the user. Upon succesful creation of the actual
+// snapshot by the volume provider it is bound to the corresponding VolumeSnapshotData through
+// the VolumeSnapshotSpec
```
jingxu97: I think it is a typo, it should be VolumeSnapshotStatus. Similar to PVC and PV, the bound information is updated in PersistentVolumeClaimStatus

snapshot/volume-snapshotting-proposal.md
```diff
@@ -0,0 +1,517 @@
...
+// AWS EBS volume snapshot source
+type AWSElasticBlockStoreVolumeSnapshotSource struct {
+	// Unique id of the persistent disk snapshot resource. Used to identify the disk snapshot in AWS
+	SnapshotID string `json:"snapshotId"`
+}
+```
+
+### Event loop in controller
+
+The volume snapshot controller maintains two data structures (ActualStateOfWorld and DesiredStateOfWorld) and
```
jingxu97: DSW and ASW are used in kubelet volume manager and master attach_detach controller. The basic design is that there is a separate DesiredStatePopulator to periodically add and remove pods and their volume information into the DSW. A reconciler will peridocally compare ASW and DSW and take actions such as attach/detach mount/mount to update the ASW. In case of controller restarts, all the information in ASW and DSW will be gone. DSW could be repopulated by the populator. For AWS kubelet volume manager add reconstruct function by reading the disk directories (it contains most of the information we need to reconstruct, but we do have a few cases that could not be constructed). While the attach_detach controller keeps some volume attachement information in the node status, so we added a similar reconstruct function by reading the node status.
I think for snapshot, we should also use VolumeSnapshotStatus to recover the ASW. Or instead of having an explicit ASW and DSW, use the similar way of other controllers. 
@tsmetana @rootfs what do you think about it?

snapshot/volume-snapshotting-proposal.md
```diff
@@ -0,0 +1,517 @@
...
+        * Create a volume snapshot creation operation: only one such operation is allowed to exist for the given
+        `VolumeSnapshot` and `VolumeSnapshotData` pair.
+
+        * The operation is an asynchronous function using the volume plugin to create the actual snapshot in the back-end.
+
+        * When the plugin finishes creating the snapshot a new `VolumeSnapshotData` is created holding a reference to
+        the actual volume snapshot.
+
+    * For every snapshot present in the ASW and DSW find its `VolumeSnapshotData` and verify the bi-directional
+    binding is correct: if not, update the `VolumeSnapshotData` reference.
```
jingxu97: @tsmetana Could you explain more details about this part? Thanks!

snapshot/volume-snapshotting-proposal.md
tsmetana: Yes, this makes sense and makes better use of the conditions as they're defined now. Thanks.

snapshot/volume-snapshotting-proposal.md
tsmetana: We've had some discussion about this already. The ASW/DSW approach is definitely not the only one possible but it should work. The tricky part is the controller startup when it needs to scan the objects (VolumeSnapshot, VolumeSnapshotData and the on-disk snapshots) and reconstruct ASW according to the reality. There are several scenarios of what may go wrong, eg. controller is down and user creates a VolumeSnapshot (then there will be no VolumeSnapshotData meaning the snapshot needs to be created), user deletes a VolumeSnapshot while the controller is down (there will be VolumeSnapshotData with no corresponding VolumeSnapshot meaning the snapshot has to be removed), etc. I can't really judge how much easier would it be without the ASW/DSW but I'm open to ideas of course.

snapshot/volume-snapshotting-proposal.md
tsmetana: The binding is incorrect when it's incomplete. The code should assume the API writes are unreliable and may fail. When the user creates the VolumeSnapshot its VolumeSnapshotData reference is empty, then the controller creates the snapshot and VolumeSnapshotData: the VolumeSnapshotData is always created only if the actual snapshot creation succeeded, and always with reference to VolumeSnapshot. Then the VolumeSnapshot object is updated (this may fail). So the VolumeSnapshot may exits with missing reference to the VolumeSnapshotData and this has to be fixed...

Thinking about it: I can leave this part out eventually since it sounds like an implementation detail (there are more ways to ensure the objects consistency).

snapshot/volume-snapshotting-proposal.md
jsafrane: PV <--> PVC binding is stored in PV.Spec.ClaimRef and PVC.Spec.VolumeName. Similarly, "binding" (=scheduling) of a pod to a node is done in Pod.Spec.NodeName.

I was told that it must be possible to reconstruct the Status from observation of the system. From https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md

> The status summarizes the current state of the object in the system, and is usually persisted with the object by an automated processes but may be generated on the fly. At some cost and perhaps some temporary degradation in behavior, the status could be reconstructed by observation if it were lost.

It's not possible reconstruct binding by observing .Spec parts of the objects if the binding is not stored there.

snapshot/volume-snapshotting-proposal.md
jsafrane: Instances VolumeSnapshot / VolumeSnapshotData are usual Kubernetes objects and RBAC applies to them.

snapshot/volume-snapshotting-proposal.md
jsafrane: IMO, the main reason for not trying again was that the application that's running in a pod whose volume is snapshoted must run in some degraded mode (e.g. mysql blocks write operations, fielsystem is frozen with`fsfreeze`) and we don't want to keep the application in this mode for a long time with endless retries.

tsmetana:
Updated the VolumeSnapshot status conditions.

childsb:
Congrats on the merge!!!  i'm excited to see the code merge soon

PR #203, opened 2017-06-27
Correct chapAuth condition check

humblec:

### DIFF
--- a/gluster/block/cmd/glusterblock-provisioner/glusterblock-provisioner.go
+++ b/gluster/block/cmd/glusterblock-provisioner/glusterblock-provisioner.go
@@ -190,44 +190,54 @@ func (p *glusterBlockProvisioner) Provision(options controller.VolumeOptions) (*
 		return nil, fmt.Errorf("glusterblock: failed to create volume: %v", createErr)
 	}
 
+	iscsiVol := &iscsiSpec{}
+	if blockVol != nil {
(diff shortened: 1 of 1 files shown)


humblec:
@wongma7 Please review

gluster/block/cmd/glusterblock-provisioner/glusterblock-provisioner.go
```diff
@@ -347,7 +354,11 @@ func (p *glusterBlockProvisioner) createVolume(volSizeInt int, blockVol string)
 			}
 			json.Unmarshal([]byte(out), execBlockRes)
 
-			if execBlockRes.User == "" || execBlockRes.AuthKey == "" {
+			if *execBlockRes == nil {
```
wongma7: check for json.Unmarshal error instead? it can be non-nil but some minor error (certain field didn't get filled in)

humblec:
@wongma7 PTAL .. thanks!

gluster/block/cmd/glusterblock-provisioner/glusterblock-provisioner.go
humblec: @wongma7  Actually I have put mandatory field checks later in the code path and in the return function . However thought of having an extra caution here, as I do dereference of this struct pointer in next lines.

humblec:
@wongma7 Addressed your comment. Thanks.

PR #205, opened 2017-06-28
[Do Not Merge] Add local volume e2e

ddysher:
Add a very basic e2e tests for local volume, which
- Launch bootstrapper in shell script (local-volume/test/e2e/run_e2e.sh)
- In e2e_test.go, check provisioner daemonset is created
- Then, mkdir and test local pv is created, and re-created when pv is in released state

@wongma7 @msau42 

@wongma7 need ur help on vendor.  I have to manually copy k8s.io/test/e2e/generated/bindata.go to vendor directory; otherwise e2e won't compile.  Do you have any idea why this happens?  BTW, this file is quite large.


wongma7:
Yep that generated file is the issue with trying to use the e2e framework out of tree. It basically contains every json/yaml manifest e2e tests might need. We have 2 options here:
* vendor in a version of the framework where framework does not depend on generated. My PR to remove framework's dependence on generated upstream was finally merged, but I haven't had a chance to verify, can you try this commit in your test/e2e/glide.yaml https://github.com/kubernetes/kubernetes/commit/156ff3a72da16d8975fadb29a751aea98d8474b8 .
* generate the bindata as part of the test script. Since glide pulls in all of kube codebase into vendor/k8s.io/kubernetes theoretically we can generate it on-demand instead of manually copying from somewhere. Something like `go generate ./vendor/k8s.io/kubernetes/test/e2e/generated/gobindata_util.go`

bonus option: maintain our own mini e2e framework like service-catalog does. IMO this is overkill for the basic tests we can run on travis :)

local-volume/test/e2e/glide.yaml
```diff
@@ -0,0 +1,40 @@
+package: github.com/kubernetes-incubator/external-storage/local-volume/test/e2e
```
wongma7: I would prefer if we did not check in local-volume/test/e2e/vendor (add it to .gitignore & `make clean`?) simply because cloners will most likely not need it and it's huge. 

For the NFS e2e tests, I do glide install -v every time before I run the e2e tests, can we do something like that? If the vendor directory already exists, it takes a couple of second for glide to verify else it pulls it on-demand

msau42:
Can we add this e2e test to the main kubernetes repo?  Then it can also get run as part of CI.

wongma7:
That's the approach I recommend :) . It wouldn't hurt to have both though; local-up-cluster on Travis works fine for super basic smoke testy type things. Anything more complicated, e.g. multi node and upstream ci should be leveraged. As demonstrated by the vendor issues here, doing tests here is an additional maintenance burden and there's no guarantees about stability of e2e framework/local-up-cluster. Up to y'all if you want this.

ddysher:
@wongma7 @msau42 thanks for ur inputs.  I never thought of putting the tests into kubernetes repo since  this is totally out-of-tree. (To be honest, I don't think it's the right way to do :)  But I understand the problem here, so I'll instead put the test into kubernetes repo.

The basic e2e test is really easy to add, but I spend quite a few time dealing with vendor problems. It really hurts..... Once I get a chance, I'll help with the mini framework, please loop me in if u plan to work on it @wongma7 

I'll keep this open and close once we have the e2e at kubernetes main repo.

ianchakeres:
@ddysher @wongma7  - you might want to take a look at this PR https://github.com/kubernetes/kubernetes/pull/47999

I've made some notable changes to the existing local storage e2e tests, and you can probably re-use a bunch of it.

ddysher:
@ianchakeres thanks for the heads up.  I'll start once that PR is merged.

ddysher:
Close in favor of https://github.com/kubernetes/kubernetes/issues/48860

PR #213, opened 2017-07-03
Update Local Storage examples

Robbilie:


ddysher:
LGTM  @Robbilie you need to sign CLA

Robbilie:
@ddysher just did, took me a while to figure out what to put into the "title" field in the agreement… :D

local-volume/README.md
```diff
@@ -71,7 +71,7 @@ gcloud alpha container node-pools create ... --local-ssd-count=<n>
 2. Mount all the filesystems under one directory per StorageClass. The directories
    are specified in a configmap, see below. By default, the discovery directory is
    `/mnt/disks` and storage class is `local-storage`.
-3. Configure a Kubernetes cluster with the `PersistentLocalVolumes` feature gate.
+3. Configure the Kubernetes API Server with the `PersistentLocalVolumes` & `AffinityInAnnotations` feature gates.
```
ianchakeres: @Robbilie - Can you please add AffinityInAnnotations feature flag to the text in Option 1?

Robbilie: Sure

Robbilie: I cannot confirm that this works though and adds these feature gates to the Google hosted API server

ianchakeres: @msau42 @saad-ali - can you please confirm/deny whether its important to have AffinityInAnnotations feature flag set when using local storage on GCE?

msau42: You do not need the AffinityInAnnotations flag. That is for pod node affinity.

msau42: For GKE, you cannot pass in individual feature gates. You need to use " gcloud alpha container cluster create". So step 3 is not needed.

Robbilie: i couldve sworn that when i setup the api server with only the PersistentLocalVolumes feature gate it didnt work.
i probably didnt properly apply the change, retried it now and it works…
my bad

Robbilie: @msau42 step 3 is for bare metal clusters :)
for gke as you suggested it says you need to use the command you mentioned

local-volume/examples/local-reader.yaml
```diff
@@ -24,4 +24,4 @@ spec:
       volumes:
       - name: local-vol
         persistentVolumeClaim: 
-          claimName: "local-vol-local-test-0"
+          claimName: "example-local-claim"
```
msau42: This local reader example is paired with the statefulset example, not manual pv.  It has a bash script that reads the contents of the statefulset.  The simple pvc example is for manual pv.

PR #215, opened 2017-07-04
fix docker file error

wjin:
the deployment docker image got error:

chmod: invalid mode: 'x+o'

Signed-off-by: Wei Jin <wjin.cn@gmail.com>


wjin:
please review and update the docker image if possible, thanks.

wongma7:
I think the deployment.yaml is still broken btw, because CMD is just chmod. I will let @rootfs look at it when he gets back

wjin:
Yes, you are right. 
For `docker run`, it will be ok as we append `cephfs_provisioner` command in the end.
For deployment, maybe he added `local-start.sh` to Dockerfile when building the base image: "quay.io/external_storage/cephfs-provisioner:latest".

PR #225, opened 2017-07-10
[local-volume] Added recorder for logging of errors in local-volume deleter

ianchakeres:
This PR resolves two TODOs related to logging of errors in the deleter for local-volume.

### DIFF
--- a/local-volume/provisioner/pkg/deleter/deleter.go
+++ b/local-volume/provisioner/pkg/deleter/deleter.go
@@ -35,7 +35,9 @@ type Deleter struct {
 // NewDeleter creates a Deleter object to handle the cleanup and deletion of local PVs
 // allocated by this provisioner
 func NewDeleter(config *common.RuntimeConfig) *Deleter {
-	return &Deleter{RuntimeConfig: config}
+	return &Deleter{
+		RuntimeConfig: config,
+	}
 }
 
 // DeletePVs will scan through all the existing PVs that are released, and cleanup and
@@ -49,17 +51,17 @@ func (d *Deleter) DeletePVs() {
 			// Cleanup volume
 			err := d.cleanupPV(pv)
 			if err != nil {
-				// TODO: Log event on PV
-				glog.Errorf("Error cleaning PV %q: %v", name, err.Error())
+				cleaningLocalPVErr := fmt.Errorf("Error cleaning PV %q: %v", name, err.Error())
+				d.RuntimeConfig.Recorder.Eventf(pv, v1.EventTypeWarning, common.EventVolumeFailedDelete, cleaningLocalPVErr.Error())
 				continue
 			}
 
 			// Remove API object
 			err = d.APIUtil.DeletePV(name)
 			if err != nil {
-				// TODO: Log event on PV
 				// TODO: Does delete return an error if object has already been deleted?
-				glog.Errorf("Error deleting PV %q: %v", name, err.Error())
+				deletingLocalPVErr := fmt.Errorf("Error deleting PV %q: %v", name, err.Error())
+				d.RuntimeConfig.Recorder.Eventf(pv, v1.EventTypeWarning, common.EventVolumeFailedDelete, deletingLocalPVErr.Error())
 				continue
 			}
 			glog.Infof("Deleted PV %q", name)
--- a/local-volume/provisioner/pkg/common/common.go
+++ b/local-volume/provisioner/pkg/common/common.go
@@ -27,6 +27,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/record"
 	"k8s.io/kubernetes/pkg/kubelet/apis"
 )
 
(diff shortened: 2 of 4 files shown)


ianchakeres:
@msau42 - after applying my fix for Frsize, I was able to run make test on my mac. I need to replumb the recorder for the deleter_test. I'll let you know when I've completed the test plumbing.

ianchakeres:
@msau42 - I've refactored the Recorder to live in common RuntimeConfig, since many parts of the provisioner may need to log errors. I also updated the deleter_test, and unit tests pass. Note: I did need to use the Frsize fix for my mac to perform the unit testing.

local-volume/provisioner/pkg/deleter/deleter.go
```diff
@@ -26,6 +26,12 @@ import (
 	"k8s.io/client-go/pkg/api/v1"
 )
 
+const (
+	// kubelet event reason list
+	FailedCleaningLocalPV = "FailedCleaningLocalPV"
```
msau42: I think we need to add these events to the kubernetes main repo too.  Try looking for where the FailedMount event is defined.  Also, check if there's already a suitable event defined that we can just reuse.

ianchakeres: I didn't see any existing events that seemed quite right, but FailedUnmount or FailedDetachVolume might be close enough. 

Here are the four Volume related kevents:

```
FailedAttachVolume                   = "FailedAttachVolume"
FailedDetachVolume                   = "FailedDetachVolume"
FailedMountVolume                    = "FailedMount"
FailedUnMountVolume                  = "FailedUnMount"

```

https://github.com/kubernetes/kubernetes/blob/886e04f1fffbb04faf8a9f9ee141143b2684ae68/pkg/kubelet/events/event.go#L50

@msau42 - what do you prefer?

msau42: @wongma7 @jsafrane what do external provisioners do when the deletion fails?  I'm thinking we may need to add a new event "FailedDeleteVolume"

wongma7: VolumeFailedDelete https://github.com/kubernetes/kubernetes/blob/555fdf425da68f04a10f73440c4f6a7b76fdbf07/pkg/controller/volume/events/event.go#L27

ianchakeres:
@wongma7 - I'm getting errors via the CI about being unable to find pkg/controller/volume/events

`pkg/deleter/deleter.go:28:2: cannot find package "k8s.io/kubernetes/pkg/controller/volume/events" in any of:`

https://travis-ci.org/kubernetes-incubator/external-storage/jobs/253233018

Do I need to run glide commands to be able to reference events.go?

ianchakeres:
@msau42 - PTAL and approve if things look good.

msau42:
lgtm. Go ahead and squash the commits, but keep the vendoring commit separate.

ianchakeres:
@msau42 - commits squashed

glide.lock
```diff
@@ -271,6 +271,7 @@ imports:
   - kubernetes/typed/storage/v1/fake
   - kubernetes/typed/storage/v1beta1
   - kubernetes/typed/storage/v1beta1/fake
+  - pkg/api/v1
```
msau42: Do we need to bring in the v1 package?  Shouldn't client-go have it already?

wongma7: No because pkg/volume/util depends on "k8s.io/kubernetes/pkg/api" specifically so we've no choice but to pick it up. Plus, pkg/api was kicked out of client-go and is supposed to live in k8s.io/api now.

Tangent:
The conundrum that is k8s + golang dependency management has been coming up a lot recently, it's probably time to document it properly (see https://github.com/kubernetes-incubator/external-storage/pull/237 and my halfhearted attempts at https://github.com/kubernetes-incubator/external-storage/tree/master/docs).

It used to be just client-go, then client-go + apimachinery, now there's a half dozen repos that must be in version or commit-hash sync. (btw, client-go has broken to some extent every quarter.) To be fair, glide and other tools should be smart enough to figure that dependency compatibility stuff out, and external-storage is a bit unique in that it has no choice but to chase after k8s master instead of using conveniently tagged versions like typical consumers.

I hope that doesn't sound like complaining, just want to write down openly somewhere my present thoughts, I'll work on documenting external-storage's vendoring "policies" next week. For the record, sig-api-machinery are the experts on all this stuff & from what I can tell are well aware of the unpleasantness that comes with writing an app for k8s. I haven't been following these developments closely enough, I guess they have to balance the needs of core k8s and k8s consumers.

msau42: Thanks for the explanation!

msau42: Hmmm actually @ianchakeres, the other external provisioners are logging PV events already, but they don't have to pull in all these vendoring changes.  They are hardcoding the event name, so I think that's why they didn't need to pull that in.  Can we do that here too?

wongma7: Np, though I seem to have confused this PR with the other, oops :). This one isn't pulling pkg/volume/util it's pulling events for a single const, in this case we may as well redefine the const here since there's no value added from vendoring the extra code (unless you count extra bytes in `.git` )

ianchakeres: @msau42 & @wongma7 - For this PR I'll plan to copy in the one const, and get rid of the vendor updates.

ianchakeres:
@msau42 @wongma7 - PTAL - commits squashed and vendoring removed.

Issue #224, opened 2017-07-10
[local-volume] Frsize not supported on mac osx (darwin) - unable to run local-volume make test


ianchakeres:
On mac osx (darwin) Statfs_t.Frsize is not supported, and this situation is not allowing me to run make test on the local-volume provisioner.

```
$ make test
go test ./...
# github.com/kubernetes-incubator/external-storage/local-volume/provisioner/pkg/util
pkg/util/volume_util.go:116: s.Frsize undefined (type unix.Statfs_t has no field or method Frsize)
make: *** [test] Error 2
```

Here is the specific reference in the code: https://github.com/kubernetes-incubator/external-storage/pull/135/files#diff-2e4832ad7196c01aca3d8c395325fc45R116

ianchakeres:
Related to this PR: https://github.com/kubernetes-incubator/external-storage/pull/135

ianchakeres:
@msau42 @ddysher - FYI

ianchakeres:
Another related discussion: https://github.com/kubernetes-incubator/external-storage/issues/132

Issue #132, opened 2017-05-24
[local-volume] Detect capacity of mount points


ddysher:
In alpha release, static provisioner should post PV with capacity field; we'll start with standard linux utilities.

ddysher:
ref https://github.com/kubernetes-incubator/external-storage/pull/135

msau42:
How accurate do we want to be with the capacity?

I ran on GCE with local SSDs, which are 375GB.  The actual PV created has this capacity, which is a little unreadable.
    storage: 367238044Ki

ddysher:
Yeah, quantity package will use the best unit without losing precision.  Considering disk space is usually quite large, how about we default to GB and give the option for user to change to smaller unit?

If we prefer so, we should call out in the doc that this is a rounded down number.

msau42:
I'm fine with rounding to GB

ianchakeres:
@ddysher & @msau42 - approximately how much of the total space is reserved by the file system? I'm wondering if the amount is generally smaller than rounding errors. Re: https://github.com/kubernetes-incubator/external-storage/pull/154#discussion_r121831808

ddysher:
@ianchakeres sorry for late reply.  It depends on the type of file system. For ext file system, it defaults to 5%, and this is reserved to root users.  For xfs, it defaults to 5% or 8192 fsb, whichever is smaller.  I didn't look at other fs.

So for xfs, it's possible the reserved space is smaller than rounding errors. What's the problem do you foresee?

ianchakeres:
@ddysher - I'm just wondering whether users care and will understand the nuance of the FSAvailableBytes calculation when compared with the other FSInfo available calculation, especially after rounding. My intuition says that we probably need only one method/definition of available, and I'd love your advice on which one is the right definition/calculation.

ianchakeres:
@ddysher - we might want to add a RoundDownSize method to kubernetes/pkg/volume/util.go, like the RoundUpSize function on line 261 - https://github.com/kubernetes/kubernetes/blob/38837b018bf2a9215653054c2ec59636c0ba9440/pkg/volume/util.go#L261

msau42:
I would prefer to be consistent with what the kubernetes system is reporting in the fs volume metrics.

ddysher:
@ianchakeres sounds good!

[...]

Issue #165, opened 2017-06-13
Library consumers do not have access to some upstream volume util functions


wongma7:
Come up a couple times now but some volume util functions people expect to have available when writing volume plugins/provisioners are not automatically copied to client-go and so not readily available to us https://github.com/kubernetes/kubernetes/blob/38837b018bf2a9215653054c2ec59636c0ba9440/pkg/volume/util.go

We have copied `RoundUpSize` to a util package. We can keep copying them on an as-is needed basis or look into refactoring things upstream such that the functions we need are automatically copied to client-go

msau42:
Yeah, I would look into refactoring so we can get the functions into client-go

caesarxuchao:
The future k8s.io/pkg repository seems to be a better home. See https://github.com/kubernetes/kubernetes/issues/48209#issuecomment-313870481.

client-go shouldn't contain type specific helper functions.

wongma7:
SGTM! I'm all on board with having more purposeful repos & moving away from 'have everything arbitrarily in client-go' :)

PR #226, opened 2017-07-10
[local-volume] Refactored GetFsAvailableByte to avoid using Frsize in local-volume provisioner

ianchakeres:
GetFsAvailableByte now uses the same method as kubernetes/pkg/volume/util/fs.go:FsInfo(path) to calculate available bytes

Fixes: https://github.com/kubernetes-incubator/external-storage/issues/224
Related to this discussion: https://github.com/kubernetes-incubator/external-storage/issues/132

After this PR is merged, mac osx (darwin) uses can successfully execute make test on the local-volume provisioner.


ianchakeres:
@msau42 @ddysher - PTAL.

local-volume/provisioner/pkg/util/volume_util.go
```diff
@@ -106,14 +106,17 @@ func (u *volumeUtil) DeleteContents(fullPath string) error {
 
 // GetFsAvailableByte returns available capacity in byte about a mounted filesystem.
 // fullPath is the pathname of any file within the mounted filesystem. Capacity
-// returned here is total capacity available to non-root users, and does not include
-// fs reserved space.
+// returned here is total capacity available.
 func (u *volumeUtil) GetFsAvailableByte(fullPath string) (uint64, error) {
-	var s unix.Statfs_t
-	if err := unix.Statfs(fullPath, &s); err != nil {
+	//largely copied from kubernetes/pkg/volume/util/fs.go:FsInfo(path)
```
ddysher: +1  can we reuse it?

ianchakeres: I had some vendoring issues related to log_dir being defined twice, by glog being imported twice.

For example, 
`panic: /var/folders/59/yc7_f4fd53nbyw868zqpk78cn7shx9/T/go-build393895106/github.com/kubernetes-incubator/external-storage/local-volume/provisioner/pkg/discovery/_test/discovery.test flag redefined: log_dir
`

I was able to resolve it by locally executing
`glide install --strip-vendor --strip-vcs`

by following the instructions [here on stack overflow](https://stackoverflow.com/questions/37284423/glog-flag-redefined-error).  

I haven't dealt with vendoring before. 

Do you know where I'd add the appropriate flags to glide to ensure that dependencies are flattened?

msau42: @wongma7 can you help Ian import this K8s library

wongma7: we most likely can't (yet) because it imports "k8s.io/kubernetes/pkg/api."

but yes --strip-vendor/-v is the correct flag needed to remove nested dependencies & to flatten everything

msau42: Can you explain the limitation?

wongma7: There will only be issues if we use functions from kubernetes/kubernetes/pkg/volume/util/ that accept or return kubernetes/kubernetes/pkg/api types, because we want to work with downstream k8s.io/api types instead.

It seems Fsinfo doesn't have this issue, so we won't run into any compilation errors when vendoring kubernetes/kubernetes/pkg/volume/util/ & using it. But generally we want to solve this issue for long-term so we can make use of existing code https://github.com/kubernetes-incubator/external-storage/issues/165

wongma7: So here's the steps i would do:

import kubernetes/pkg/volume/util
call util.Fsinfo where GetFsAvailableByte was called before
run glide up -v
run glide-vc --use-lock-file

ianchakeres: @wongma7 - thanks for the advice. I was able to execute the commands.

PTAL and let me know if the additional files in this commit look good.

ianchakeres:
@msau42 @ddysher @wongma7 - After executing glide commands, I was able to use FsInfo directly. PTAL

If we agree this code looks good, I'll refactor the further (by calling FsInfo directly inside discover.go, currently stashed) and squash some of the commits.

msau42:
This is a lot of files to pull into vendor...  Should we try moving FsInfo to a higher level package so we can pull in from client-go instead?

wongma7:
What I got is it won't be accepted if we try to put the code into client-go when the code will need to be moved to k8s.io/pkg anyway. So we are left with either vendoring huge amounts of files *or* copying code as-needed. Either case would be only temporary, ideally. TBH I would prefer we copy code as needed, we are all under the same k8s org&license after all, and it's not like we'll need to manually update the copied code of simple util functions frequently if at all.

And FWIW, from an initial attempt, refactoring pkg/volume/util is near impossible anyway atm, it is heavily tied to pkg/volume and idk how it will be made consumable for k8s.io/pkg or anywhere...

msau42:
Ah sorry, I instead of client-go, I meant putting it into k8s.io/pkg.  Instead of pulling all of pkg/volume/util, can we just pull out the pieces we need for now?

ianchakeres:
@msau42 @wongma7 @ddysher - To handle the immediate issue with Frsize, should we go ahead with the current commit direction (aka pull dependencies into vendor) or revert back to leveraging the unix.Statfs directly?

wongma7:
I'm okay with current direction, don't feel strongly either way. When k8s.io/pkg becomes a thing we can simply change a few import lines, run glide again, and all should be well. And I'd hate for your glide troubles to have been for nothing :)

msau42:
Ok sounds good, let's continue with the vendoring direction.

ianchakeres:
@msau42 & @wongma7 ok - I will proceed with vendoring direction.

ianchakeres:
@msau42 @ddysher @wongma7 - PTAL. I squashed my commits and left the vendor commit separate.

msau42:
/lgtm

@wongma7 can you check out the vendoring changes?

ddysher:
LGTM :+1:

PR #231, opened 2017-07-12
fix cephfs deployment

### LINKED ISSUE #217: CephFS deployment yaml example does not work
https://github.com/kubernetes-incubator/external-storage/blob/master/ceph/cephfs/deployment.yaml needs an entrypoint. Or Dockerfile needs to be modified so that its CMD isn't just chmod

rootfs:
fix #217


wongma7:
do we need also "-id=cephfs-provisioner-1" arg to match the README `docker run`?

rootfs:
@wongma7 done

Issue #230, opened 2017-07-12
Vendored package needs updating


dhirajh:
Requesting the update of the package golang.org/x/sys/unix which has been vendored. The update has the BLK* constants needed for some of the block device work happening in local volume.

wongma7:
Sounds good to me! can run `glide up -v` and `glide-vc --use-lock-file` to pick up updates

wongma7:
closing; I'm going to be updating docs on the way we are doing vendoring, has come up a lot recently, and could maybe explore a better/easier way.

PR #221, opened 2017-07-06
Incorporating K8s api changes

kokhang:
Incorporating k8s API changes.
- Updating client-go and apimachinery version
- Importing K8S types from k8s.io/api
- Removing lib/helper and importing the kubernetes one instead temporary until helper go to k8s.io/util

PR #232, opened 2017-07-13
Update vendored packages

dhirajh:
This pull request represents the changes from a "glide update -v". It was done primarily to get updated version of golang.org/x/sys/unix and the BLK* constants introduced in it. This addresses issue #230 

@wongma7


dhirajh:
Please re-check for cla.

wongma7:
Thanks; I'm not 100% sure but I think https://github.com/kubernetes-incubator/external-storage/pull/221 may have inadvertently updated golang.org/x/sys/unix just now! Can you check if this is the case?

dhirajh:
@wongma7 Yes, it looks like #221 has updated the desired files. I will close this PR.

PR #237, opened 2017-07-13
client-go integration strategy

pospispa:
client-go integration strategy of this library is not documented.

That's why the documentation is added.

In addition, Go version is increased to 1.8 because currently used version of client-go requires Go 1.8.


wongma7:
Thanks. docs/README.md in general is quite outdated, I will need to overhaul it soon.

PR #240, opened 2017-07-13
Script for Running All Unit Tests Except Blacklisted Ones

pospispa:
It's convenient for local development to have a script that runs all unit tests.

That's why such script is added. Unfortunately, some unit tests require installation of additional packages into the vendor/ directory. Such unit tests are blacklisted.

### DIFF
--- a/unittests.sh
+++ b/unittests.sh
@@ -1,0 +1,37 @@
+#!/bin/bash
+
+# Copyright 2017 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


wongma7:
lgtm but we should put it somewhere other than repo-infra* because that should stay a "clean" subtree of https://github.com/kubernetes/repo-infra. Root of repo is fine imo, we don't have too much stuff there I don't think...

pospispa:
OK, I've moved it into the repo root.

pospispa:
@wongma7 please, is it possible to merge it?

Issue #241, opened 2017-07-13
FAQ: For aws/efs


shadycuz:
I'm new to kubernetes and was able to set up EFS on my pods very easily, how ever I do have some questions. I will be updating the aws/efs documentation in an upcoming PR and this will be a FAQ at the end. Some of the questions might stem from my lack of knowledge with kubernetes.

### [configmap.yaml](https://github.com/kubernetes-incubator/external-storage/blob/master/aws/efs/deploy/configmap.yaml)

1. I have witnessed how handy a configmap can be with prometheus, but why here? Couldn't those values have been hard coded into the deployment file? Or does the abstraction serve another purpose I am missing?

### [deployment.yaml](https://github.com/kubernetes-incubator/external-storage/blob/master/aws/efs/deploy/deployment.yaml)

2. I noticed the AWS EFS gets mounted to the efs-provisioner container directly. What happens when that pod goes down?

3. Why shouldn't we just mount the EFS directly to our apps? Why use the PVC's?

4. Can you run multiple pods for HA? Will that work?

5. What is best practice? Should I use multiple claims on 1 EFS to serve all my pods? Or multiple EFS drives with multiple efs-provisioner deployments/configmaps to server seperate pods?

6. Can I put my files in the root of the EFS?

7. Can I automate making the `/persistentvolumes` directory some how?

### [class.yaml](https://github.com/kubernetes-incubator/external-storage/blob/master/aws/efs/deploy/class.yaml)

On some guides on the internet using PV's that come out of the box, I noticed some people would create a claim directly to the PV with out a StorageClass.

8. Can you create a claim directly to an EFS PV with out a class?

### [claim.yaml](https://github.com/kubernetes-incubator/external-storage/blob/master/aws/efs/deploy/claim.yaml)

I have a claim of 5Mi but I wrote 20Gb's to it.

9. Do claims actually limit the storage the pod can use? or is that not functional with EFS?

wongma7:
1. No particular reason, tbh. But it's arguably easier to tell users to 'make this configMap' with a single kubectl command than it is to say 'edit these x lines of deployment.yaml'
2. PVs will not be provisioned nor deleted while the provisioner is down. Existing PVs will continue to work though since they're pointing straight at EFS. When the provisioner comes back, it will delete PVs that need to be deleted in Released state, and create PVs for PVCs that are Pending.
[...]

PR #246, opened 2017-07-17
Readme 2.0

shadycuz:


aws/efs/README.md
```diff
@@ -5,9 +5,109 @@
...
+      - name: efs-pvc
+        mountPath: "/mnt"
+  restartPolicy: "Never"
+  volumes:
+    - name: efs-pvc
+      persistentVolumeClaim:
+        claimName: efs
+```
+
+Because our claim is ReadWriteMany if you scale this pod each aditional pod will also be able to read and write the same files. You may also reference the same claimName in another type of pod so your 2 applications can read and write the same files. If you wish to have a second application that uses EFS storage but don't want other pods to access the files, create a new claim using a new name but the same storage class.
```
shadycuz: Not sure if this is completely accurate.

aws/efs/README.md
```diff
@@ -134,3 +235,37 @@ Add a reference to the secret in the deployment yaml.
...
+
+It's not needed but it is helpful if you are going to use your EFS for other things than Kubernetes. In your efs-provisioner deployment set your path to `/`
+
+- I noticed when creating the claim it has request for a really small ammount of storage?
+
+The storage section size is a requirment because most other PersistentVolumes need it. Every pod accessing EFS will have unlimited storage. I use 1Mi to remind my self it's unlimited.
+
+- Can I omit that part of the claim?
+
+Good question...
```
shadycuz: Wasn't sure

shadycuz:
Oh I and I didn't test my manifest file or some of the assumptions I made. I will test tomorrow.

aws/efs/README.md
```diff
@@ -5,9 +5,109 @@
...
+  restartPolicy: "Never"
+  volumes:
+    - name: efs-pvc
+      persistentVolumeClaim:
+        claimName: efs
+```
+
+Because our claim is ReadWriteMany if you scale this pod each aditional pod will also be able to read and write the same files. You may also reference the same claimName in another type of pod so your 2 applications can read and write the same files. If you wish to have a second application that uses EFS storage but don't want other pods to access the files, create a new claim using a new name but the same storage class.
+
+Some times you want the replica pods to be on EFS but you do not wish them to share the same files. In those situations it's best to use a [StatefulSet](./https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/) with the setting ReadWriteOnce. When a StatefulSet scales up it will dynamicly create new claims for your pods.
```
wongma7: RWO and RWM are just labels, they don't actually enforce anything at the moment. Even if you ask for RWO, the RWO PV can still be read/written to from multiple pods if multiple pods consume the PVC, kubernetes won't care.

wongma7: For that reason I would avoid mentioning RWO and RWM in the line above, it is just confusing IMO. EFS is NFS which has the property of being RWM, I would leave it at that

aws/efs/README.md
```diff
@@ -5,9 +5,109 @@
...
+        volume.beta.kubernetes.io/storage-class: aws-efs
+    spec:
+      accessModes: [ "ReadWriteOnce" ]
+      resources:
+        requests:
+          storage: 1Mi
+```
+Note: We do not reference a claim name, instead we give the new claim the name efs and we list the StorageClass we wish to use when creating the claim. I like to use the type of storage, because the name of the pod will be be added to it. This example when ran will create the claim `efs-web-0` and scaling the pod up to 3 will create `efs-web-1` and `efs-web-2`.
+
+If you wish to learn more about efs-provisioner and how to change additional settings you contine on in the deployment section. You might also want to check the FAQ at the bottom.
```
wongma7: ...settings, continue on...

aws/efs/README.md
wongma7: I would just remove this 'Because our claim is ReadWriteMany...', see below comments

aws/efs/README.md
```diff
@@ -5,9 +5,109 @@
 quay.io/external_storage/efs-provisioner:v0.1.0
 ```
 
+The efs-provisioner is a storage plugin for kubernetes. It consists of a container that has access to an AWS [EFS](https://aws.amazon.com/efs/) resource. The container reads a configmap which contains the EFS filesystem ID, the AWS region and the name you want to use for your efs-provisioner. This name will be used later when you create a storage class.
```
wongma7: storage plugin isn't right, at best this can be called a subset of a storage plugin. storage plugin implies a full flex volume plugin or container storage interface implementation which are much more than this.

shadycuz:
@wongma7 I will make the changes as soon as I can.

shadycuz:
@wongma7 I have fixed the issues you found and have changed the manifest and updated the getting started section to no longer need an existing directory thanks to [#261(comment) ](https://github.com/kubernetes-incubator/external-storage/issues/241#issuecomment-317336524).

PR #248, opened 2017-07-18
Initial changes to provisioner for blocks storage.

dhirajh:
@msau42 @ddysher Here are the block related changes at this time for your review. It simulates block discovery, but will not actually be effective until block metadata is in place with PVs. The unit tests are yet to be addressed, will do so after feedback on the implementation.

### DIFF
--- a/local-volume/provisioner/deployment/docker/Dockerfile
+++ b/local-volume/provisioner/deployment/docker/Dockerfile
@@ -12,6 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM scratch
+FROM busybox
 ADD local-volume-provisioner /local-provisioner
 ENTRYPOINT ["/local-provisioner"]
--- a/local-volume/provisioner/pkg/common/common.go
+++ b/local-volume/provisioner/pkg/common/common.go
@@ -29,7 +29,6 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/kubernetes/pkg/kubelet/apis"
-	"strings"
 )
 
 const (
@@ -51,8 +50,6 @@ const (
 	// EventVolumeFailedDelete copied from k8s.io/kubernetes/pkg/controller/volume/events
 	EventVolumeFailedDelete = "VolumeFailedDelete"
 
-	// DefaultVolumeType is default type of volume being discovered in the specified host dir.
-	DefaultVolumeType = VolumeTypeFile
 )
 
 // UserConfig stores all the user-defined parameters to the provisioner
@@ -69,8 +66,6 @@ type MountConfig struct {
 	HostDir string `json:"hostDir"`
 	// The mount point of the hostpath volume
 	MountDir string `json:"mountDir"`
-	// The volumeType "block" or "file". If nil, the type defaults to "file"
-	VolumeType string `json:"volumeType"`
 }
 
 // RuntimeConfig stores all the objects that the provisioner needs to run
@@ -141,9 +136,8 @@ func GetVolumeConfigFromConfigMap(client *kubernetes.Clientset, namespace, name 
 func GetDefaultVolumeConfig() map[string]MountConfig {
 	return map[string]MountConfig{
 		"local-storage": {
-			HostDir:    DefaultHostDir,
-			MountDir:   DefaultMountDir,
-			VolumeType: DefaultVolumeType,
+			HostDir:  DefaultHostDir,
+			MountDir: DefaultMountDir,
 		},
 	}
 }
@@ -169,17 +163,6 @@ func ConfigMapDataToVolumeConfig(data map[string]string) (map[string]MountConfig
 		config := MountConfig{}
 		if err := json.Unmarshal([]byte(val), &config); err != nil {
 			return nil, fmt.Errorf("unable to unmarshal config for class %v: %v", class, err)
-		}
-		if config.VolumeType != "" {
-			config.VolumeType = strings.ToLower(config.VolumeType)
-		} else {
-			config.VolumeType = DefaultVolumeType
-		}
-		switch config.VolumeType {
-		case VolumeTypeFile:
-		case VolumeTypeBlock:
-		default:
-			return nil, fmt.Errorf("Invalid volume type %v for class %v", config.VolumeType, class)
 		}
 		mountConfig[class] = config
 	}
--- a/local-volume/provisioner/pkg/common/common.go
+++ b/local-volume/provisioner/pkg/common/common.go
@@ -70,7 +70,7 @@ type MountConfig struct {
 	// The mount point of the hostpath volume
 	MountDir string `json:"mountDir"`
 	// The volumeType "block" or "file". If nil, the type defaults to "file"
-	VolumeType string `json:"volumeType""`
+	VolumeType string `json:"volumeType"`
 }
 
 // RuntimeConfig stores all the objects that the provisioner needs to run
@@ -141,8 +141,8 @@ func GetVolumeConfigFromConfigMap(client *kubernetes.Clientset, namespace, name 
 func GetDefaultVolumeConfig() map[string]MountConfig {
 	return map[string]MountConfig{
 		"local-storage": {
-			HostDir:  DefaultHostDir,
-			MountDir: DefaultMountDir,
+			HostDir:    DefaultHostDir,
+			MountDir:   DefaultMountDir,
 			VolumeType: DefaultVolumeType,
 		},
 	}
--- a/local-volume/provisioner/pkg/common/common.go
+++ b/local-volume/provisioner/pkg/common/common.go
@@ -29,6 +29,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/kubernetes/pkg/kubelet/apis"
+	"strings"
 )
 
 const (
@@ -37,6 +38,10 @@ const (
 	// NodeLabelKey is the label key that this provisioner uses for PV node affinity
 	// hostname is not the best choice, but it's what pod and node affinity also use
 	NodeLabelKey = apis.LabelHostname
+	// VolumeTypeFile represents file type volumes
+	VolumeTypeFile = "file"
+	// VolumeTypeBlock represents block type volumes
+	VolumeTypeBlock = "block"
 
 	// DefaultHostDir is the default host dir to discover local volumes.
 	DefaultHostDir = "/mnt/disks"
@@ -45,6 +50,9 @@ const (
 
 	// EventVolumeFailedDelete copied from k8s.io/kubernetes/pkg/controller/volume/events
 	EventVolumeFailedDelete = "VolumeFailedDelete"
+
+	// DefaultVolumeType is default type of volume being discovered in the specified host dir.
+	DefaultVolumeType = VolumeTypeFile
 )
 
 // UserConfig stores all the user-defined parameters to the provisioner
@@ -61,6 +69,8 @@ type MountConfig struct {
 	HostDir string `json:"hostDir"`
 	// The mount point of the hostpath volume
 	MountDir string `json:"mountDir"`
+	// The volumeType "block" or "file". If nil, the type defaults to "file"
+	VolumeType string `json:"volumeType""`
 }
 
 // RuntimeConfig stores all the objects that the provisioner needs to run
@@ -133,6 +143,7 @@ func GetDefaultVolumeConfig() map[string]MountConfig {
 		"local-storage": {
 			HostDir:  DefaultHostDir,
 			MountDir: DefaultMountDir,
+			VolumeType: DefaultVolumeType,
 		},
 	}
 }
@@ -158,6 +169,17 @@ func ConfigMapDataToVolumeConfig(data map[string]string) (map[string]MountConfig
 		config := MountConfig{}
 		if err := json.Unmarshal([]byte(val), &config); err != nil {
 			return nil, fmt.Errorf("unable to unmarshal config for class %v: %v", class, err)
+		}
+		if config.VolumeType != "" {
+			config.VolumeType = strings.ToLower(config.VolumeType)
+		} else {
+			config.VolumeType = DefaultVolumeType
+		}
+		switch config.VolumeType {
+		case VolumeTypeFile:
+		case VolumeTypeBlock:
+		default:
+			return nil, fmt.Errorf("Invalid volume type %v for class %v", config.VolumeType, class)
 		}
 		mountConfig[class] = config
 	}
--- a/local-volume/provisioner/pkg/common/common.go
+++ b/local-volume/provisioner/pkg/common/common.go
@@ -49,7 +49,6 @@ const (
 
 	// EventVolumeFailedDelete copied from k8s.io/kubernetes/pkg/controller/volume/events
 	EventVolumeFailedDelete = "VolumeFailedDelete"
-
 )
 
 // UserConfig stores all the user-defined parameters to the provisioner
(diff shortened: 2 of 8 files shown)


msau42:
You could actually start working on the block scrubbing too.  This will create the PV and you can create a PVC to bind to.  Just don't create a pod yet.  But you can then delete the PVC and trigger the PV cleanup logic.

local-volume/provisioner/pkg/common/common.go
```diff
@@ -36,11 +37,17 @@ const (
...
+	VolumeTypeFile = "file"
+	// VolumeTypeBlock represents block type volumes
+	VolumeTypeBlock = "block"
 
 	// DefaultHostDir is the default host dir to discover local volumes.
 	DefaultHostDir = "/mnt/disks"
 	// DefaultMountDir is the container mount point for the default host dir.
 	DefaultMountDir = "/local-disks"
+	// DefaultVolumeType is default type of volume being discovered in the specified host dir.
+	DefaultVolumeType = VolumeTypeFile
```
dhirajh: @ianchakeres GetDefaultVolumeConfig() is only invoked if no configmap has been supplied. This default value is used when a configmap has been used and a volumeType has not been specified. It basically indicates that the default setting for volumetype when not specified is file.

local-volume/provisioner/pkg/common/common.go
```diff
@@ -57,6 +64,8 @@ type MountConfig struct {
 	HostDir string `json:"hostDir"`
 	// The mount point of the hostpath volume
 	MountDir string `json:"mountDir"`
+	// The volumeType "block" or "file". If nil, the type defaults to "file"
```
dhirajh: OK, will update the comment.

local-volume/provisioner/pkg/util/volume_util.go
```diff
@@ -66,6 +73,18 @@ func (u *volumeUtil) IsDir(fullPath string) (bool, error) {
 	return stat.IsDir(), nil
 }
 
+// IsBlock checks if the given path is a block device
+func (u *volumeUtil) IsBlock(fullPath string) (bool, error) {
+	var st unix.Stat_t
+	err := unix.Stat(fullPath, &st)
```
dhirajh: @msau42  Yes, this was tested using symlinks. It follows the symlinks.

local-volume/provisioner/pkg/util/volume_util.go
```diff
@@ -116,6 +135,23 @@ func (u *volumeUtil) GetFsAvailableByte(fullPath string) (uint64, error) {
...
+// fullPath is the pathname of block device.
+func (u *volumeUtil) GetBlockSpaceByte(fullPath string) (uint64, error) {
+	file, err := os.OpenFile(fullPath, os.O_RDWR, 0)
+	if err != nil {
+		return 0, err
+	}
+	defer file.Close()
+
+	var size int64
+	if _, _, err := unix.Syscall(unix.SYS_IOCTL, file.Fd(), unix.BLKGETSIZE64, uintptr(unsafe.Pointer(&size))); err != 0 {
```
dhirajh: It fetches the size of block device in bytes into a 64 bit value (see http://www.microhowto.info/howto/get_the_size_of_a_linux_block_special_device_in_c.html).

local-volume/provisioner/pkg/util/volume_util.go
```diff
@@ -116,6 +135,23 @@ func (u *volumeUtil) GetFsAvailableByte(fullPath string) (uint64, error) {
 	return uint64(s.Frsize) * (s.Blocks - s.Bfree + s.Bavail), nil
 }
 
+// GetBlockSpaceByte returns  capacity in bytes of a block device.
+// fullPath is the pathname of block device.
+func (u *volumeUtil) GetBlockSpaceByte(fullPath string) (uint64, error) {
+	file, err := os.OpenFile(fullPath, os.O_RDWR, 0)
```
dhirajh: Yes, that is correct. I will fix that issue.

local-volume/provisioner/pkg/deleter/deleter.go
```diff
@@ -74,17 +74,23 @@ func (d *Deleter) cleanupPV(pv *v1.PersistentVolume) error {
...
+		return fmt.Errorf("Unknown storage class name %v", pv.Spec.StorageClassName)
 	}
 
-	specPath := pv.Spec.Local.Path
-	relativePath, err := filepath.Rel(config.HostDir, specPath)
-	if err != nil {
-		return fmt.Errorf("Could not get relative path: %v", err)
-	}
+	if config.VolumeType == common.VolumeTypeFile {
+		specPath := pv.Spec.Local.Path
```
dhirajh: OK, will do.

local-volume/provisioner/pkg/deleter/deleter.go
```diff
@@ -74,17 +74,23 @@ func (d *Deleter) cleanupPV(pv *v1.PersistentVolume) error {
...
+
+		glog.Infof("Deleting PV %q contents at hostpath %q, mountpath %q", pv.Name, specPath, mountPath)
+		err = d.VolUtil.DeleteContents(mountPath)
+		return err
+	}
 
-	glog.Infof("Deleting PV %q contents at hostpath %q, mountpath %q", pv.Name, specPath, mountPath)
-	return d.VolUtil.DeleteContents(mountPath)
+	// TODO: block device
+	return nil
```
dhirajh: OK, will do.

local-volume/provisioner/pkg/discovery/discovery.go
```diff
@@ -100,15 +100,36 @@ func (d *Discoverer) discoverVolumesAtPath(class string, config common.MountConf
...
-			glog.Errorf("Mount path %q validation failed: %v", filePath, err)
-			continue
-		}
-		availByte, err := d.VolUtil.GetFsAvailableByte(filePath)
-		if err != nil {
-			glog.Errorf("Path %q fs stats error: %v", filePath, err)
-			continue
+		var availByte uint64
+		if config.VolumeType == common.VolumeTypeBlock {
+			var ret bool
```
dhirajh: OK, will do.

local-volume/provisioner/pkg/discovery/discovery.go
```diff
@@ -100,15 +100,36 @@ func (d *Discoverer) discoverVolumesAtPath(class string, config common.MountConf
...
+			ret, err = d.VolUtil.IsBlock(filePath)
```
dhirajh: @msau42 How about I follow the same pattern as the validateFile() method and create a validateBlock()  method here? It will return just an err object as outlined by you. I would like to keep the IsBlock() function with its lower level interface that distinguishes a request failure from boolean return value as it provides more flexibility to the caller.

local-volume/provisioner/pkg/discovery/discovery.go
```diff
@@ -100,15 +100,35 @@ func (d *Discoverer) discoverVolumesAtPath(class string, config common.MountConf
...
+				glog.Errorf("%q is not Block device: %v", filePath)
+				continue
+			}
+
+			availByte, err = d.VolUtil.GetBlockSpaceByte(filePath)
+			if err != nil {
+				glog.Errorf("Path %q block stats error: %v", filePath, err)
+				continue
+			}
+		} else {
```
dhirajh: OK will do.

local-volume/provisioner/pkg/common/common.go
```diff
@@ -57,6 +64,8 @@ type MountConfig struct {
...
+	VolumeType string `json:"volumeType"`
```
dhirajh: @msau42 I actually went back and forth on this one. The question is what should we do if a discovery directory has a mix of block sym links and normal files? Should we expect users to keep a pristine discovery directory or else error out? By having an explicit declaration on what the directory should contain, we can discover what the user intended and ignore any entries that don't match that declaration. It seemed like a more forgiving process. Do let me know your thoughts on this, as I was on the fence about this.

msau42: It is nice that this can catch user errors, although I imagine, it would be difficult to make a mistake because as an admin you would need to do two entirely different commands?  For block devices, you need to symlink the device into the directory.  For fs volumes, you need to run a mount command into the directory.

With auto-detection, we could support a mix of both in the same directory (and storage class)

local-volume/provisioner/pkg/discovery/discovery.go
msau42: sounds good

local-volume/provisioner/pkg/common/common.go
dhirajh: @msau42 Does that mean a given class of storage can consist of both file PVs and block PVs? It seemed like in last week's discussion the SIG was leaning towards not having volumeType in PVC (or did I misunderstand that?). In that case, wouldn't specifying a PVC with a specific storage class give the POD either a block device or file system, depending on which one is available?

msau42: The discussion last week was to not have fstype in the PVC, and that local storage wouldn't support the file-on-block case for now.  But I think volume type still has to be specified, so you can only bind a PVC with volume type matching on the PV.

local-volume/provisioner/pkg/deleter/deleter.go
```diff
@@ -86,5 +99,6 @@ func (d *Deleter) cleanupPV(pv *v1.PersistentVolume) error {
 	mountPath := filepath.Join(config.MountDir, relativePath)
 
 	glog.Infof("Deleting PV %q contents at hostpath %q, mountpath %q", pv.Name, specPath, mountPath)
-	return d.VolUtil.DeleteContents(mountPath)
+	err = d.VolUtil.DeleteContents(mountPath)
```
msau42: You can return directly here

dhirajh:
@msau42 Yes, currently working on unit test. I forgot that the active PR was going reflect any push I make to my git repo.

local-volume/provisioner/pkg/util/volume_util.go
ianchakeres: It would be nice to plop this info into a comment.

dhirajh:
@msau42 Here are the updates to the PR:

1. Vendor packages seem to have regressed by some other check-in into main. Ran into this issue as I rebased from main. Bringing vendor packages back upto date.  
2. Added unit testing for block devices.
3. Had to modify bootstrapper code as it was launching provisioner in non-privileged mode, which does not let it look at block devices via symlinks
4. Addressed last couple of comments to the PR.
5. Modified isDir/isBlock signature as returning an error seemed like a better approach as the functions handle both file and block paths.

local-volume/provisioner/pkg/util/volume_util.go
```diff
@@ -45,7 +46,7 @@ type VolumeUtil interface {
 	GetFsCapacityByte(fullPath string) (int64, error)
 
 	// Get capacity of the block device
-	GetBlockSpaceByte(fullPath string) (uint64, error)
+	GetBlockSpaceByte(fullPath string) (int64, error)
```
ianchakeres: I'd recommend GetBlockCapacityByte for this method. This name is consistent with GetFsCapacityByte.

local-volume/provisioner/pkg/util/volume_util.go
```diff
@@ -58,30 +59,38 @@ func NewVolumeUtil() VolumeUtil {
 }
 
 // IsDir checks if the given path is a directory
-func (u *volumeUtil) IsDir(fullPath string) (bool, error) {
+func (u *volumeUtil) IsDir(fullPath string) error {
```
ianchakeres: I'd be more inclined to return bool here. If passing errors up is required, then pass both bool and error. I looked at some other code, and this seems to be there standard for k8s.

If we want the reason why the method is returning false, the reason could be logged.

local-volume/provisioner/pkg/util/volume_util.go
dhirajh: OK, will change.

local-volume/provisioner/pkg/util/volume_util.go
dhirajh: Had a longer offline chat with Ian about this. My concern was that the function returns err in times when a boolean was preferable, so going with one return type seemed simpler. Having  said that, keeping in mind the k8s standards, I will change it back to its original signature.

msau42:
Can you keep the vendoring changes separate, so that it's easier to review.  It may be even better to have the vendoring changes as part of a different PR

wongma7:
don't forget to run `glide-vc --use-lock-file` , could explain why the vendor changes are huger than usual

dhirajh:
@wongma7 @msau42 I have removed vendor changes from this PR and limited it to local-volume provisioner changes. Fortunately, it looks like the merge into main of  raffaelespazzoliorg/iscsi-targetd2 has fixed the regression that happened in vendor folder, so I don't need to update it anymore. Please let me know if any other changes are needed. Thanks!

local-volume/provisioner/deployment/docker/Dockerfile
```diff
@@ -12,6 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM scratch
+FROM busybox
```
msau42: Any reason to make it busybox?

local-volume/provisioner/pkg/util/volume_util.go
```diff
@@ -186,15 +197,31 @@ func (u *FakeVolumeUtil) IsDir(fullPath string) (bool, error) {
...
+	dir, file := filepath.Split(fullPath)
+	dir = filepath.Clean(dir)
+	files, found := u.directoryFiles[dir]
+	if !found {
+		return false, fmt.Errorf("Directory %q not found", dir)
+	}
+
+	for _, f := range files {
+		if file == f.Name {
+			return f.VolumeType == FakeEntryBlock, nil
```
msau42: How come `IsFile` returns an err, but this doesn't?

local-volume/provisioner/pkg/util/volume_util.go
```diff
@@ -131,37 +131,48 @@ func (u *volumeUtil) GetFsCapacityByte(fullPath string) (int64, error) {
...
-type FakeFile struct {
-	Name     string
-	IsNotDir bool
+const (
+	// FakeEntryFile is mock dir entry of type file.
+	FakeEntryFile = "file"
+	// FakeEntryBlock is mock dir entry of type block.
+	FakeEntryBlock = "block"
+	// FakeEntryUnknown is mock dir entry of type unknown.
+	FakeEntryUnknown = "unknown"
```
msau42: Can you also add a test case for unknown type?

local-volume/provisioner/deployment/docker/Dockerfile
dhirajh: @msau42 I used busybox because it allows a certain level of debug-ability as you can ssh into the container and use some basic linux utils to see what the container is viewing on a host. Seems useful for a service that is yet to mature. I can remove it if thats not something we want.

msau42: That's fine, those are good reasons :-)

local-volume/provisioner/pkg/util/volume_util.go
dhirajh: @msau42 : The reason that I return an err in isDir() is because I found the real code exiting with an err when the path was not a dir or file (because the OS responds with an error when open is called on a non dir/file). I decided to simulate this accurately in the mock (which it was not doing earlier) and hence an error is returned. This behavior is not seen by isBlock() and hence does not need to be mocked.

local-volume/provisioner/pkg/util/volume_util.go
dhirajh: OK, will do.

dhirajh: @msau42 I just realized that this is already covered in the unit test TestDiscoverVolumes_BadVolume. Let me know if any other changes are needed.

PR #252, opened 2017-07-24
Removed unreachable code

pospispa:
@rootfs I noticed some unreachable code in the CephFS provisioner so IMO, it's good to remove it.


rootfs:
lgmt

PR #254, opened 2017-07-24
Annotation Format for Storing Custom Data in Persistent Volume

pospispa:
External provisioners may need to store custom data in annotations in Persistent Volume.
Recommended format for annotations is added to documentation.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -106,6 +106,9 @@ Now, actually giving provisioners identities and effectively making them pets ma
 
 This repository is structured such that each external provisioner gets its own directory for its code, docs, examples, yamls, etc. What they don't get is individual "vendor" directories for their respective dependencies, they must depend on the shared top-level vendor and lib directories. This helps reduce the size of the repo and forces all parts of it to stay updated, but introduces some complications for contributors.
 
+### Conventions
+[Kubernetes project](https://github.com/kubernetes/kubernetes/) conventions are followed if not otherwise stated.
+
 ### Adding a provisioner
 
 Basically you create a directory to house everything you want to check in, add build and/or test invocations to [travis](../.travis.yml), and add dependencies to the top-level vendor directory.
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -121,3 +124,15 @@ Any breaking update to a vendor dependency requires an update to every external
 Generally, breaking vendor dependency updates won't happen often (at least every time kubernetes/client-go updates, maybe) and all the provisioners can be updated with ease, without requiring explicit approval from their respective OWNERS, unless the change is big enough or they've asked that it be required.
 
 As the contributor of a dependency/library update, you're usually responsible for updating the dependents so travis CI passes, as it shouldn't be harder than a find/replace. Otherwise, if it's decided that you don't need to be responsible, some other solution will be worked out to make sure everything stays in a buildable state.
+
+### Using Persistent Volume annotations
+External provisioners may need to store custom data in Persistent Volume annotations. An annotation shall have the below format:
```
wongma7: it would be good to start recommending this convention now, but it's not backwards compatible and IMO it should not be a requirement, so "shall" is too strong a word

docs/README.md
```diff
@@ -121,3 +124,15 @@ Any breaking update to a vendor dependency requires an update to every external
...
+```
+annotations:
+  <provisioner-type>.external-storage.incubator.kubernetes.io/<variable> : <value>
```
wongma7: idk if we are allowed to use kubernetes.io. Technically we are allowed to write any annotations we want as long as they don't contain that official kubernetes string reserved by the system

pospispa: I discussed the string `...kubernetes.io` with @jsafrane.
@jsafrane please, what caused that you believe we can use the string `...kubernetes.io`

docs/README.md
pospispa: @wongma7 so you prefer to use "should" instead of "shall", don't you?

wongma7: yea

docs/README.md
wongma7: ok, we are probably safe :) we are surely official enough to use it

PR #255, opened 2017-07-24
[local-volume] Changed Capacity for local PVs to use FsInfo capacity

ianchakeres:
Earlier version used "available" from FsInfo
New version uses "capacity" from FsInfo
I also changed uint64 to int64


ianchakeres:
@msau42 @ddysher - PTAL

PR #264, opened 2017-07-29
Support filesystems that always report zero available blocks, for example RAM FS.

Monnoroch:
Based on https://chromium.googlesource.com/chromium/src/base/+/master/sys_info_posix.cc#98.


Monnoroch:
An image for the change can be found here: https://hub.docker.com/r/monnoroch/nfs-provisioner/tags.

wongma7:
lgtm, need CLA though

wongma7:
closing, this is a good change but could not merge without cla sign

PR #266, opened 2017-07-31
Add glusterfs simple provisioner

yuanying:
Actually `Heketi` is a great gluster volume provisioner for k8s, but I want to manage
gluster volume more simple way.

GlusterFS simple provisioner is an external provisioner which
dynamically provision glusterfs volumes on demand.

Unlike `Heketi`, this provisioner will not manage Gluster cluster.
So that you must manage Gluster cluster by yourself. This will simply
create Gluster volume in the specified Gluster cluster like nfs-client
provisioner.

It means, for example, if you want to add brick to your Gluster volume,
you can use familiar `gluster vol add-brick` command.

Currently this provision only support Gluster cluster which is running 
on the top of Kubernetes cluster.


humblec:
@yuanying  Nice attempt. However Can you update the PR with more details about this provisioner ?

yuanying:
@humblec Hi, thank you for your comment. I'll update commit message.

gluster/fs/pkg/allocator/minmax.go
```diff
@@ -0,0 +1,191 @@
+/*
```
wongma7: is this the same as aws/efs allocator or copied from upstream allocator? (either way it's fine, I will be putting the allocator in `lib` soon)

yuanying: Yes, these were copied from aws/efs allocator. I hope these should be in `lib` directory, nice!

gluster/fs/pkg/volume/provision.go
```diff
@@ -0,0 +1,278 @@
...
+			glog.V(3).Infof("glusterfs: dynamic ep %v and svc : %v ", endpoint, service)
+			return &v1.GlusterfsVolumeSource{
+				EndpointsName: endpoint.Name,
+				Path:          cfg.VolumeName,
+				ReadOnly:      false,
+			}, nil
+		}
+	}
+
+	// XXX: Must not delete volume if creating volume is failed
```
wongma7: curious, why not?

gluster/fs/pkg/volume/delete.go
```diff
@@ -0,0 +1,135 @@
...
+		glog.Errorf("glusterfs: error to release GID: %v", err)
+	}
+
+	return nil
+}
+
+func (p *glusterfsProvisioner) deleteVolume(
+	namespace string, name string, cfg *ProvisionerConfig) {
+
+	if cfg.DeletionPolicy == "delete" {
```
wongma7: IMO we can't do this, this goes against the meaning of the `Delete` reclaim policy on the PV. `Delete` must mean `delete PVC -> delete PV's storage asset -> delete PV (+endpoint+service)`. If the admin configures DeletionPolicy != "delete" then it becomes `delete PVC -> delete PV (+endpoint+service)` which may go against user's expectations, i.e. they expect their data to not be accessible by anyone anymore.

The alternative is to set `Retain` on the PV according to a storage class parameter, in which case Delete() will never be called on the PV and the PV will simply go to the `Released` state. https://github.com/kubernetes/kubernetes/pull/47987 will be coming in 1.8, so there could be potential conflicts between the paramater and storageClass.reclaimpolicy to resolve if we go this route.

gluster/fs/README.md
```diff
@@ -0,0 +1,95 @@
+# GlusterFS Simple Provisioner for Kubernetes 1.5+
```
yuanying: ack

gluster/fs/pkg/volume/delete.go
```diff
@@ -0,0 +1,135 @@
...
+	"path/filepath"
+	"strings"
+
+	"github.com/golang/glog"
+	"k8s.io/api/core/v1"
+)
+
+func (p *glusterfsProvisioner) Delete(volume *v1.PersistentVolume) error {
+	var err error
+	class, err := GetClassForVolume(p.client, volume)
```
yuanying: Ack. Anyway I'll change the strategy of getting parameters (which are used for deletion of gluster volume) to use actual volume information which are get from `gluster` command output.

gluster/fs/pkg/volume/provision.go
yuanying: I'll delete this line and uncomment following code. Because these are derived from my old code.

gluster/fs/pkg/volume/delete.go
yuanying: https://github.com/kubernetes/kubernetes/pull/47987 is nice! This is what I want to do. I 'll delete `DeletionPolicy` in my storage class parameter for now.

gluster/glusterfs/README.md
```diff
@@ -0,0 +1,95 @@
...
+    forceCreate: "true"
+    deletionPolicy: "retain"
+```
+
+* `brickRootPaths`: Bricks will be created under this directories.
+* `volumeType`: Storage class will create this type of volume.
+* `namespace`: Namespace which GlusterFS pods are consisted.
+* `selector`: Label selector which will specify GlusterFS pods.
+* `forceCreate`: If true, glusterd create volume forcefully.
+* `deletionPolicy`: This value is used to actually decide whether or not to delete the gluster volume and bricks when `PVC` is deleted. If `delete` is specified, gluster volume and bricks will be deleted. If you want to retain the gluster volume, use `retain`.
```
wongma7: README needs update to remove parameter, else lgtm

yuanying: Done!

PR #267, opened 2017-08-01
RBD provisioner use provisioner name as identitiy by default instead of random string

cofyc:

In production, provisioner should not use random identity on start.

Because, provisioner Pod may crash or be recreated on different nodes, if a provisioner generates random identity string each time, then it will cannot delete old PVs it provisioned before. Example errors logs:

```
I0801 11:03:05.685653    1901 controller.go:1068] scheduleOperation[delete-pvc-e2be4095-76a8-11e7-88c1-000c291fbe71[e2c845d1-76a8-11e7-88c1-000c291fbe71]]
I0801 11:03:05.688484    1901 controller.go:1035] deletion of volume "pvc-e2be4095-76a8-11e7-88c1-000c291fbe71" ignored: ignored because identity annotation on PV does not match ours
...
```

Most users deployed their RBD provisioner without specifying `-id` flag. So I change the default behavior to use provisioner name as identity (like some other provisions, `gluster/block`, etc) instead of random string.

**Special notes for your reviewer**:

In this PR, I also add `PROVISIONER_NAME` environment variable support, like many other provisioners do.

### DIFF
--- a/ceph/rbd/cmd/rbd-provisioner/main.go
+++ b/ceph/rbd/cmd/rbd-provisioner/main.go
@@ -18,11 +18,11 @@ package main
 
 import (
 	"flag"
+	"os"
 
 	"github.com/golang/glog"
 	"github.com/kubernetes-incubator/external-storage/ceph/rbd/pkg/provision"
 	"github.com/kubernetes-incubator/external-storage/lib/controller"
-	"k8s.io/apimachinery/pkg/util/uuid"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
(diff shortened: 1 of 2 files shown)


kfox1111:
Hmmm.... Does id have to be unique within the same type of provisioner?

Say I have a statefulset of ceph rbd provisioners. If I use string for id of 'rbd', then multiple could handle the request, but maybe thats ok? If not, maybe using the statefulset's hostname is a better default?

wongma7:
@kfox1111 id doesn't have to be unique, except in rare cases where only the provisioner that created a PV is capable of deleting it. By giving all provisioners serving the same class the same id (same prName), they will simply race to Delete PVs and everything should turn out fine. 

I'm okay with using prName as the default.

On one hand, it just "makes sense" that the provisioner that created a PV should be responsible for deleting it, so the recommendation made when gluster/block and cephfs were created was to force the user to specify an ID.

On the other, users expect provisioners to behave the same as the upstream mega-controller, i.e. if a PV has Delete it should be deleted, it doesn't matter by which provisioner instance. And maintaining state via an ID argument is painful, plus impossible with a statefulset.

kfox1111:
ok. sounds good to me.

Does the cephfs one have the same issue currently or is it defaulting to hostname?

wongma7:
Yes cephfs has the same issue, it is defaulting to random UUID atm.

PR #285, opened 2017-08-13
Added Makefiles for ceph/rbd and ceph/cephfs and build software for l…

ianchakeres:
I also updated the top-level makefile to leverage new Makefiles


wongma7:
thanks, lgtm!

Issue #284, opened 2017-08-13
[iscsci-provisioner] Need updated roles for Openshift 3.6


woopla:
Hello,

I'm following the [README for the iscsi-targetd provisioner](https://github.com/kubernetes-incubator/external-storage/blob/master/iscsi/targetd/README.md) on an Openshift Origin 3.6.0 install, and I cannot get past the provisioner pod install stage, because all the `system:pv-*` roles have been removed (see this [comment ](https://github.com/openshift/origin/issues/15752#issuecomment-322020633) in the issue I opened with Openshift).

What's the updated way of doing things?

Thanks,

Clément

wongma7:
Please try the new "persistent-volume-provisioner" role. It's still possible it's missing a permission that targetd needs, so if you run into troubles please post the provisoner logs.

woopla:
thanks @wongma7! This seems to be working - I have another unrelated problem (see #288), so it can't be 100% sure.

The README needs an update I guess :)

Cheers,

Clément

PR #290, opened 2017-08-14
If openshift >= 3.6, use new clusterrole because older pv-* ones are gone

wongma7:
fixes https://github.com/kubernetes-incubator/external-storage/issues/284

@raffaelespazzoli fyi.

This should be sufficient, if ugly. I am no expert in ansible.

### DIFF
--- a/iscsi/targetd/README.md
+++ b/iscsi/targetd/README.md
@@ -231,10 +231,11 @@ Run the following commands. The secret correspond to username and password you h
 ```
 oc new-project iscsi-provisioner
 oc create sa iscsi-provisioner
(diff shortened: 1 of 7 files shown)


wongma7:
Actually it's better that we ship our own clusterrole and not rely on system:pv- at all like I did for EFS, NFS, etc.: https://github.com/kubernetes-incubator/external-storage/blob/master/aws/efs/deploy/auth/openshift-clusterrole.yaml . I'll work on it tomorrow.

wongma7:
tested playbook to work on openshift with and without system:persistent-volume-provisioner

PR #311, opened 2017-08-23
Provision method should catch IgnoredError to ignoring the pv creation

weekface:


wongma7:
thanks. this could be used for some kind of sharding right?

weekface:
I used an improved version `hostpath-provisioner` on every kubernetes nodes, every provisioner need to be ignore pv creation not managed by itself.

wongma7:
I see, that’s good enough for me thanks!

PR #312, opened 2017-08-23
Initial check-in of Block Device cleaner

dhirajh:
@msau42 Here are the changes for block cleaner. Couple of things that I am still working on include
1) A sample custom script that could be useful as a reference.
2) Documentation of the changes to configMap for provisioner.

cc @ianchakeres


local-volume/provisioner/pkg/deleter/blockcleaner_test.go
```diff
@@ -0,0 +1,211 @@
...
+	mountConfig.BlockConf.Iterations = 6
+	mountConfig.BlockConf.CustomExec = "/bin/hello"
+	verifyCommandCreated(t, "/container/dir/foo", mountConfig, 6, NoErrExpected, ShellCommand,
+		"-c /bin/hello /container/dir/foo")
+}
+
+func TestBlockCleaner_BasicProcessExec(t *testing.T) {
+	blkCleaner := testBlockCleanerSetup(t, &testConfig{})
+	pv := createPVSpec("testpv", "/mnt/disks", v1.VolumeReleased)
+	err := blkCleaner.asyncDeleteBlockPV(pv, 1, NoErrExpected, ShellCommand, "-c", "echo 1")
```
msau42: Do the tests actually run the actual commands?  If so, do they have the chance to accidentally mess up the host system?

local-volume/provisioner/pkg/deleter/blockcleaner_test.go
```diff
@@ -0,0 +1,211 @@
...
+	"github.com/kubernetes-incubator/external-storage/local-volume/provisioner/pkg/util"
+	"io/ioutil"
+	"k8s.io/api/core/v1"
+	"k8s.io/client-go/tools/record"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
```
msau42: Can you also add a test for duplicate operations?

local-volume/provisioner/pkg/deleter/deleter.go
```diff
@@ -49,27 +50,20 @@ func (d *Deleter) DeletePVs() {
 			glog.Infof("Deleting PV %q", name)
 
 			// Cleanup volume
-			err := d.cleanupPV(pv)
+			err := d.deletePV(pv)
 			if err != nil {
 				cleaningLocalPVErr := fmt.Errorf("Error cleaning PV %q: %v", name, err.Error())
 				d.RuntimeConfig.Recorder.Eventf(pv, v1.EventTypeWarning, common.EventVolumeFailedDelete, cleaningLocalPVErr.Error())
```
msau42: Event type should probably be an error

local-volume/provisioner/pkg/deleter/deleter.go
```diff
@@ -79,27 +73,41 @@ func (d *Deleter) cleanupPV(pv *v1.PersistentVolume) error {
...
+		mountPath)
+	if err := d.VolUtil.DeleteContents(mountPath); err != nil {
+		return err
 	}
 
-	mountPath := filepath.Join(config.MountDir, relativePath)
-
-	glog.Infof("Deleting PV %q contents at hostpath %q, mountpath %q", pv.Name, specPath, mountPath)
-	return d.VolUtil.DeleteContents(mountPath)
+	// Remove API object
```
msau42: Maybe file delete should also be asynchronous, so you have the same handling for both.

local-volume/provisioner/pkg/deleter/blockcleaner.go
```diff
@@ -0,0 +1,247 @@
...
+}
+
+// NewBlockCleaner returns a BlockCleaner
+func NewBlockCleaner(config *common.RuntimeConfig) *BlockCleanerImpl {
+	return &BlockCleanerImpl{RuntimeConfig: config,
+		procTable: make(map[string]time.Time),
+	}
+}
+
+func (v *BlockCleanerImpl) isRunning(pvName string) bool {
```
msau42: What happens if the process restarts and we lose the in-memory state?  Is it safe to run the command twice?

local-volume/provisioner/pkg/deleter/blockcleaner.go
```diff
@@ -0,0 +1,247 @@
...
+	iters, expErr, exe, args, err := v.getBlockDeleteCommand(mountPath, config)
+	if err != nil {
+		return err
+	}
+
+	if v.isRunning(pv.Name) {
+		// Run in progress, nothing to do,
+		return nil
+	}
+	v.markRunning(pv.Name)
```
msau42: It may be good to also log an event on the PV when we start the cleanup and warn that it could take a long time.  Otherwise users could get impatient and just delete the PV object.  Actually, what happens if the user deletes the PV object and we're still cleaning the device?  Will the discovery kick in and recreate the PV?  Maybe we need some sort of check there.  I also still wonder if we need to persist the status somewhere and/or prevent the PV from being destroyed by the user unless some force flag is set.

local-volume/provisioner/pkg/common/common.go
```diff
@@ -49,6 +50,15 @@ const (
...
 	EventVolumeFailedDelete = "VolumeFailedDelete"
+
+	// BlockCleanupFastRest does a quick reset of any previous file system information.
+	BlockCleanupFastRest = "quick-reset"
+	// BlockCleanupZero does a more exhaustive zeroing out of all content in block device
+	BlockCleanupZero = "zero"
+	// BlockCleanupRandom Overwrite with random values.
+	BlockCleanupRandom = "random"
+	// BlockCleanupCustom Use custome script to do cleanup.
+	BlockCleanupCustom = "custom"
```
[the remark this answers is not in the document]
dhirajh: OK, will do.

local-volume/provisioner/pkg/common/common.go
```diff
@@ -122,6 +144,18 @@ func CreateLocalPVSpec(config *LocalPVConfig) *v1.PersistentVolume {
 	}
 }
 
+// GetPathFromPV gets the local path (within provisioner container) of the PV
+func GetPathFromPV(pv *v1.PersistentVolume, config MountConfig) (string, error) {
```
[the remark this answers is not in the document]
dhirajh: OK.

local-volume/provisioner/pkg/deleter/blockcleaner.go
```diff
@@ -0,0 +1,247 @@
...
+	// DDFullErrExpected Error ouptput expected when running dd to fill a device
+	DDFullErrExpected = "No space left on device"
+	// NoErrExpected No error expected from running command
+	NoErrExpected = ""
+)
+
+// BlockCleaner Interface for invoking async script to do cleanup of block device.
+type BlockCleaner interface {
+	// DoCleanup deletes
+	DeleteBlockPV(pv *v1.PersistentVolume, mountPath string, config common.MountConfig) error
```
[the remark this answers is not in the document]
dhirajh: OK

local-volume/provisioner/pkg/deleter/blockcleaner.go
dhirajh: Yes. Repeating it would only result in idempotent behavior (except in random, where the output would be a different set of random values, something that is desirable). It would repeat the exact same process and finish the job.

local-volume/provisioner/pkg/deleter/blockcleaner.go
```diff
@@ -0,0 +1,247 @@
...
+			deletingLocalPVErr.Error())
+		return err
+	}
+
+	return nil
+}
+
+func (v *BlockCleanerImpl) execAllIterations(pvName string, iterations int, expectErrOut string,
+	exe string, exeArgs ...string) error {
+
```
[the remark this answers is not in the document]
dhirajh: Cool, yes, that would be more elegant.

local-volume/provisioner/pkg/deleter/blockcleaner_test.go
dhirajh: @msau42 Can you clarify more on what duplicate operation means here? Do you mean specify some command twice in a single exec call and see if it actually ran twice? or do you mean run the same command in 2 iterations? There is a test for latter. I wasn't quite sure what the former scenario is attempting to test.

local-volume/provisioner/pkg/deleter/blockcleaner_test.go
dhirajh: There are couple of actual commands being run. One does a basic script that only deals in stdout. The other writes 12 bytes to a temp file (which is less than most logging), but I think the code can be modified to give feedback via stdout. Let me adjust that code.

local-volume/provisioner/pkg/deleter/deleter.go
dhirajh: OK

local-volume/provisioner/pkg/deleter/deleter.go
dhirajh: OK

local-volume/provisioner/pkg/common/common.go
```diff
@@ -122,6 +144,18 @@ func CreateLocalPVSpec(config *LocalPVConfig) *v1.PersistentVolume {
...
+	specPath := pv.Spec.Local.Path
+	relativePath, err := filepath.Rel(config.HostDir, specPath)
+	if err != nil {
+		return "", fmt.Errorf("Could not get relative path for pv %q: %v", pv.Name, err)
+	}
+	mountPath := filepath.Join(config.MountDir, relativePath)
```
ianchakeres: I think we could do without specPath and mountPath varibles, and simply place the variable/function onto the single line referencing each of those variables.

local-volume/provisioner/pkg/deleter/blockcleaner.go
```diff
@@ -0,0 +1,247 @@
...
+	// DefaultBlockSize is the default block size to use in dd
+	DefaultBlockSize = "4096"
+	// ShredCommand the shred cli.
+	ShredCommand = "shred"
+	// MkfsCommand is the mkfs executable.
+	MkfsCommand = "mkfs"
+	// DDCommand is the dd executable.
+	DDCommand = "dd"
+	// WipeComand is the dd executable.
+	WipeComand = "wipefs"
```
ianchakeres: WipefsCommand

local-volume/provisioner/pkg/deleter/blockcleaner.go
```diff
@@ -0,0 +1,247 @@
...
+	markDone(pvName string)
+}
+
+var _ BlockCleaner = &BlockCleanerImpl{}
+
+// BlockCleanerImpl Implementation of BLockCleaner interface
+type BlockCleanerImpl struct {
+	*common.RuntimeConfig
+	mutex     sync.Mutex
+	procTable map[string]time.Time
```
ianchakeres: I like that we're putting the start time into the procTable. How about procStartTimeTable?

Alternatively, if we don't display or use the time value perhaps procTable should simply be a simple boolean.

local-volume/provisioner/pkg/deleter/blockcleaner.go
```diff
@@ -0,0 +1,247 @@
...
+	v.mutex.Lock()
+	defer v.mutex.Unlock()
+	_, ok := v.procTable[pvName]
+	return ok
+}
+
+func (v *BlockCleanerImpl) markRunning(pvName string) {
+	v.mutex.Lock()
+	defer v.mutex.Unlock()
+	v.procTable[pvName] = time.Now()
```
ianchakeres: Should we show an error if v.procTable[pvName] is presently running (aka has a time value) instead of just overwriting the value?

local-volume/provisioner/pkg/deleter/blockcleaner.go
```diff
@@ -0,0 +1,247 @@
...
+}
+
+func (v *BlockCleanerImpl) markDone(pvName string) {
+	v.mutex.Lock()
+	defer v.mutex.Unlock()
+	delete(v.procTable, pvName)
+}
+
+// DeleteBlockPV is meant to execute cleanup concurrently with other such cleanup processes.
+func (v *BlockCleanerImpl) DeleteBlockPV(pv *v1.PersistentVolume, mountPath string, config common.MountConfig) error {
```
ianchakeres: Let's add some logging so that someone who's trying to figure out what's happening can see some output.
I'd recommend having at least one higher visibility log line at the start and another at the end of cleanup. These lines will allow an operator to see how long the cleaning operation took.

local-volume/provisioner/pkg/deleter/blockcleaner.go
```diff
@@ -0,0 +1,247 @@
...
+		// Run in progress, nothing to do,
+		return nil
+	}
+	v.markRunning(pv.Name)
+	go v.asyncDeleteBlockPV(pv, iters, expErr, exe, args...)
+
+	return nil
+}
+
+func (v *BlockCleanerImpl) getBlockDeleteCommand(mountPath string, config common.MountConfig) (int, string, string,
```
ianchakeres: Can you please add a comment about each of the return values?

local-volume/provisioner/pkg/deleter/blockcleaner.go
dhirajh: OK, will log. I will also put in a check before creating a PV to see if its currently being cleaned. If it is, then a log message will be generated and the creation blocked. I feel this largely covers reasonable use cases for now.

Having something in the API server/API itself for Local PVs to prevent deletion might be the longterm solution here. This is much like pods are not terminated until they get kubelet to shutdown the docker/rkt containers. Perhaps the PV needs a CLEANING and CLEANED phase (not just RELEASED) and refuse any un-forced deletion in CLEANING phase.

local-volume/provisioner/pkg/deleter/blockcleaner.go
```diff
@@ -0,0 +1,247 @@
...
+		// Atleast one iteration must be run.
+		iters = 1
+	}
+
+	switch config.BlockConf.Cleaner {
+
+	case common.BlockCleanupFastRest, "":
+		// mkfs -F /dev/sda && wipefs -a /dev/sda
+		mkfsCmd := fmt.Sprintf("%s -F %q && %s -a %q", MkfsCommand, mountPath, WipeComand, mountPath)
+		return 1, NoErrExpected, ShellCommand, []string{"-c", mkfsCmd}, nil
```
ianchakeres: Might be worth a comment that executing this command multiple times will result in the same output, therefore iters = 1.

local-volume/provisioner/pkg/deleter/blockcleaner.go
```diff
@@ -0,0 +1,247 @@
...
+	v.markRunning(pv.Name)
+	go v.asyncDeleteBlockPV(pv, iters, expErr, exe, args...)
+
+	return nil
+}
+
+func (v *BlockCleanerImpl) getBlockDeleteCommand(mountPath string, config common.MountConfig) (int, string, string,
+	[]string, error) {
+	iters := config.BlockConf.Iterations
+	if config.BlockConf.Iterations <= 0 {
```
ianchakeres: Should error out of this method if someone sets config.BlockConf.Iterations to zero or a negative value?

local-volume/provisioner/pkg/deleter/blockcleaner.go
```diff
@@ -0,0 +1,247 @@
...
+		// Atleast one iteration must be run.
+		iters = 1
+	}
+
+	switch config.BlockConf.Cleaner {
+
```
ianchakeres: The commands in this section are very dangerous. Should we add any special checks on mountPath here or when mountPath is constructed? For example, excluding some paths explicitly or checking that the path is within the path controlled by this provisioner.

local-volume/provisioner/pkg/deleter/blockcleaner.go
```diff
@@ -0,0 +1,247 @@
...
+	errReader, err := cmd.StderrPipe()
+	if err != nil {
+		return err
+	}
+
+	go func() {
+		defer wg.Done()
+		errScanner := bufio.NewScanner(errReader)
+		for errScanner.Scan() {
+			glog.Infof("Cleaner %q stderr: %q", pvName, errScanner.Text())
```
ianchakeres: Cleaner pvname: %q

local-volume/provisioner/pkg/deleter/blockcleaner.go
```diff
@@ -0,0 +1,247 @@
...
+
+	return nil
+}
+
+func (v *BlockCleanerImpl) execAllIterations(pvName string, iterations int, expectErrOut string,
+	exe string, exeArgs ...string) error {
+
+	for i := 0; i < iterations; i++ {
+		if i > 0 {
+			glog.Infof("PV %s : Cleaner %s %s", pvName, exe, exeArgs)
```
ianchakeres: Cleaner pvname: %s cmd: %s

More consistent with other log lines on L210 L223

local-volume/provisioner/pkg/deleter/blockcleaner_test.go
msau42: Since we perform the operation asynchronously, the main control loop will invoke the cleanup on this PV every few sounds, especially if the cleanup takes a long time.  So a test to make sure that we drop the subsequent requests while the 1st is still ongoing would be good.

local-volume/provisioner/pkg/deleter/blockcleaner.go
msau42: I like having start and end timestamps.  We can make metrics out of it.

local-volume/provisioner/pkg/deleter/blockcleaner.go
msau42: Yeah, let's make sure the paths are controlled by this provisioner.

local-volume/provisioner/pkg/common/common.go
```diff
@@ -49,6 +50,15 @@ const (
 
 	// EventVolumeFailedDelete copied from k8s.io/kubernetes/pkg/controller/volume/events
 	EventVolumeFailedDelete = "VolumeFailedDelete"
+
+	// BlockCleanupFastRest does a quick reset of any previous file system information.
+	BlockCleanupFastRest = "quick-reset"
```
msau42: I think it would be good to have an option for "blkdiscard" as a quick zeroing (but not secure) method for SSDs.  In general, for SSDs, writing zeros or random bytes at the block layer is not effective.  It's much better and faster to use the device commands available.  Some SSDs may support "secure erase" using hdparm as well.

local-volume/provisioner/pkg/deleter/blockcleaner_test.go
ianchakeres: If we do execute actual destructive commands here in the test, I'd prefer the happen in /tmp and with files/dirs that we create with an associated UUID.

I think that testHostDir is presently /mnt/disks. https://github.com/kubernetes-incubator/external-storage/blob/master/local-volume/provisioner/pkg/deleter/deleter_test.go#L32

local-volume/provisioner/pkg/deleter/blockcleaner.go
```diff
@@ -0,0 +1,247 @@
...
+	return ok
+}
+
+func (v *BlockCleanerImpl) markRunning(pvName string) {
+	v.mutex.Lock()
+	defer v.mutex.Unlock()
+	v.procTable[pvName] = time.Now()
+}
+
+func (v *BlockCleanerImpl) markDone(pvName string) {
```
ianchakeres: Do we need to have a timeout for these operations? I don't think we can assume that things always complete successfully and keeping their state around forever. Its probably better to error out and drop tasks that aren't responsive over some time period.

dhirajh: @ianchakeres I do want to track a process which is still deleting or writing to block as it is likely still active in some manner that makes the disk unusable until it terminates. The best way to actually drop the task is to kill it. Killing a process which is locked in I/O can quite often be a problem as it can ignore the signal when blocked in kernel I/O.  Also, with custom scripts, killing might have unintended consequences ( grand child processes stuck?). 

I think its highly preferable to let the issue be highlighted by a wedged PV and log messages. The admin fix would typically be to terminate the POD in which the provisioner is running, as all related processes would terminate. If we find that this happens often enough, then perhaps we could introduce a timeout which causes this self termination. I think its best to see some of these scenarios play out, and based on that design a timeout and termination approach.

local-volume/provisioner/pkg/common/common.go
dhirajh: I assuming that you would like to rename quick-reset to blkdiscard, will do so, Regarding zero-ing, yes there is an option to do that (and also random), but its not a quick process. It will run for a while for larger disks and as you mentioned, less effective for ssd. hdparm is going to be supported, I plan to provide a custom script which would script that.

local-volume/provisioner/pkg/deleter/blockcleaner.go
dhirajh: ok

local-volume/provisioner/pkg/deleter/blockcleaner.go
dhirajh: ok

local-volume/provisioner/pkg/deleter/blockcleaner.go
msau42: Even doing a kill -9 on a process stuck in the kernel won't help, and could potentially have bad consequences of causing any operation on that disk to hang.  The only way out I've found, is to reboot the node.  Maybe (as a separate feature), we should also have health probe on the provisioner if it ever gets into this situation, so we can notify the admin that some intervention may be needed.

local-volume/provisioner/pkg/common/common.go
msau42: No I mean an additional cleanup method for blkdiscard.  We can still keep quick-reset for destroying the filesystem.

local-volume/provisioner/pkg/deleter/blockcleaner.go
dhirajh: ok

local-volume/provisioner/pkg/common/common.go
dhirajh: So that option is currently named "zero". I can rename that to "blkdiscard".

Here are the ones currently supported:
	// BlockCleanupFastRest does a quick reset of any previous file system information.
	BlockCleanupFastRest = "quick-reset"
	// BlockCleanupZero does zeroing out of all content in block device
	BlockCleanupZero = "zero"  ==> "blkdiscard"
	// BlockCleanupRandom Overwrite with random values.
	BlockCleanupRandom = "random"
	// BlockCleanupCustom Use custom script to do cleanup.
	BlockCleanupCustom = "custom"

msau42: I'm only aware of using blkdiscard for SSDs.  Do HDDs support it?  I'm wondering if we need two separate zero routines.

dhirajh: Ah OK. I apologize, I thought you meant the choice of names for the different operations. You were referring to the utility blkdiscard. Yes, that could be added as an option here. The other possibility is to make it one of the currently available custom scripts. I imagine we might get other requests like "scrub", "wipe", etc and those could follow a similar pattern. Let me know what you think.

msau42: Or should we have no built-in options and just have everything be a plugin via script?

dhirajh: I like that idea.

msau42:
Also, can you write up a quick proposal for the discovery and cleanup of block volumes?  We should have the storage sig review it.

dhirajh:
@msau42 OK, will write that up.

local-volume/provisioner/pkg/deleter/deleter.go
dhirajh: @msau42 I looked into this more. It looks like core/v1 only defines constants for EventTypeNormal and EventTypeWarning, which is why the previous author of provisioner used warning in this situation and I copied it here. Let me know if there is a different constant that I could use here, I could not find another looking through k8s source code. The use cases seemed to be to use warning to report errors.

local-volume/provisioner/pkg/deleter/blockcleaner.go
dhirajh: OK, I made the log messages more consistent, but a bit different from above.

local-volume/provisioner/pkg/deleter/blockcleaner.go
dhirajh: same... made it more consistent.

local-volume/provisioner/pkg/common/common.go
dhirajh: fixed

local-volume/provisioner/pkg/deleter/blockcleaner.go
dhirajh: The main reason I have the table is to track the running processes. The hope is to eventually (not in this check-in) to use some of the stats to do some notifications on if things are running for too long. Since the primary purpose of the table is indicate currently running processes, procTable seems appropriate.

local-volume/provisioner/pkg/deleter/blockcleaner.go
dhirajh: This code is now gone.

local-volume/provisioner/pkg/deleter/blockcleaner.go
dhirajh: Done.

local-volume/provisioner/pkg/deleter/blockcleaner.go
dhirajh: Done.

local-volume/provisioner/pkg/common/common.go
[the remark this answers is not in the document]
dhirajh: This code is now gone.

local-volume/provisioner/pkg/deleter/blockcleaner.go
dhirajh: code is now gone.

local-volume/provisioner/pkg/deleter/blockcleaner.go
dhirajh: code removed.

local-volume/provisioner/pkg/deleter/blockcleaner.go
dhirajh: code removed.

local-volume/provisioner/pkg/deleter/deleter.go
msau42: Oh ok that's fine then

local-volume/provisioner/pkg/deleter/blockcleaner_test.go
dhirajh: With the up coming change, none of the unit test commands will do any write changes. The most they do is invoke echo to stdout or exit with error code.

local-volume/provisioner/pkg/deleter/deleter.go
dhirajh: done

local-volume/provisioner/pkg/common/common.go
dhirajh: done.

local-volume/provisioner/pkg/deleter/deleter.go
```diff
@@ -46,60 +52,191 @@ func (d *Deleter) DeletePVs() {
...
+	err := d.ProcTable.MarkRunning(pv.Name)
+	if err != nil {
+		return err
+	}
+	go d.asyncCleanupBlockPV(pv, mountPath, config.BlockCleanerCommand[0], config.BlockCleanerCommand[1:]...)
+
+	return nil
+}
+
+func (d *Deleter) asyncCleanupBlockPV(pv *v1.PersistentVolume, blkdevPath string, exe string, exeArgs ...string) error {
```
msau42: Can we have a common async cleanup method for both file and block PV?  It can take in a func() as an arg, that will call the file-specific methods or the block-specific methods.  Then we can have a common function that marks marks proc as running, mark as done, and delete PV object.

local-volume/provisioner/pkg/discovery/discovery_test.go
```diff
@@ -224,7 +225,7 @@ func TestDiscoverVolumes_BadVolume(t *testing.T) {
```
msau42: Can you add a test case here to make sure we don't discover a volume that is being cleaned?

local-volume/provisioner/deployment/docker/scripts/blkdiscard.sh
```diff
@@ -0,0 +1,28 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+# Usage:
+# $ quick_reset.sh
```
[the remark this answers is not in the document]
dhirajh: @msau42 validateBlock() function actually validates that the content of that env variable is a valid block device. I believe that does what you are suggesting and a bit more. Let me know if that works.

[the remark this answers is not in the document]
msau42: Does it work if the env variable is empty or not set?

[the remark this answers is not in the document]
dhirajh: Yes, it does. But the error message is not the best, it says that - "" is not a block device. I will do a check to see if it is set, so the error message is better.

local-volume/provisioner/pkg/deleter/deleter.go
```diff
@@ -126,69 +126,53 @@ func (d *Deleter) asyncDeleteFilePV(pv *v1.PersistentVolume, mountPath string, c
...
-// cleanupBlockPV is meant to execute cleanup concurrently with other such cleanup processes
-// cmdio is a param that can suppoly io to the script being invoked. Its main purpose currently is
-// to enable better testing of the code path with output validation etc.
-func (d *Deleter) cleanupBlockPV(pv *v1.PersistentVolume, mountPath string, config common.MountConfig) error {
+func (d *Deleter) cleanupBlockPV(pv *v1.PersistentVolume, blkdevPath string, config common.MountConfig) error {
+
 	// Make absolutely sure here that we are not deleting anything outside of mounted dir
-	if !strings.HasPrefix(mountPath, config.MountDir) {
-		err := fmt.Errorf("Unexpected error pv %q mountPath %s but mount dir is %s", pv.Name, mountPath,
+	if !strings.HasPrefix(blkdevPath, config.MountDir) {
```
msau42: Can we have this check for filesystem type too?

local-volume/provisioner/pkg/deleter/deleter.go
```diff
@@ -126,69 +126,53 @@ func (d *Deleter) asyncDeleteFilePV(pv *v1.PersistentVolume, mountPath string, c
...
-}
-
-func (d *Deleter) asyncCleanupBlockPV(pv *v1.PersistentVolume, blkdevPath string, exe string, exeArgs ...string) error {
-
-	// Make sure we mark as done when we exit this function
-	defer d.ProcTable.MarkDone(pv.Name)
 	glog.Infof("Deleting PV block volume %q device hostpath %q, mountpath %q", pv.Name, pv.Spec.Local.Path,
 		blkdevPath)
 
 	glog.Infof("Starting cleanup of pv %q", pv.Name)
```
msau42: This log msg can be removed since there's a similar one above

local-volume/provisioner/pkg/discovery/discovery_test.go
dhirajh: This has already been implemented in TestDiscoverVolumes_CleaningInProgress

local-volume/provisioner/pkg/deleter/deleter.go
dhirajh: done

local-volume/provisioner/pkg/deleter/deleter.go
dhirajh: done

local-volume/provisioner/pkg/common/common.go
```diff
@@ -182,6 +203,18 @@ func ConfigMapDataToVolumeConfig(data map[string]string, provisionerConfig *Prov
...
+		if config.BlockCleanerCommand == nil {
+			// supply a default block cleaner command.
+			config.BlockCleanerCommand = []string{DefaultBlockCleanerCommand}
+		} else {
+			// Validate that array is non empty
+			if len(config.BlockCleanerCommand) < 1 {
+				return fmt.Errorf("Invalid empty block cleaner command for class %v", class)
+			}
+		}
+		provisionerConfig.StorageClassConfig[class] = config
```
msau42: config is a local variable?

local-volume/provisioner/pkg/deleter/proctable.go
```diff
@@ -0,0 +1,125 @@
...
+	IsRunning(pvName string) bool
+	IsEmpty() bool
+	MarkRunning(pvName string) error
+	MarkDone(pvName string)
+}
+
+var _ ProcTable = &ProcTableImpl{}
+
+// ProcEntry represents an entry in the proc table
+type ProcEntry struct {
```
msau42: This can be done as a future task, but can we add metrics support so that we can get stats on how long deletion takes?

local-volume/provisioner/pkg/common/common.go
dhirajh: Yes, but the value is being saved back into the map. Go doesn't allow changing a struct member when the struct is in a map, one has the set the entire struct.

local-volume/provisioner/pkg/deleter/proctable.go
dhirajh: OK, will do. Assuming this will be reported out via gRPC in minions, I will introduce it once the gRPC layer is in place.

local-volume/provisioner/pkg/deleter/deleter.go
```diff
@@ -46,60 +52,178 @@ func (d *Deleter) DeletePVs() {
...
 	}
 
-	mountPath := filepath.Join(config.MountDir, relativePath)
+	glog.Infof("Deleted PV %q", pv.Name)
+
+}
+
+func (d *Deleter) deleteFilePV(pv *v1.PersistentVolume, mountPath string, config common.MountConfig) error {
+	// Make absolutely sure here that we are not deleting anything outside of mounted dir
+	if !strings.HasPrefix(mountPath, config.MountDir) {
```
msau42: Can this check be done outside since it's common to both file and block?

local-volume/provisioner/pkg/discovery/discovery.go
```diff
@@ -33,13 +34,17 @@ import (
...
+	// it is being cleaned
+	ProcTable       deleter.ProcTable
 	nodeAffinityAnn string
+	blockDisabled   bool
 }
 
 // NewDiscoverer creates a Discoverer object that will scan through
 // the configured directories and create local PVs for any new directories found
-func NewDiscoverer(config *common.RuntimeConfig) (*Discoverer, error) {
+func NewDiscoverer(config *common.RuntimeConfig, procTable deleter.ProcTable, blockDisabled bool) (*Discoverer, error) {
```
msau42: Can you add it to the common.RuntimeConfig instead? It'll be easier to remove in the future.

local-volume/provisioner/pkg/deleter/deleter.go
dhirajh: Fixed.

local-volume/provisioner/pkg/discovery/discovery.go
dhirajh: @msau42 Fixed.

PR #314, opened 2017-08-24
[glusterfs] Disable CGO to build static binary

yuanying:
`glusterfs-simple-provisioner` binary was dynamically linked
unless disable CGO_ENABLED flag. So that container image doesn't
work because alpine image doesn't contain dependent libraries.

This patch disable CGO_ENABLED flag to build container image
correctly.


yuanying:
Please check following command.
Docker responds with panic message.

```bash
$ docker run --rm -ti quay.io/external_storage/glusterfs-simple-provisioner:latest
```

wongma7:
I'll push the fixed image later today

PR #331, opened 2017-09-01
Add volume snapshots feature

tsmetana:
This patch set adds code for the volume snapshots support. The code comes from github.com/rootfs/snapshot master branch. In order to make the new code compile there are updated vendor dependencies. I have also modified the Makefile by adding new target for snapshots and also traditional "all" target to compile everything in the tree.

cc: @rootfs


tsmetana:
Sorry for the failed tests. Will fix that.

wongma7:
@tsmetana given how excessive/annoying golint can be, I am going to look into copying upstreams `hack/.golint_failures` so that we can exempt snapshots from it

tsmetana:
OK. Thank you. I will fix the easy ones (gofmt) and the "errors" ignoring the "warnings" for the moment.

wongma7:
i've opened https://github.com/kubernetes/repo-infra/pull/43, if nobody responds today I'll just patch our repo-infra myself

wongma7:
okay, patched
rebase onto master, then add a line `snapshots` or lines `snapshots/foo` `snapshots/bar` to a new file .golintignore at the root of external-storage, and golint will stop complaining.

tsmetana:
OK... I definitely must not have done that correctly.

wongma7:
no, it's working, golint isn't running on snapshots anymore. But there's `deadcode`, `vetshadow` and `gocyclo`, which I don't think upstream runs hence the errors

wongma7:
I think upstream only runs govet, of the four gometalinters https://github.com/kubernetes-incubator/external-storage/blob/master/repo-infra/verify/go-tools/verify-gometalinter.sh

have to wonder why repo-infra imposes stricter requirements than upstream but oh well. I guess there needs to be a way to skip  not just golint but the other ones too, I will work on it.

wongma7:
k, https://github.com/kubernetes-incubator/external-storage/pull/342 is merged. next tests should pass...

snapshot/pkg/controller/snapshotter/snapshotter.go
```diff
@@ -268,12 +268,9 @@ func (vs *volumeSnapshotter) updateSnapshotDataStatus(snapshotName string, snaps
 
 	if len(snapshotDataObj.Status.Conditions) < 1 ||
 		snapshotDataObj.Status.Conditions[0].Type != crdv1.VolumeSnapshotDataConditionReady {
-		pv, err := vs.getPVFromVolumeSnapshot(snapshotName, snapshot)
```
rootfs: remove `getPVFromVolumeSnapshot` func as well

snapshot/pkg/apis/crd/v1/types.go
```diff
@@ -249,6 +249,22 @@ func GetSupportedVolumeFromPVSpec(spec *core_v1.PersistentVolumeSpec) string {
 	return ""
 }
 
+func GetSupportedVolumeFromSnapshotDataSpec(spec *VolumeSnapshotDataSpec) string {
```
rootfs: remove `GetSupportedVolumeFromPVSpec` as well

snapshot/cmd/snapshot-pv-provisioner/snapshot-pv-provisioner.go
```diff
@@ -0,0 +1,302 @@
...
+	var config *rest.Config
+	var err error
+	if *master != "" || *kubeconfig != "" {
+		config, err = clientcmd.BuildConfigFromFlags(*master, *kubeconfig)
+	} else {
+		config, err = rest.InClusterConfig()
+	}
+	prId := string(uuid.NewUUID())
+	if *id != "" {
+		prId = *id
```
wongma7: We should default to provisionerName instead of generating every start up like https://github.com/kubernetes-incubator/external-storage/blob/master/ceph/rbd/cmd/rbd-provisioner/main.go#L67

this means Delete will succeed from any instance of the provisioner with same provisionername, I assume that is safe.

snapshot/pkg/controller/snapshotter/snapshotter.go
```diff
@@ -0,0 +1,943 @@
...
+			glog.Infof("waitForSnapshot: Snapshot %s created successfully. Adding it to Actual State of World.", snapshotName)
+			vs.actualStateOfWorld.AddSnapshot(newSnapshot)
+			// Break out of the for loop
+			return nil
+		} else if newstatus == statusError {
+			return fmt.Errorf("Failed to create snapshot %s.", snapshotName)
+		}
+	}
+
+	return fmt.Errorf("Snapshot %s is NOT completed successfully.", snapshotName)
```
wongma7: CI govet says unreachable code, we will never break out of the above loop

tsmetana: It's intentional. The offending method is being called from a goroutine that actually is meant to either succeed or fail... And wait forever when none of that happens.

wongma7: okay, can safely remove this line then

wongma7:
need boilerplate :( can run `make verify` locally to see what the ci is complaining about.

Will merge asap afterwards :)

wongma7:
merging, pls consider https://github.com/kubernetes-incubator/external-storage/pull/331#discussion_r137557534 at a later date, a new identity every time means if the snapshot provisioner restarts it will avoid deleting old snapshots which is maybe overly cautious

tsmetana:
Thanks. I'll take a look at the provisioner name.

PR #335, opened 2017-09-01
fixed role names change in openshift 3.6

raffaelespazzoli:


wongma7:
This breaks compatibility with Openshift 3.5, which doesn't have the persistent-volume-provisioner role.

Moreover, I think that in 3.6, persistent-volume-provisioner alone will suffice, we shouldn't need the recycler and binder roles on top of that.

What we could do that works for both 3.5 and 3.6 and beyond is create our own custom role like  https://raw.githubusercontent.com/kubernetes-incubator/external-storage/master/aws/efs/deploy/auth/openshift-clusterrole.yaml and tell users to just create that

here was my attempt to fix it but i haven't had time https://github.com/kubernetes-incubator/external-storage/pull/290/files

raffaelespazzoli:
ok, I'm good either way. since you were already working on this can you take care of it? Should I just remove this PR? 
Please note that I also found an error in a kubernetes config file. If you complete this task you could fix it.

wongma7:
okay, I will take care of it all in my PR. 👍

PR #346, opened 2017-09-08
Use exponential backoff

xing-yang:
This PR uses exponential backoff instead of a fixed internal
when waiting for snapshot to be created and when retrying
the creation of VolumeSnapshotData.

### DIFF
--- a/snapshot/pkg/controller/snapshotter/snapshotter.go
+++ b/snapshot/pkg/controller/snapshotter/snapshotter.go
@@ -26,6 +26,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/uuid"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/kubernetes/pkg/util/goroutinemap"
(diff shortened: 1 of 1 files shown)


xing-yang:
@jingxu97 @tsmetana @rootfs

snapshot/pkg/controller/snapshotter/snapshotter.go
```diff
@@ -620,22 +634,26 @@ func (vs *volumeSnapshotter) createVolumeSnapshotData(snapshotName string, snaps
...
-			// https://github.com/jpillora/backoff
 			// Re-Try it as errors writing to the API server are common
-			time.Sleep(createVolumeSnapshotDataInterval)
+			return false, err
 		} else {
-			break
+			return true, nil
 		}
-	}
+		return false, nil
```
rootfs: there is a govet error
```
snapshot/pkg/controller/snapshotter/snapshotter.go:655: unreachable code
exit status 1
FAILED   /home/travis/gopath/src/github.com/kubernetes-incubator/external-storage/repo-infra/verify/go-tools/verify-govet.sh	3s
```

xing-yang: Fixed the govet error.  Thanks.

tsmetana:
Looks OK now. I'll try to test it.

xing-yang:
@jingxu97 Regarding "If updateVolumeSnapshot failed once, should we break the loop?", you mean we should retry if it failed for the first time, but return failure if it failed for the 2nd time?

PR #354, opened 2017-09-08
Fix golint errors

xing-yang:
This PR fixes the golint errors in all the files
under external-storage/snapshot. A few changes
are highlighted here:
- Renames VolumePlugin to Plugin because it is referred
  by other packages as volume.Plugin.
- Renames aws_ebs to awsebs and gce_pd to gcepd because
  golint says do not use an underscore in package name.
- Renames GCECloud to Cloud because it is referred by
  other packages as gce.Cloud.

### DIFF
--- a/snapshot/cmd/snapshot-controller/snapshot-controller.go
+++ b/snapshot/cmd/snapshot-controller/snapshot-controller.go
@@ -36,9 +36,9 @@ import (
 	"github.com/kubernetes-incubator/external-storage/snapshot/pkg/cloudprovider/providers/openstack"
 	snapshotcontroller "github.com/kubernetes-incubator/external-storage/snapshot/pkg/controller/snapshot-controller"
 	"github.com/kubernetes-incubator/external-storage/snapshot/pkg/volume"
-	"github.com/kubernetes-incubator/external-storage/snapshot/pkg/volume/aws_ebs"
+	"github.com/kubernetes-incubator/external-storage/snapshot/pkg/volume/awsebs"
 	"github.com/kubernetes-incubator/external-storage/snapshot/pkg/volume/cinder"
-	"github.com/kubernetes-incubator/external-storage/snapshot/pkg/volume/gce_pd"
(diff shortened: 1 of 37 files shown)


xing-yang:
@rootfs @tsmetana @jingxu97

This PR fixes all the golint errors for snapshot.

snapshot/cmd/snapshot-controller/snapshot-controller.go
```diff
@@ -36,9 +36,9 @@ import (
 	"github.com/kubernetes-incubator/external-storage/snapshot/pkg/cloudprovider/providers/openstack"
 	snapshotcontroller "github.com/kubernetes-incubator/external-storage/snapshot/pkg/controller/snapshot-controller"
 	"github.com/kubernetes-incubator/external-storage/snapshot/pkg/volume"
-	"github.com/kubernetes-incubator/external-storage/snapshot/pkg/volume/aws_ebs"
+	"github.com/kubernetes-incubator/external-storage/snapshot/pkg/volume/awsebs"
```
rootfs: this is annoying

rootfs:
gofmt error
```
Verifying /home/travis/gopath/src/github.com/kubernetes-incubator/external-storage/repo-infra/verify/go-tools/verify-gofmt.sh
!!! 'gofmt -s' needs to be run on the following files: 
./snapshot/pkg/cloudprovider/providers/gce/gce.go
./snapshot/pkg/cloudprovider/providers/openstack/openstack.go
./snapshot/pkg/apis/crd/v1/types.go
./snapshot/pkg/apis/crd/v1/register.go
FAILED   /home/travis/gopath/src/github.com/kubernetes-incubator/external-storage/repo-infra/verify/go-tools/verify-gofmt.sh	0s
```

snapshot/cmd/snapshot-controller/snapshot-controller.go
xing-yang: Agreed, but what else can we do?  I noticed that persistentvolume and attachdetach are in the same situation.

xing-yang:
Fixed the gofmt errors.  Thanks.

Issue #356, opened 2017-09-11
[OpenEBS] Update the prefix of pvName to openebs-volume for OpenEBS provisioned volumes


satyamz:
We want to change the persistent volume name created by OpenEBS provisioner from `pvc-[unique-id]` to `openebs-volume-[unique-id]` 
For example: 
We want change NAME from  
```
$ kubectl get pv
NAME                                       CAPACITY   ACCESSMODES   STATUS    CLAIM                        REASON    AGE
pvc-bb6d2f0c-534c-11e6-9348-42010af00002   3Gi        RWO           Bound     default/claim1  
```
To 
  
```
$ kubectl get pv
NAME                                                  CAPACITY   ACCESSMODES   STATUS    CLAIM                        REASON    AGE
openebs-volume-bb6d2f0c-534c-11e6-9348-42010af00002   3Gi        RWO           Bound     default/claim1  
```

PR #357, opened 2017-09-11
Change name of volume provisioned by openebs

satyamz:
This will change the name of the volume provisioned by OpenEBS.
Earlier, it used to create volume with name pvc-[unique-id]
which was quite difficult to recognize.Hence, changing it to
openebs-volume-[unique-id].

Closes : #356

### DIFF
--- a/openebs/openebs-provisioner.go
+++ b/openebs/openebs-provisioner.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"os"
 
+	"strings"
 	"syscall"
 
 	"github.com/golang/glog"
@@ -85,6 +86,13 @@ func (p *openEBSProvisioner) Provision(options controller.VolumeOptions) (*v1.Pe
 	var openebsVol mApiv1.OpenEBSVolume
 
 	volSize := options.PVC.Spec.Resources.Requests[v1.ResourceName(v1.ResourceStorage)]
+
+	if strings.HasPrefix(options.PVName, "pvc") {
+		pvNameArr := strings.SplitAfter(options.PVName, "pvc")
+		if len(pvNameArr) > 1 {
+			options.PVName = "openebs-volume" + pvNameArr[1]
+		}
+	}
 
 	_, err := openebsVol.CreateVsm(options.PVName, volSize.String())
 	if err != nil {


satyamz:
Sample output after applying this change: 
```
ubuntu@kubemaster-01:~/openebs/k8s$ kubectl get pods 
NAME                                                              READY     STATUS    RESTARTS   AGE
jupyter-server-2764185079-zdqdx                                   1/1       Running   0          33m
maya-apiserver-1733699612-hcbgj                                   1/1       Running   0          35m
openebs-provisioner-1512667371-025df                              1/1       Running   0          35m
openebs-volume-d692ff6c-96db-11e7-ad39-021c6f7dbe9d-ctrl-3bh7tm   1/1       Running   0          33m
openebs-volume-d692ff6c-96db-11e7-ad39-021c6f7dbe9d-rep-246pvlc   1/1       Running   0          33m
openebs-volume-d692ff6c-96db-11e7-ad39-021c6f7dbe9d-rep-24kpzn5   1/1       Running   0          33m
ubuntu@kubemaster-01:~/openebs/k8s$ kubectl get pv 
NAME                                                  CAPACITY   ACCESSMODES   RECLAIMPOLICY   STATUS    CLAIM                            STORAGECLASS      REASON    AGE
openebs-volume-d692ff6c-96db-11e7-ad39-021c6f7dbe9d   5G         RWO           Delete          Bound     default/jupyter-data-vol-claim   openebs-jupyter             33m
ubuntu@kubemaster-01:~/openebs/k8s$ kubectl get pvc
NAME                     STATUS    VOLUME                                                CAPACITY   ACCESSMODES   STORAGECLASS      AGE
jupyter-data-vol-claim   Bound     openebs-volume-d692ff6c-96db-11e7-ad39-021c6f7dbe9d   5G         RWO           openebs-jupyter   33m

```

wongma7:
At the moment we assume the PV has a deterministic name https://github.com/kubernetes-incubator/external-storage/blob/master/lib/controller/controller.go#L751 so that we don't create a volume if it was somehow already created. So this PR will change behavior of the provisioner somewhat, are you okay with that?

perhaps we can add an optional function to the Provisioner interface where you define getProvisionedVolumeNameForClaim yourself.

humblec:
@satyamz  Ideally keeping pvc-{uniqueid} is good enough, do you have any specific requirement to be met by having a custom name ? One of the advantage is that, it gives an abstraction and at any point of user can define `their PVC name` as they want.

humblec:
./assign @humblec
.

satyamz:
Hi, @humblec & @wongma7 ,  Our intention to change the name is simple, we want to recognize the volumes provisioned by OpenEBS provisioner. Just for usability purpose. If this not the good practice to follow and will break things up then I will not proceed with this.  

@humblec : 
> One of the advantage is that, it gives an abstraction and at any point, user can define their PVC name as they want.

I agree with you on this. However, I think we can manage this behavior in the case with custom `pvc` name (if we follow the design suggested by @wongma7). 
 
@wongma7 The method suggested by you sounds good, but I need to think more about the design. Currently, I am going through `lib/controller`. 
I would love to have more pointers towards this apart from what you've already mentioned. I want to give it a try :slightly_smiling_face:

humblec:
`, Our intention to change the name is simple, we want to recognize the volumes provisioned by OpenEBS provisioner. Just for usability purpose`

@satyamz There are other ways, have a `annotation` as in other provisioners.

satyamz:
@humblec I agree that annotations will definitely work in this case. As I was having look at kubernetes dashboard and I was thinking about how to let users know that what kind of volumes are provisioned as soon as users click on `volumes`. 
I also liked the idea of having a generic method to the Provisioner interface which is better than only modifying single provisioner. If this is fine with you then shall I raise an issue and start working on it? and also let's close this PR and focus on the other.

wongma7:
Yes, @satyamz I think a new method would be nice. IIRC you are not the first to want to name the pvs something more useful. However, we should make the method optional by creating a new interface that extends the Provisioner one. I'm not sure what it would be called. https://stackoverflow.com/questions/44577539/go-optional-methods-in-an-interface/44577611#44577611 Let's discuss further if need be in the issue

satyamz:
@wongma7  That's great! I am going to close this PR. I will create the new issue for the addition of optional method. Thanks, @wongma7  & @humblec

PR #362, opened 2017-09-12
Simplify Controller Workflow

xing-yang:
This PR makes the following changes to the snapshot controller
workflow:

- After a snapshot creation request is sent to the plugin,
  create VolumeSnapshotData object first, and then bind
  VolumeSnapshotData to VolumeSnapshot and update VolumeSnapshot
  status.
- Simplify getSnapshotStatus function because now VolumeSnapshotData
  would have been created if status is Pending.

### DIFF
--- a/snapshot/pkg/controller/snapshotter/snapshotter.go
+++ b/snapshot/pkg/controller/snapshotter/snapshotter.go
@@ -371,74 +371,52 @@ func (vs *volumeSnapshotter) getPlugin(snapshotName string, snapshot *crdv1.Volu
 	return &plugin, nil
 }
 
-func (vs *volumeSnapshotter) getSnapshotStatus(snapshot *crdv1.VolumeSnapshot) (string, *crdv1.VolumeSnapshotData, error) {
-	var bCreateSnapData = false
+func (vs *volumeSnapshotter) updateSnapshotIfExists(snapshot *crdv1.VolumeSnapshot) (string, *crdv1.VolumeSnapshotData, error) {
 	var snapshotName string = snapshot.Metadata.Name
 	var snapshotDataObj *crdv1.VolumeSnapshotData
 	var snapshotDataSource *crdv1.VolumeSnapshotDataSource
 	var conditions *[]crdv1.VolumeSnapshotCondition
 	var err error
-	status := vs.getSimplifiedSnapshotStatus(&snapshot.Status.Conditions)
-	if status == statusReady {
-		return status, nil, nil
-	} else if status == statusPending {
-		// If we are here, takeSnapshot has already happened.
-		// Check whether the VolumeSnapshotData object is already created
-		snapshotDataObj = vs.getSnapshotDataFromSnapshotName(snapshotName)
-		if snapshotDataObj == nil {
-			bCreateSnapData = true
-			// Find snapshot by existing tags, and create VolumeSnapshotData
-		} else {
-			// Bind VolumeSnapshotData to VolumeSnapshot if it has not happened yet
-			if snapshot.Spec.SnapshotDataName != snapshotDataObj.Metadata.Name {
-				glog.Infof("getSnapshotStatus: bind VolumeSnapshotData to VolumeSnapshot %s.", snapshotName)
-				err = vs.bindVolumeSnapshotDataToVolumeSnapshot(snapshotName, snapshotDataObj.Metadata.Name)
-				if err != nil {
-					glog.Errorf("getSnapshotStatus: Error updating volume snapshot %s: %v", snapshotName, err)
-					return statusError, nil, err
-				}
-			}
-			return status, snapshotDataObj, nil
-		}
-	} else if status == statusError {
-		return status, nil, fmt.Errorf("Failed to find snapshot %s", snapshotName)
-	}
-	bCreateSnapData = true
 	// Find snapshot by existing tags, and create VolumeSnapshotData
-
-	if bCreateSnapData {
-		snapshotDataSource, conditions, err = vs.findSnapshot(snapshotName, snapshot)
-		if err != nil {
-			return statusNew, nil, nil
-		}
-		glog.Infof("getSnapshotStatus: create VolumeSnapshotData object for VolumeSnapshot %s.", snapshotName)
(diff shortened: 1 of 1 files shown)


snapshot/pkg/controller/snapshotter/snapshotter.go
```diff
@@ -382,56 +381,35 @@ func (vs *volumeSnapshotter) getSnapshotStatus(snapshot *crdv1.VolumeSnapshot) (
...
-			}
-		}
-		glog.Infof("getSnapshotStatus: bind VolumeSnapshotData to VolumeSnapshot %s.", snapshotName)
-		err = vs.bindVolumeSnapshotDataToVolumeSnapshot(snapshotName, snapshotDataObj.Metadata.Name)
-		if err != nil {
-			glog.Errorf("getSnapshotStatus: Error binding VolumeSnapshotData to VolumeSnapshot %s: %v", snapshotName, err)
-			return statusError, nil, err
-		}
-		return statusPending, snapshotDataObj, nil
+	// Find snapshot by existing tags, and create VolumeSnapshotData
```
tsmetana: Shouldn't we split the function here? It's named 'getSnapshotStatus', so it's rather surprising that is also creates VolumeSnapshotData and binds them to the VolumeSnapshot. It's also called just from syncSnapshot so perhaps returning here and calling another method that creates and binds VolumeSnapshotData from syncSnapshot when needed would improve the readability.

snapshot/pkg/controller/snapshotter/snapshotter.go
```diff
@@ -382,56 +381,35 @@ func (vs *volumeSnapshotter) getSnapshotStatus(snapshot *crdv1.VolumeSnapshot) (
...
+	snapshotDataSource, conditions, err = vs.findSnapshot(snapshotName, snapshot)
+	if err != nil {
+		return statusNew, nil, nil
+	}
+	glog.Infof("getSnapshotStatus: create VolumeSnapshotData object for VolumeSnapshot %s.", snapshotName)
```
rootfs: add a comment here explaining why no snapshot found and why we need to create snapshot now

snapshot/pkg/controller/snapshotter/snapshotter.go
xing-yang: The function name "getSnapshotStatus" is confusing.  Maybe rename it to "checkAndUpdateSnapshot"?

snapshot/pkg/controller/snapshotter/snapshotter.go
xing-yang: Sure.

snapshot/pkg/controller/snapshotter/snapshotter.go
tsmetana: I don't see that as an improvement: The function does not update the status, it creates and binds VolumeSnapshotData based on the status. Splitting it in two functions (one to get the status and other to create and bind the object) looks more "natural" to me.

xing-yang: Do you mean to add crdv1.VolumeSnapshotConditionPendingMissingData as well so CloudProvider will only return this new state, or do you mean only add statusPendingMissingData within snapshotter.go itself (so we'll have statusPending plus statusPendingMissingData)?

xing-yang: I split out the code to find snapshot, create VolumeSnapshotData, and update VolumeSnapshot status and bind VolumeSnapshotData to VolumeSnapshot to a new function called updateSnapshotIfExists.

snapshot/pkg/controller/snapshotter/snapshotter.go
```diff
@@ -372,73 +372,60 @@ func (vs *volumeSnapshotter) getPlugin(snapshotName string, snapshot *crdv1.Volu
 }
 
 func (vs *volumeSnapshotter) getSnapshotStatus(snapshot *crdv1.VolumeSnapshot) (string, *crdv1.VolumeSnapshotData, error) {
```
jingxu97: looks like we can combine getSnapshotStatus and getSimplifiedSnapshotStatus after this change. both functions are doing the similar things.

xing-yang: Yes.

snapshot/pkg/controller/snapshotter/snapshotter.go
```diff
@@ -371,74 +371,52 @@ func (vs *volumeSnapshotter) getPlugin(snapshotName string, snapshot *crdv1.Volu
...
 // Below are the closures meant to build the functions for the GoRoutineMap operations.
 // syncSnapshot is the main controller method to decide what to do to create a snapshot.
 func (vs *volumeSnapshotter) syncSnapshot(snapshotName string, snapshot *crdv1.VolumeSnapshot) func() error {
 	return func() error {
-		status, snapshotDataObj, err := vs.getSnapshotStatus(snapshot)
+		var err error
+		var snapshotDataObj *crdv1.VolumeSnapshotData
+		status := vs.getSimplifiedSnapshotStatus(&snapshot.Status.Conditions)
+		if status == statusNew {
+			status, snapshotDataObj, err = vs.updateSnapshotIfExists(snapshot)
```
jingxu97: Because the current workflow is
1. Create metadata and update in API object
2. Take a snapshot
3. Create VolumeSnapshotData and update object
4. Bind object by updating VolumeSnapshot (the status is also changed to pending at this time)

If failure happens between 3 and 4, it means that snapshot is already taken (don't even need to findSnapshot through cloud provider) since we already have a VolumeSnapshotData object created which refers to the volume snapshot. so   I think in updateSnapshotIfExists and findSnapshot function we can do the following
1. Check metadata exist or not, if no return statusNew
2. search volumeSnapshotData object which refers the the volume, if yes, update volumeSnapshot object with the snapshotData object name, returns statusPending. 
3 findSnapshot through cloud provider. If found, create snapshotData and bind it, and return statusPending. If not, return statusNew.

jingxu97: but we can merge this change now, and do another round of refactoring

PR #363, opened 2017-09-12
Use GB units instead of GiB units.

humblec:


humblec:
@wongma7 PLAL :). Thanks

PR #366, opened 2017-09-12
Add UID to Snapshot Tag

xing-yang:
Add UID from VolumeSnapshot Metadata as a tag to the snapshot created
by the plugin to make sure we can find the right snapshot through the
plugin corresponding to the VolumeSnapshot.

### DIFF
--- a/snapshot/pkg/cloudprovider/providers/openstack/openstack_snapshots.go
+++ b/snapshot/pkg/cloudprovider/providers/openstack/openstack_snapshots.go
@@ -226,6 +226,9 @@ func (os *OpenStack) FindSnapshot(tags map[string]string) ([]string, []string, e
 
 	glog.Infof("Looking for matching tags [%#v] in snapshots.", tags)
 	// Loop around to find the snapshot with the matching input metadata
+	// NOTE(xyang): Metadata based filtering for snapshots is supported by Cinder volume API
+	// microversion 3.21 and above. Currently the OpenStack Cloud Provider only supports V2.0.
+	// Revisit this later when V3.0 is supported.
 	for _, snapshot := range snapshots {
 		glog.Infof("Looking for matching tags in snapshot [%#v].", snapshot)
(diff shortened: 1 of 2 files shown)


snapshot/pkg/cloudprovider/providers/openstack/openstack_snapshots.go
```diff
@@ -234,12 +234,17 @@ func (os *OpenStack) FindSnapshot(tags map[string]string) ([]string, []string, e
...
 				if ok {
 					if tags[ctrlsnap.CloudSnapshotCreatedForVolumeSnapshotNameTag] == nameVal {
-						timeVal, ok := snapshot.Metadata[ctrlsnap.CloudSnapshotCreatedForVolumeSnapshotTimestampTag]
+						uidVal, ok := snapshot.Metadata[ctrlsnap.CloudSnapshotCreatedForVolumeSnapshotUIDTag]
 						if ok {
-							if tags[ctrlsnap.CloudSnapshotCreatedForVolumeSnapshotTimestampTag] == timeVal {
-								snapshotIDs = append(snapshotIDs, snapshot.ID)
-								statuses = append(statuses, snapshot.Status)
-								glog.Infof("Add snapshot [%#v].", snapshot)
+							if tags[ctrlsnap.CloudSnapshotCreatedForVolumeSnapshotUIDTag] == uidVal {
```
jingxu97: Does openstack support filter when listing snapshots? Use filter could avoid manually checking the meta data?

xing-yang: Metadata based filtering is supported by Cinder volume API microversion 3.21 and above. Currently the OpenStack Cloud Provider only supports V2.0.  I'll add a note to revisit this later.

xing-yang: A note is added.

PR #374, opened 2017-09-20
Snapshot: Hostpath should count with errors in creating the tarball

tsmetana:
During my tests I noticed that HostPath plugin tends to create empty snapshots on errors in PV (invalid path for hostpath). This is because tar first creates a file and if something fails it leaves it empty on the disk. The controller re-tries to take a snapshot on error. We should count with these: first check whethere there is really something to "snapshot" and then delete the broken snapshot if tar returned error.

### DIFF
--- a/snapshot/pkg/volume/hostpath/processor.go
+++ b/snapshot/pkg/volume/hostpath/processor.go
@@ -61,6 +61,9 @@ func (h *hostPathPlugin) SnapshotCreate(pv *v1.PersistentVolume, tags *map[strin
 		return nil, nil, fmt.Errorf("invalid PV spec %v", spec)
 	}
 	path := spec.HostPath.Path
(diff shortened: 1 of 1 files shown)


snapshot/pkg/volume/hostpath/processor.go
```diff
@@ -61,6 +61,9 @@ func (h *hostPathPlugin) SnapshotCreate(pv *v1.PersistentVolume, tags *map[strin
 		return nil, nil, fmt.Errorf("invalid PV spec %v", spec)
 	}
 	path := spec.HostPath.Path
+	if _, err := os.Stat(path); err != nil {
```
rootfs: validate file mode as well? in case path exists but it is a file

tsmetana: Good point. Will fix that.

PR #375, opened 2017-09-20
Update CHANGELOG for heketi block API hash for Gluster Block Provisioner

humblec:
Signed-off-by: hchiramm <hchiramm@redhat.com>


humblec:
@wongma7 The tag is `1.0.5`.

wongma7:
Why do we want to increment the version 5x? That begs the question what happened to versions 1-4.

If we want to signify a new feature that is what semver is for imo.

humblec:
@wongma7 we have tar for `1.0.1` and `1.0.2` , to have a clear view I put it on `1.0.5`, any way I updated the patch set for `1.0.3`, PTAL.

PR #376, opened 2017-09-20
NFS: Improve error reporting for multiple service endpoints 

miminar:
When the provisioner is deployed as a daemonset, its service has one address for each pod deployed. The IP selection shouldn't assume just one address.

Example of endpoints object on OpenShift when the nfs-provisioner is deployed as a daemonset to two nodes:

```
apiVersion: v1
items:
- apiVersion: v1
  kind: Endpoints
  metadata:
    name: nfs-provisioner
    namespace: ns
  subsets:
  - addresses:
    - ip: 10.128.2.12
      nodeName: f25-ose2.vm
      targetRef:
        kind: Pod
        name: nfs-provisioner-4-sp7ch
        namespace: ns
    - ip: 10.130.0.9
      nodeName: f25-ose2.vm
      targetRef:
        kind: Pod
        name: nfs-provisioner-4-xbxhq
        namespace: ns
    ports:
    - name: 2049-tcp
      port: 2049
      protocol: TCP
    - name: 20048-tcp
      port: 20048
      protocol: TCP
    - name: 111-tcp
      port: 111
      protocol: TCP
    - name: 111-udp
      port: 111
      protocol: UDP
```

Maybe I am missing something and there's a good reason for ignoring multi-address endpoints. If so, feel free to close this with some brief explanation.

### DIFF
--- a/nfs/pkg/volume/provision.go
+++ b/nfs/pkg/volume/provision.go
@@ -385,6 +385,8 @@ func (p *nfsProvisioner) getServer() (string, error) {
 	}
 	endpoints, err := p.client.Core().Endpoints(namespace).Get(serviceName, metav1.GetOptions{})
 	for _, subset := range endpoints.Subsets {
+		// One service can't have multiple nfs-provisioner endpoints. If it had, kubernetes would round-robin
+		// the request which would probably go to the wrong instance.
 		if len(subset.Addresses) != 1 {
 			continue
 		}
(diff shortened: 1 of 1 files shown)


erikbsap:
Thanks for making this PR, miminar. I'm interested in this as well.

Usecase: I have a cluster with 10 nodes, each of them has about 50 GB to spare. So I'd like to provide the nfs-provisioner on each of these nodes, but put all behind one service. All of these nodes are bare metal in a data center not under my control, so just repprovisioning the 50 GB to the other hosts would not be easy, maybe impossible if they are not separate disks.

wongma7:
One service can't have multiple nfs-provisioner endpoints because if somebody tries to mount using the service IP, kubernetes will round-robin the request and it will probably go to the wrong instance.

The docs talk about using nodePort as a workaround, the downside being that now you're taking up & exposing the node's entire nfs port. I don't recall any other workaround.

miminar:
Thank you @wongma7 for the explanation. I've changed the PR to reword the error reporting to make the cause more clear.

wongma7:
/lgtm
/area/nfs

thank you

PR #377, opened 2017-09-21
iscsi-targetd: support for Chap authentication

raffaelespazzoli:
this PR add support for chap authentication.
The ansible installer sets up chap authentication for discovery in `target` and stores the chap authentication credentials for session in a secret for later use by the provisioner.
It also sets up a storage class with chap authentication if configured.

The provisioner sets up chap session authentication in `target` using the previously configure secret via `targetd`. This call currently does not seem to work, see also this: https://bugzilla.redhat.com/show_bug.cgi?id=1490631

the project tenant is responsible to set up the secret that the iscsi storage provider will use. This secret will contain both discovery and session authentication.

The code in the PR should be fine. I'm still investigating the above bug. Any help with that is appreciated.


wongma7:
/assign @humblec 
PTAL.

looks fine to me but I know nothing about CHAP

raffaelespazzoli:
@wongma7 @humblec could we merge this? It there anything stopping it?

wongma7:
/lgtm

no let's get it merged,

PR #381, opened 2017-09-22
Snapshot: Unit test for snapshotter, makefile test target

tsmetana:
This is a basic unit test for snapshotter (the most complex part of snapshot controller). It's not complete but tests some of the important code paths. I have also fixed the cloudprovider tests so they compile and added makefile test targets for snapshot which should help with basic CI testing (something needs to compile before the merge...).


snapshot/pkg/controller/snapshotter/snapshotter_test.go
```diff
@@ -0,0 +1,439 @@
...
+	if err == nil {
+		t.Errorf("Test failed, expected error got nil")
+	}
+
+	if tp.DeleteCallCount != 2 {
+		t.Errorf("Test failed, expected 2 DeleteSnapshot calls in plugin, got %d", tp.CreateCallCount)
+	}
+}
+
+func Test_createSnapshot(t *testing.T) {
```
rootfs: nit: Test_createSnapshotData

tsmetana: Will rename it. Thanks.

wongma7:
you may add `make test-snapshot` here https://github.com/kubernetes-incubator/external-storage/blob/master/test.sh#L95

we may also want to make snapshot its own "TEST_SUITE" https://github.com/kubernetes-incubator/external-storage/blob/master/.travis.yml

(I am probably going to slim down the "everything-else" test suite, though the 'nfs e2e tests' are still the bottleneck at ~15 mins)

tsmetana:
The tests fail to compile. Problem is that the failure doesn't seem to originate in our code:
```
../vendor/github.com/stretchr/testify/assert/assertions.go:1205: unknown field 'DisablePointerAddresses' in struct literal of type spew.ConfigState

../vendor/github.com/stretchr/testify/assert/assertions.go:1206: unknown field 'DisableCapacities' in struct literal of type spew.ConfigState
```

rootfs:
assertion is used by aws_test.go

rootfs:
need latest go-spew per https://github.com/stretchr/testify/issues/378

wongma7:
Add go spew to glide.yaml to force it to get the latest version. if package a wants an old version of c and b a new one, glide will tend to pick the old.

tsmetana:
@wongma7 thanks. I'll try it.

PR #382, opened 2017-09-22
Snapshot: add some kubernetes deployments and make container target

tsmetana:
This should help using snapshot controller: Makefile in snapshot directory can now create containers and I have added also some deployments for kubernets (HostPath and AWS).

I have bundled a small fix for snapshotter: we ignored error from API server when GETting the VolumeSnapshot. That lead to somewhat mysterious errors.

PTAL.

### DIFF
--- a/snapshot/Makefile
+++ b/snapshot/Makefile
@@ -12,16 +12,47 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-.PHONY: all controller clean test
+ifeq ($(REGISTRY),)
+	REGISTRY = quay.io/external_storage/
+endif
+ifeq ($(VERSION),)
+	VERSION = latest
+endif
+IMAGE_CONTROLLER = $(REGISTRY)snapshot-controller:$(VERSION)
+MUTABLE_IMAGE_CONTROLLER = $(REGISTRY)snapshot-controller:latest
+IMAGE_PROVISIONER = $(REGISTRY)snapshot-provisioner:$(VERSION)
+MUTABLE_IMAGE_PROVISIONER = $(REGISTRY)snapshot-provisioner:latest
 
-all: controller
+.PHONY: all controller provisioner clean container container-quick push test
+
+all: controller provisioner
 
 controller:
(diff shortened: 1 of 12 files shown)


snapshot/deploy/kubernetes/aws/snapshot-rbac.yaml
```diff
@@ -0,0 +1,48 @@
...
+    resources: ["persistentvolumeclaims"]
+    verbs: ["get", "list", "watch", "update"]
+  - apiGroups: ["storage.k8s.io"]
+    resources: ["storageclasses"]
+    verbs: ["get", "list", "watch"]
+  - apiGroups: [""]
+    resources: ["events"]
+    verbs: ["list", "watch", "create", "update", "patch"]
+  - apiGroups: [""]
+    resources: ["services", "endpoints"]
```
rootfs: why do we need services and endpoints?

tsmetana: We don't. I just started with nfs as template and didn't realize this part should go away.

rootfs: i don't know if nfs deploy needs services/endpoints @wongma7

wongma7:
Yes, only the nfs needs these

snapshot/Makefile
```diff
@@ -12,13 +12,32 @@
...
 	go build -o _output/bin/snapshot-provisioner cmd/snapshot-pv-provisioner/snapshot-pv-provisioner.go
 
 clean:
+	-rm -f deploy/snapshot-controller
+	-rm -f deploy/snapshot-provisioner
 	-rm -rf _output
+
+container: controller quick-container
+
+container-quick:
```
jsafrane: it's `quick-container` on the previous line

tsmetana: Oops. I could not decide how to name the target... And this is the outcome. Will fix it.

tsmetana:
So... any chance of merging this one? Is there still something I should change?

PR #387, opened 2017-09-25
[local-volume] Best practices regarding local volume discovery

ianchakeres:
In this PR I describe best practices for adding new volumes to be discovered and PVs created. These best practices avoid race conditions and the fact that local volumes leverage node affinity using the node name.

### DIFF
--- a/local-volume/README.md
+++ b/local-volume/README.md
@@ -231,6 +231,21 @@ go run hack/e2e.go -- -v --test --test_args="--ginkgo.focus=\[Feature:LocalPersi
 * Avoid recreating nodes with the same node name while there are still old PVs
   with that node's affinity specified. Otherwise, the system could think that
   the new node contains the old PVs.
+* While leveraging the local PV provisioner, race conditions can occur when
+  an admin is adding new volumes and they are being discovered. To avoid 
+  this situation, in the discovery directories we recommend utilizing bind
+  mounts for mount points and symlinks for raw block volumes. For example, 
+  if the discovery directory is /mnt/volumes and we want the provisioner 
+  to expose /dev/sda1 as a volume named vol1, then we would bind mount 
+  /dev/sda1 at /mnt/volumes/vol1.   
+* For volumes with a filesystem, its recommended to utilize their UUID (e.g.
+  the output from `ls -l /dev/disk/by-uuid`) both in fstab mounting logic and
+  in the directory name representing that volume. This practice ensures
+  that the wrong local volume is not mistakenly mounted, even if its mountpoint
+  changes (e.g. if /dev/sda1 becomes /dev/sdb1 when a new disk is added).
+  Additionally, this practice will ensure that if another node with the
+  same name is created, that any volumes on that node are unique and not
+  mistaken for a volume on another node with the same name.
 
 ### Deleting/removing the underlying volume
 
--- a/local-volume/README.md
+++ b/local-volume/README.md
@@ -224,6 +224,12 @@ go run hack/e2e.go -- -v --test --test_args="--ginkgo.focus=\[Feature:LocalPersi
 [GCE Alpha](https://k8s-testgrid.appspot.com/sig-storage#gce-alpha)
 [GCE GCI Alpha](https://k8s-testgrid.appspot.com/sig-storage#gci-gce-alpha)
 
+
+## Requirements
+
+* The local-volume plugin expects paths to be stable, including across 
+  reboots and when disks are added or removed.
+
 ## Best Practices
 
 * For IO isolation, a whole disk per volume is recommended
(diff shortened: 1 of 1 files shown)


local-volume/README.md
```diff
@@ -200,6 +200,21 @@ go run hack/e2e.go -- -v --test --test_args="--ginkgo.focus=\[Feature:LocalPersi
 * Avoid recreating nodes with the same node name while there are still old PVs
   with that node's affinity specified. Otherwise, the system could think that
   the new node contains the old PVs.
+* While leveraging the local PV provisioner, race conditions can occur when
```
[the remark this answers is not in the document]
ianchakeres: @msau42 - at the bottom of this bullet, I added a sentence and link to the related Kubernetes issue. Does this look good to merge now?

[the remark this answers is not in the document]
msau42: Thinking about this some more, I don't think there is a workaround for this issue without a code change.  Even bind mounting requires that you create the destination directory beforehand.  So we must have a code change to only discover mounted directories.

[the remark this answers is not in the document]
ianchakeres: How about I remove the statements about race conditions, since we don't have a solution yet.

This change would leave our recommendations saying that bindmount should be used for mount points and symlinks should be used for block devices.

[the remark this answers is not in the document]
msau42: I think we can just leave out the whole statement.  The final solution is looking like: bind mount for shared directories, and regular mount for mount points.  I don't think symlink is necessary for block.

msau42:
I think we should have a section about requirements though, mentioning things like the paths need to be stable across reboots and if you add/delete disks to the node.

local-volume/README.md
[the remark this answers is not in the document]
ianchakeres: I removed the whole statement, including recommendations on bind mount vs regular mount.

ianchakeres:
@msau42 - I added a brief requirements section, that states the local-volume plugin expectations for stability of paths.

local-volume/README.md
```diff
@@ -224,13 +224,27 @@ go run hack/e2e.go -- -v --test --test_args="--ginkgo.focus=\[Feature:LocalPersi
...
 
 * For IO isolation, a whole disk per volume is recommended
 * For capacity isolation, separate partitions per volume is recommended
 * Avoid recreating nodes with the same node name while there are still old PVs
   with that node's affinity specified. Otherwise, the system could think that
   the new node contains the old PVs.
+* For volumes with a filesystem, its recommended to utilize their UUID (e.g.
+  the output from `ls -l /dev/disk/by-uuid`) both in fstab mounting logic and
+  in the directory name representing that volume. This practice ensures
+  that the wrong local volume is not mistakenly mounted, even if its mountpoint
```
msau42: even if it's device path changes

local-volume/README.md
```diff
@@ -224,13 +224,27 @@ go run hack/e2e.go -- -v --test --test_args="--ginkgo.focus=\[Feature:LocalPersi
...
 ## Best Practices
 
 * For IO isolation, a whole disk per volume is recommended
 * For capacity isolation, separate partitions per volume is recommended
 * Avoid recreating nodes with the same node name while there are still old PVs
   with that node's affinity specified. Otherwise, the system could think that
   the new node contains the old PVs.
+* For volumes with a filesystem, its recommended to utilize their UUID (e.g.
+  the output from `ls -l /dev/disk/by-uuid`) both in fstab mounting logic and
+  in the directory name representing that volume. This practice ensures
```
msau42: The mount directory name doesn't need to use the uuid

local-volume/README.md
```diff
@@ -224,13 +224,27 @@ go run hack/e2e.go -- -v --test --test_args="--ginkgo.focus=\[Feature:LocalPersi
...
+
 ## Best Practices
 
 * For IO isolation, a whole disk per volume is recommended
 * For capacity isolation, separate partitions per volume is recommended
 * Avoid recreating nodes with the same node name while there are still old PVs
   with that node's affinity specified. Otherwise, the system could think that
   the new node contains the old PVs.
+* For volumes with a filesystem, its recommended to utilize their UUID (e.g.
+  the output from `ls -l /dev/disk/by-uuid`) both in fstab mounting logic and
```
msau42: fstab entries

local-volume/README.md
ianchakeres: The mount directory name doesn't need to use the UUID. If the directory name does contain the UUID, then it would not be confused with any other volume as the directory name is used while generating the PV name. 

This practice would help avoid the situation you've raised where the same PV name is used for two different volumes, when the directory name and node name are the same even if the volume is different. 

Would you like me to remove this recommendation?

msau42: I agree, let's leave it.

ianchakeres:
@msau42 - PTAL updated.

local-volume/README.md
```diff
@@ -224,13 +224,27 @@ go run hack/e2e.go -- -v --test --test_args="--ginkgo.focus=\[Feature:LocalPersi
...
+
 ## Best Practices
 
 * For IO isolation, a whole disk per volume is recommended
 * For capacity isolation, separate partitions per volume is recommended
 * Avoid recreating nodes with the same node name while there are still old PVs
   with that node's affinity specified. Otherwise, the system could think that
   the new node contains the old PVs.
+* For volumes with a filesystem, its recommended to utilize their UUID (e.g.
+  the output from `ls -l /dev/disk/by-uuid`) both in fstab entries mounting logic
```
msau42: remove "mounting logic"

local-volume/README.md
```diff
@@ -224,13 +224,27 @@ go run hack/e2e.go -- -v --test --test_args="--ginkgo.focus=\[Feature:LocalPersi
...
+  and in the directory name representing that volume. This practice ensures
```
msau42: "directory name of the mount point"

local-volume/README.md
```diff
@@ -224,13 +224,27 @@ go run hack/e2e.go -- -v --test --test_args="--ginkgo.focus=\[Feature:LocalPersi
...
+  reboots and when disks are added or removed.
+
 ## Best Practices
 
 * For IO isolation, a whole disk per volume is recommended
 * For capacity isolation, separate partitions per volume is recommended
 * Avoid recreating nodes with the same node name while there are still old PVs
   with that node's affinity specified. Otherwise, the system could think that
   the new node contains the old PVs.
+* For volumes with a filesystem, its recommended to utilize their UUID (e.g.
```
msau42: it's

ianchakeres:
@wongma7 - PTAL

PR #388, opened 2017-09-25
iscsi: Use centos image for the base image

nak3:
This patch replace rhel7-atmoic base image with centos7 in Dockerfile.

### DIFF
--- a/iscsi/targetd/Dockerfile
+++ b/iscsi/targetd/Dockerfile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM registry.access.redhat.com/rhel7-atomic:latest
+FROM centos:7
 LABEL authors="Raffaele Spazzoli <rspazzol@redhat.com>" 
 ADD iscsi-controller /iscsi-controller
 ENTRYPOINT ["/iscsi-controller"]


wongma7:
/area iscsi/targetd
@raffaelespazzoli please taek a look at this

raffaelespazzoli:
I don't think it's going to be an issue. 
But why are we doing it? License issues? 
I would only recommend if possible to use a lightweight image. not sure if the centos family has one.

nak3:
The license is the reason for this patch. When accessed `registry.access.redhat.com/rhel-atomic` repo via below site, we can see the message of RH subscription agreement:

https://access.redhat.com/containers?tab=images&platform=docker#/registry.access.redhat.com/rhel-atomic

> Before downloading or using this Container, you must agree to the Red Hat subscription agreement located at redhat.com/licenses. If you do not agree with these terms, do not download or use the Container. If you have an existing Red Hat Enterprise Agreement (or other negotiated agreement with Red Hat) with terms that govern subscription services associated with Containers, then your existing agreement will control.

Although I haven't gone through the RH subscription agreement mentioned here, I feel that it is not good idea to use an image which we must agree to such company agreement within the community project.

And since (as far as I know) CentOS does not release the official centos-atomic image, I have sent this patch to use centos:7.

raffaelespazzoli:
@nak3 ok, thanks. I have no issue with the update. Could someone push a new image to quay, after the merge?

wongma7:
/lgtm

I'll push a v0.0.3

PR #389, opened 2017-09-25
iscsi: Change default provisioner name of storageclass to iscsi-targetd

nak3:
This patch changes default provisioner-name to `iscsi`, as all of PVC
and storageClass examples in this project defined the provisioner as
`iscsi`.


nak3:
For example, [iscsi-provisioner-pvc.yaml](https://github.com/kubernetes-incubator/external-storage/blob/8732beaa43a3b00d39e2c9ced21f2d662e6a0969/iscsi/targetd/kubernetes/iscsi-provisioner-pvc.yaml#L6)
has `volume.beta.kubernetes.io/storage-class: "iscsi"`. 

Therefore, the check in `shouldProvision()`
https://github.com/kubernetes-incubator/external-storage/blob/8732beaa43a3b00d39e2c9ced21f2d662e6a0969/lib/controller/controller.go#L573-L578
regards inconsistent class.

raffaelespazzoli:
the provisioner name and the storage class name are two different things. not sure about the value of this change, but if you do it all the examples will break. the link between the storage class and the storage provisioner is in the storage class, so at least the example storage classes should be updated.

nak3:
I'm sorry. I made a wrong comment. As pvc's provisioner need to be consistent with storageclass's name, but the patch should have just changed provisioner name in storageclass. I updated the PR.

raffaelespazzoli:
There is still an inconsistency here, now the storage classes look for an `iscsi-provisioner` but in the examples the iscsi-provisioner pod is stated with:
```
            - name: PROVISIONER_NAME
              value: iscsi
```

nak3:
Thank you. Updated.

raffaelespazzoli:
lgtm.

nak3:
ping @wongma7 I'm sorry for bothering you, but I feel that you missed this.

raffaelespazzoli:
@nak3 and @wongma7 yesterday I suggest to use the following name for the provisioner name: `iscsi-targetd`.
So all storage class examples and the default in the code (in start.go) should be set to `iscsi-targetd`
The documentation should also be updated.
The reason is that as discussed before, there could be other iscsi implementations in the future. This works only for targetd.

nak3:
@raffaelespazzoli Sure, sounds good idea. Updated this PR.

mtanino:
@nak3 
Please update title of this PR and commit from iscsi-provisioner to iscsi-targetd too.

nak3:
Sure, thanks. I have done. (If squash of commit is also necessary, please let me know.)

mtanino:
yeah, once @raffaelespazzoli or @wongma7 give lgtm, usually it's time to squash commits to one.

raffaelespazzoli:
Hi, I have been giving more thoughts to this issue (I am restructuring the ansible playbooks and trying to come up with sensible defaults).
Multiple provisioners can be deployed in a cluster and a provisioner can connect to one targetd server, but can create volumes on multiple volume groups. Which volume groups is used is determined by the stroageclass.
So a good naming convention for the storage class would be to associate a provisioner and a volume group, so to avoid clashes in a crwded cluster (unlikely, but possible). Therefore the patter could be: `<provisioner-name>-<volume-group>`
Here are my defaults for the ansible script:
```
volume-group name: vg-targetd
provisioner name: iscsi-targetd
storageClass name: iscsi-targetd-vg-targetd
```
Let me know what you think.

nak3:
I agree that `provisioner-name + volume-group = storage class` would be a good naming convention. It could be unique for multiple volume group name.

But, I am concerned about it when we use thin provisioning. It would be something like:

~~~
storageClass name: iscsi-targetd-vg-targetd/somepool
~~~

It is a little bit long and storage, I feel. But anyway I'll follow your decision.

nak3:
@raffaelespazzoli based on your previous comment, I updated this PR with:

- provisioner name metadata.name in StrageClass of README to `iscsi-targetd`
- default command option's value of `provisioner-name` to `iscsi-targetd`

raffaelespazzoli:
hi,
the changes proposed in this PR have been introduced by another PR:
https://github.com/kubernetes-incubator/external-storage/pull/430

Please verify.

nak3:
Yes, I checked https://github.com/kubernetes-incubator/external-storage/pull/430. But still a few points (REAMDE and command default value) should be updated, I think.

PR #379, opened 2017-09-21
Snapshot: add user guide and move documentation to a common directory

tsmetana:
Here's update of the snapshot documentation: I put everything to one directory named "doc" and added some basic user guide written by @pospispa .

PR #391, opened 2017-09-25
Fix typo and sync names

liangxia:
1. Typo, exiting -> existing
2. Fix URL 404.
3. This example is for GCE, so volume type is "GCE PD".
4. In the same example, some part reference the volumesnapshot name as snapshot-1, the others use snapshot-demo. Make them all using snapshot-demo now. Also the different names is caused by two claim files with different names, so removing the duplicated one.
```
$ pwd
/opt/github/external-storage/snapshot/examples/gce
$ diff fromsnapshotclaim.yaml claim.yaml 
7c7
<     snapshot.alpha.kubernetes.io/snapshot: snapshot-1
---
>     snapshot.alpha.kubernetes.io/snapshot: snapshot-demo
12c12
<       storage: 8Gi
---
>       storage: 4Gi
14d13
< 
```


rootfs:
also see #379

tsmetana:
Right. We should include this one in #379.

tsmetana:
@liangxia I've included the fix in #379 which was merged, so I guess it's OK to close this PR.

PR #392, opened 2017-09-26
Support `Retain` reclaim policy in external storage

### LINKED ISSUE #326: External Controller should support `Retain` volume reclaim policy
As patch https://github.com/kubernetes/kubernetes/pull/47987 is merged, I think we can support `Retain` policy in external-storage.

https://github.com/kubernetes-incubator/external-storage/blob/master/lib/controller/controller.go#L785

I'm working on a patch on this.

/cc @wongma7

xingzhou:
Fixed #326 

@wongma7, please take a review, thanks.


lib/controller/controller.go
```diff
@@ -1106,3 +1110,31 @@ func (ctrl *ProvisionController) getStorageClassFields(name string) (string, map
...
+	case *storage.StorageClass:
+		reclaimPolicy = *class.ReclaimPolicy
+	case *storagebeta.StorageClass:
+		reclaimPolicy = *class.ReclaimPolicy
+	default:
+		return reclaimPolicy, fmt.Errorf("Cannot convert object to StorageClass: %+v", classObj)
+	}
+
+	// At present, the reclaim policy for dynamic provisioned volume only support Retain or Delete.
+	if reclaimPolicy != v1.PersistentVolumeReclaimRetain && reclaimPolicy != v1.PersistentVolumeReclaimDelete {
```
wongma7: it is kubernetes's job to make sure this never happens, we can safely remove it

lib/controller/controller.go
```diff
@@ -1106,3 +1110,31 @@ func (ctrl *ProvisionController) getStorageClassFields(name string) (string, map
...
+		return "", err
+	}
+	if !found {
+		return "", fmt.Errorf("StorageClass %q not found", storageClassName)
+	}
+
+	var reclaimPolicy v1.PersistentVolumeReclaimPolicy
+	switch class := classObj.(type) {
+	case *storage.StorageClass:
+		reclaimPolicy = *class.ReclaimPolicy
```
wongma7: this will cause a nil pointer dereference if the provisioner is running on an older kube version. We need to do an `if version` check before we call `fetchReclaimPolicy`, `else` we keep defaulting to `Delete`. See also https://github.com/kubernetes-incubator/external-storage/pull/392/files#diff-16daa5c3f02eee06add7f1af0a3fa901R370

lib/controller/controller_test.go
```diff
@@ -38,9 +38,9 @@ import (
 	"k8s.io/client-go/kubernetes/fake"
```
wongma7: please edit `newStorageClass` in this file so that it has a reclaimpolicy. Because we want to pretend it has gone through kubernetes defaulting which would set it to Delete for us.

lib/controller/controller.go
xingzhou: So for this case, the version should be v1.8.0, right?

lib/controller/controller.go
xingzhou: sure, for current controller_test, it's only testing the controller's behavior for v1.5.0, I'd like also to add a test for reclaimPolicy for v1.8.0

lib/controller/controller.go
wongma7: yes

lib/controller/controller.go
wongma7: that sgtm

xingzhou:
@wongma7, have updated the patch according to your comments, PTAL, thanks. For testing side, I tried to add one more test case for this specific case, but I think we'd better update or refactor the whole controller tests as there are some version relevant differences in the controller code already, we may want to test different behaviors under different versions.

wongma7:
/lgtm
/area lib

I agree refactoring the tests could be good. Not urgent atm, but good.

thanks! apparently some vendor needs rebase/cleaning though, sorry bout that.

xingzhou:
@wongma7, thanks for the review. I've updated the patch, please let me know if the vendor folder still needs specific handling.
For the testings, I'd like to supplement some test cases for different versions first.

wongma7:
the glusterfs needs to be updated, seems a package was removed from client-go, most likely moved to another k8s/package:

pkg/volume/exec.go:67: undefined: "github.com/kubernetes-incubator/external-storage/vendor/k8s.io/client-go/tools/remotecommand".NewExecutor
pkg/volume/exec.go:77: unknown field 'SupportedProtocols' in struct literal of type "github.com/kubernetes-incubator/external-storage/vendor/k8s.io/client-go/tools/remotecommand".StreamOptions
make[1]: *** [build] Error 2
make[1]: Leaving directory `/home/travis/gopath/src/github.com/kubernetes-incubator/external-storage/gluster/glusterfs'
make: *** [gluster/glusterfs] Error 2

xingzhou:
yes, @wongma7, the client-go code is updated. But while updating the code, I found out that seems we also need to update `k8s.io/apiextensions-apiserver` package to a newer version(the CI process will run `go build -o _output/bin/snapshot-controller cmd/snapshot-controller/snapshot-controller.go` in `snapshot` directory which requires an update of `apiextensions-apiserver` package). But after the upgrade, the `snapshot` package can not be built, the errors on my local laptop are like:
```
# github.com/kubernetes-incubator/external-storage/snapshot/pkg/apis/crd/v1
pkg/apis/crd/v1/register.go:52: cannot use VolumeSnapshot literal (type *VolumeSnapshot) as type runtime.Object in argument to scheme.AddKnownTypes:
	*VolumeSnapshot does not implement runtime.Object (missing DeepCopyObject method)
```
This is sth related to the update of package `apimachinery` 's version. But I can see that seems `snapshot` is the only package that depends on `apiextensions-apiserver` package, and `snapshot `controller relies on its own make process, why not add an independent `glide.yaml` to manage `snapshot`'s own dependencies?

Or we may need to update `snapshot` code to fix the above build issue.

wongma7:
Updating snapshot code is the more palatable action.

@rootfs can you PTAL at this? we are upgrading the whole repo to 1.8 kube dependencies, and it seems CRDs were changed recently so they have to implement this new DeepCopyObject method:
```
# github.com/kubernetes-incubator/external-storage/snapshot/pkg/apis/crd/v1
pkg/apis/crd/v1/register.go:52: cannot use VolumeSnapshot literal (type *VolumeSnapshot) as type runtime.Object in argument to scheme.AddKnownTypes:
	*VolumeSnapshot does not implement runtime.Object (missing DeepCopyObject method)
```
Thanks

humblec:
@xingzhou Can you please revisit this PR ? more users are requesting this feature.

xingzhou:
@humblec, I think the commits included in this PR has been merged here: https://github.com/kubernetes-incubator/external-storage/pull/419/commits

and, from these lines in the code, the reclaim policy shall be extracted and put into the volume options for provisioner:
https://github.com/kubernetes-incubator/external-storage/blob/master/lib/controller/controller.go#L870-L887

nrobert13:
@xingzhou I'm trying reclaimPolicy: Retain with nfs-client-provisioner latest docker image, but the pv's still have delete.
```
$ kubectl get pvc
NAME         STATUS    VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS          AGE
test-claim   Bound     pvc-bb3abcd2-215a-11e8-837d-0050560150de   1Mi        RWX            managed-nfs-storage   3s
$ kubectl get pv
NAME                                       CAPACITY   ACCESS MODES   RECLAIM POLICY   STATUS    CLAIM                STORAGECLASS          REASON    AGE
nfs-data-mysql-0                           1Gi        RWX            Retain           Bound     mysql/data-mysql-0                                   4d
nfs-data-mysql-1                           1Gi        RWX            Retain           Bound     mysql/data-mysql-1                                   4d
nfs-data-mysql-2                           1Gi        RWX            Retain           Bound     mysql/data-mysql-2                                   4d
pvc-bb3abcd2-215a-11e8-837d-0050560150de   1Mi        RWX            Delete           Bound     default/test-claim   managed-nfs-storage             6s
$ kubectl describe storageclass 
Name:            managed-nfs-storage
IsDefaultClass:  No
Annotations:     <none>
Provisioner:     plusserver/nfs
Parameters:      <none>
ReclaimPolicy:   Retain
Events:          <none>
```

qianlei90:
@nrobert13 +1, any solution?

nrobert13:
@qianlei90 , none from my side.

wongma7:
@qianlei90 @nrobert13 the code has been merged in another PR but many of the provisioner images have not been updated. Sorry for the delay, nfs-client I will update by tomorrow

PR #400, opened 2017-09-28
Refactor the snapshotter code for more robustness.

jingxu97:
The main logic in snapshotter is as follows

```
syncSnapshot is the main controller method to decide what to do with a snapshot object. 

Step1. [Check snapshot condition] It first checks the status of this snapshot object from its VolumeSnapshotCondition field.  

- If the condition is Ready or Failed, do nothing
- If the condition is Pending, go to step 3
- If there is no condition recorded yet, check the following

  1. Check if snapshot object metadata has label of “snapshotMetadataTimeStamp”. If no such label, it indicates that snapshot has not yet triggered, return statusNew which will continues with step 2.
  2. If the label exist, check if there is already a VolumeSnapshotData object exist that references the given snapshot. If this object exists, update snapshot object to bind the VolumeSnapshotData with it. Return statusPending which continues with step 3
  3. If no VolumeSnapshotData object, use the metadata information to construct tag and query the cloud whether a snapshot is already taken with the tag. If such snapshot can be found, construct the VolumeSnapshotData object and bind it with VolumeSnapshot object. Return statusPending which continues with step 3.
  4. If no snapshot is found,  return statusNew which will continue with step 2.



Step2. [Take snapshot and Create SnapshotData object to bind]

1. **getPVFromVolumeSnapshot**: From snapshot data object, get PVC name and then retrieve pv object from API server. If failed, return error.
2. **updateVolumeSnapshotMetadata**: Construct tag information with current timestamp and VolumeSnapshot object’s metadata (namespace, name, and uid). Also update the VolumeSnapshot object with the timestamp and pv name (which will be used later). If update failed, return error.
3. **takeSnapshot**: call cloud provider to take the snapshot and attach the tags to the snapshot constructed from above. If it failed, return the error. Otherwise it returns the snapshot data source information and condition.
4. **createVolumeSnapshotData**: The name of the VolumeSnapshotData object is constructed from the UID of the VolumeSnapshot for easy query at later point. If object is constructed and updated to API server successfully, move to the next step. Otherwise, return error.
5. **bindandUpdateVolumeSnapshot**: It updates the VolumeSnapshot object with the VolumeSnapshotData object name and new status.

Step3. [Wait until snapshot is ready or failed] This step creates a loop with exponential backoff time. Inside the loop

1. It queries the cloud provider to describe the snapshot with the status information.
2. It updates the VolumeSnapshot object with the new condition if there is any changes compared with the old condition
3. If the latest status is ready or error, break the loop and return.
```

This PR also includes the changes
1. SnapshotCreate function in gce
2. Remove Namespace for VolumeSnapshotData object

### DIFF
--- a/snapshot/pkg/controller/reconciler/reconciler.go
+++ b/snapshot/pkg/controller/reconciler/reconciler.go
@@ -23,11 +23,8 @@ import (
 	"time"
 
 	"github.com/golang/glog"
-	v1 "k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 
-	crdv1 "github.com/kubernetes-incubator/external-storage/snapshot/pkg/apis/crd/v1"
 	"github.com/kubernetes-incubator/external-storage/snapshot/pkg/controller/cache"
 	"github.com/kubernetes-incubator/external-storage/snapshot/pkg/controller/snapshotter"
 )
@@ -128,26 +125,6 @@ func (rc *reconciler) reconcile() {
 			// It's likely that the operation exists already: it should be fired by the controller right
 			// after the event has arrived.
 			rc.snapshotter.CreateVolumeSnapshot(snapshot)
-		} else {
-			// The VolumeSnapshot is in both states of world: check its OK and fix it eventually.
-			aswSnapshot := rc.actualStateOfWorld.GetSnapshot(name)
-			if aswSnapshot.Spec.SnapshotDataName == "" {
-				// Seems the write to API server failed before. Find the SnapshotData and fix
-				// this VolumeSnapshot reference and status
-				glog.Infof("Volume snapshot %s is missing the snapshot data name - updating.", name)
-				snapConditions := []crdv1.VolumeSnapshotCondition{
-					{
-						Type:               crdv1.VolumeSnapshotConditionReady,
-						Status:             v1.ConditionTrue,
-						Message:            "Snapshot created succsessfully",
-						LastTransitionTime: metav1.Now(),
-					},
-				}
-				_, err := rc.snapshotter.UpdateVolumeSnapshot(name, &snapConditions)
-				if err != nil {
-					glog.Errorf("Error updating VolumeSnapshot %s: %v", name, err)
-				}
-			}
 		}
 	}
 }
--- a/snapshot/pkg/controller/snapshotter/snapshotter.go
+++ b/snapshot/pkg/controller/snapshotter/snapshotter.go
@@ -22,6 +22,9 @@ import (
 
 	"github.com/golang/glog"
 
+	crdv1 "github.com/kubernetes-incubator/external-storage/snapshot/pkg/apis/crd/v1"
+	"github.com/kubernetes-incubator/external-storage/snapshot/pkg/controller/cache"
+	"github.com/kubernetes-incubator/external-storage/snapshot/pkg/volume"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -31,22 +34,22 @@ import (
 	"k8s.io/client-go/rest"
 	"k8s.io/kubernetes/pkg/util/goroutinemap"
 	"k8s.io/kubernetes/pkg/util/goroutinemap/exponentialbackoff"
-
-	crdv1 "github.com/kubernetes-incubator/external-storage/snapshot/pkg/apis/crd/v1"
-	"github.com/kubernetes-incubator/external-storage/snapshot/pkg/controller/cache"
-	"github.com/kubernetes-incubator/external-storage/snapshot/pkg/volume"
 )
 
 const (
+	snapshotMetadataTimeStamp        = "SnapshotMetadata-Timestamp"
+	snapshotMetadataPVName           = "SnapshotMetadata-PVName"
+	snapshotDataNamePrefix           = "k8s-volume-snapshot"
+	pvNameLabel                      = "pvName"
 	defaultExponentialBackOffOnError = true
 
 	// volumeSnapshot* is configuration of exponential backoff for
-	// waiting for snapshot operation to complete. Starting with 10
-	// seconds, multiplying by 1.2 with each step and taking 15 steps at maximum.
-	// It will time out after 12.00 minutes.
-	volumeSnapshotInitialDelay = 10 * time.Second
-	volumeSnapshotFactor       = 1.2
-	volumeSnapshotSteps        = 15
+	// waiting for snapshot operation to complete. Starting with 2
+	// seconds, multiplying by 1.5 with each step and taking 20 steps at maximum.
+	// It will time out after 20 steps (6650 seconds).
+	volumeSnapshotInitialDelay = 2 * time.Second
+	volumeSnapshotFactor       = 1.5
+	volumeSnapshotSteps        = 20
 )
 
 // VolumeSnapshotter does the "heavy lifting": it spawns goroutines that talk to the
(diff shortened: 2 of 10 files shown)


tsmetana:
Looks OK. I was not sure for a while that we might not try to create the VolumeSnapshotData from two places simultaneously (createSnapshot and updateSnapshotIfExists) but it looks to be safe.

snapshot/pkg/controller/snapshotter/snapshotter.go
```diff
@@ -38,15 +38,18 @@ import (
...
 
 	// volumeSnapshot* is configuration of exponential backoff for
-	// waiting for snapshot operation to complete. Starting with 10
-	// seconds, multiplying by 1.2 with each step and taking 15 steps at maximum.
-	// It will time out after 12.00 minutes.
-	volumeSnapshotInitialDelay = 10 * time.Second
-	volumeSnapshotFactor       = 1.2
-	volumeSnapshotSteps        = 15
+	// waiting for snapshot operation to complete. Starting with 2
+	// seconds, multiplying by 2 with each step and taking 15 steps at maximum.
```
xing-yang: volumeSnapshotFactor = 1.5 and volumeSnapshotSteps = 20, so it should be:

multiplying by 1.5 with each step and taking 20 steps at maximum.

snapshot/pkg/controller/snapshotter/snapshotter.go
```diff
@@ -38,15 +38,18 @@ import (
...
+	// It will time out after 15 steps.
+	volumeSnapshotInitialDelay = 2 * time.Second
+	volumeSnapshotFactor       = 1.5
+	volumeSnapshotSteps        = 20
```
xing-yang: Based on my calculation, starting with 2 seconds (at step 0), and multiplying by 1.5 with each step, it will be 6650 seconds after 20 steps.  Do we really want to wait that long?

snapshot/pkg/controller/snapshotter/snapshotter.go
```diff
@@ -21,11 +21,11 @@ import (
 	"time"
 
 	"github.com/golang/glog"
+	"github.com/pborman/uuid"
```
xing-yang: Is it better to use "github.com/pborman/uuid" than "k8s.io/apimachinery/pkg/util/uuid”?

snapshot/pkg/controller/snapshotter/snapshotter.go
jingxu97: I don't have data about how long a snapshot might take? But from what I heard is that is it possible snapshot might take hours..

snapshot/pkg/controller/snapshotter/snapshotter.go
```diff
@@ -38,15 +38,18 @@ import (
...
 	// volumeSnapshot* is configuration of exponential backoff for
-	// waiting for snapshot operation to complete. Starting with 10
-	// seconds, multiplying by 1.2 with each step and taking 15 steps at maximum.
-	// It will time out after 12.00 minutes.
-	volumeSnapshotInitialDelay = 10 * time.Second
-	volumeSnapshotFactor       = 1.2
-	volumeSnapshotSteps        = 15
+	// waiting for snapshot operation to complete. Starting with 2
+	// seconds, multiplying by 2 with each step and taking 15 steps at maximum.
+	// It will time out after 15 steps.
```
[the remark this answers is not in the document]
jingxu97: fixed

snapshot/pkg/controller/snapshotter/snapshotter.go
jingxu97: fixed

snapshot/pkg/controller/snapshotter/snapshotter.go
xing-yang: That's possible depending on how much data needs to be copied over.  So I'm fine with the total time here.

rootfs:
there is a build failure
```
pkg/controller/snapshotter/snapshotter_test.go:432: cannot use fakeNewVolumeSnapshot() (type *"github.com/kubernetes-incubator/external-storage/snapshot/pkg/apis/crd/v1".VolumeSnapshot) as type string in argument to vs.createVolumeSnapshotData
```

snapshot/pkg/controller/snapshotter/snapshotter.go
```diff
@@ -375,169 +350,209 @@ func (vs *volumeSnapshotter) getPlugin(snapshotName string, snapshot *crdv1.Volu
 	return &plugin, nil
 }
 
-func (vs *volumeSnapshotter) updateSnapshotIfExists(snapshot *crdv1.VolumeSnapshot) (string, *crdv1.VolumeSnapshotData, error) {
-	var snapshotName string = snapshot.Metadata.Name
+func (vs *volumeSnapshotter) findVolumeSnapshotData(snapshot *crdv1.VolumeSnapshot) (*crdv1.VolumeSnapshotData, error) {
```
rootfs: same as `getSnapshotDataFromSnapshot`?

snapshot/pkg/controller/snapshotter/snapshotter.go
```diff
@@ -375,169 +350,209 @@ func (vs *volumeSnapshotter) getPlugin(snapshotName string, snapshot *crdv1.Volu
...
+	}
+	// Find snapshot through cloud provider by existing tags, and create VolumeSnapshotData if such snapshot is found
+	snapshotDataSource, conditions, err = vs.findSnapshotByTags(snapshotName, snapshot)
 	if err != nil {
-		return statusNew, nil, nil
+		return statusNew, snapshot, nil
+	}
+	// Snapshot is found. Create VolumeSnapshotData, bind VolumeSnapshotData to VolumeSnapshot, and update VolumeSnapshot status
+	glog.Infof("updateSnapshotIfExists: create VolumeSnapshotData object for VolumeSnapshot %s.", uniqueSnapshotName)
+	pvName, ok := snapshot.Metadata.Labels["pvName"]
```
rootfs: add a const for `pvName` too?

snapshot/pkg/controller/reconciler/reconciler.go
```diff
@@ -130,7 +127,7 @@ func (rc *reconciler) reconcile() {
 			rc.snapshotter.CreateVolumeSnapshot(snapshot)
 		} else {
 			// The VolumeSnapshot is in both states of world: check its OK and fix it eventually.
-			aswSnapshot := rc.actualStateOfWorld.GetSnapshot(name)
```
rootfs: log it here with info about snapshot data and timestamp

xing-yang:
Looks good other than the Travis CI failure.

tsmetana:
@jingxu97 please rebase to upstream HEAD (it should go without conflict), then the "everything-else" test suite failure fix is trivial:
```
diff --git a/snapshot/pkg/controller/snapshotter/snapshotter_test.go b/snapshot/pkg/controller/snapshotter/snapshotter_test.go
index a509d5a..bd10934 100644
--- a/snapshot/pkg/controller/snapshotter/snapshotter_test.go
+++ b/snapshot/pkg/controller/snapshotter/snapshotter_test.go
@@ -429,7 +429,7 @@ func Test_createSnapshotData(t *testing.T) {
                        Type:               crdv1.VolumeSnapshotConditionReady,
                },
        }
-       retData, err := vs.createVolumeSnapshotData("default/new-snapshot-test-1", fakeNewVolumeSnapshot(), &snapDataSource, &snapConditions)
+       retData, err := vs.createVolumeSnapshotData("default/new-snapshot-test-1", "fake-pv-1", &snapDataSource, &snapConditions)
        if err != nil {
                t.Errorf("Test failed, unexpected error: %v", err)
        }
```

The failed nfs test looks like a network hiccup...

snapshot/pkg/controller/snapshotter/snapshotter.go
jingxu97: fixed

snapshot/pkg/controller/reconciler/reconciler.go
jingxu97: I actually removed the whole else part. If snapshot is already in actual state, there is nothing to do. We make sure the snapshot is already ready before updating it into the actual state.

PR #401, opened 2017-10-02
don't use dash in volume snapshot

rootfs:
per findings from @jingxu97, we don't need dash in volume snapshot to use capital letter in VolumeSnapshot object after switching to CRD.
 
@xing-yang @tsmetana PTAL, thanks

### DIFF
--- a/snapshot/pkg/apis/crd/v1/register.go
+++ b/snapshot/pkg/apis/crd/v1/register.go
@@ -23,7 +23,7 @@ import (
 )
 
 // GroupName is the group name use in this package.
-const GroupName = "volume-snapshot-data.external-storage.k8s.io"
+const GroupName = "volumesnapshotdata.external-storage.k8s.io"
 
 var (
 	// SchemeBuilder is the new scheme builder
--- a/snapshot/pkg/apis/crd/v1/register.go
+++ b/snapshot/pkg/apis/crd/v1/register.go
@@ -23,7 +23,7 @@ import (
 )
 
 // GroupName is the group name use in this package.
-const GroupName = "volumesnapshotdata.external-storage.k8s.io"
+const GroupName = "volumesnapshot.external-storage.k8s.io"
 
 var (
 	// SchemeBuilder is the new scheme builder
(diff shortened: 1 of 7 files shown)


snapshot/pkg/apis/crd/v1/types.go
```diff
@@ -27,12 +27,12 @@ import (
 const (
 	// VolumeSnapshotDataResourcePlural is "volumesnapshotdatas"
 	VolumeSnapshotDataResourcePlural = "volumesnapshotdatas"
-	// VolumeSnapshotDataResource is "volume-snapshot-data"
-	VolumeSnapshotDataResource = "volume-snapshot-data"
+	// VolumeSnapshotDataResource is "volumesnapshotdata"
+	VolumeSnapshotDataResource = "volumesnapshotdata"
```
jingxu97: cannot find any place uses this constant

snapshot/pkg/apis/crd/v1/types.go
```diff
@@ -27,12 +27,12 @@ import (
...
 	// VolumeSnapshotResourcePlural is "volumesnapshots"
 	VolumeSnapshotResourcePlural = "volumesnapshots"
-	// VolumeSnapshotResource is "volume-snapshot"
-	VolumeSnapshotResource = "volume-snapshot"
+	// VolumeSnapshotResource is "volumesnapshot"
+	VolumeSnapshotResource = "volumesnapshot"
```
jingxu97: same as here

snapshot/pkg/apis/crd/v1/register.go
```diff
@@ -23,7 +23,7 @@ import (
 )
 
 // GroupName is the group name use in this package.
-const GroupName = "volume-snapshot-data.external-storage.k8s.io"
+const GroupName = "volumesnapshotdata.external-storage.k8s.io"
```
jingxu97: is the word "data" needed here?

snapshot/pkg/apis/crd/v1/types.go
rootfs: right, another TPR legacy. Will remove

xing-yang:
lgtm

PR #402, opened 2017-10-03
nfs-provisioner - added maxExports flag

aka-bo:
The `maxExports` flag controls the maximum number of volumes a provisioner
is allowed to export. Once this limit is reached new claims will be
ignored with the hope that another provisioner will provision it.

Omitting this flag (or setting to a negative value) will disable this
feature, effectively allowing for unlimited exports.


aka-bo:
Corp CLA should now be associated with my account

wongma7:
/lgtm
/area nfs

thanks, I will update the changelog and release a new version tomorrow.

PR #403, opened 2017-10-05
[local-volume] Added recommendation for using ID on block vol to README

ianchakeres:
Block volume ID's can be found by looking at `ls -l /dev/disk/by-id`. Using these IDs in directory names can help ensure that the related local PVs are unique and not confused with other local PVs, even if the node name is the same.

### DIFF
--- a/local-volume/README.md
+++ b/local-volume/README.md
@@ -245,6 +245,13 @@ go run hack/e2e.go -- -v --test --test_args="--ginkgo.focus=\[Feature:LocalPersi
   Additionally, this practice will ensure that if another node with the
   same name is created, that any volumes on that node are unique and not
   mistaken for a volume on another node with the same name.
+* For raw block volumes without a filesystem, use a unique ID as the symlink
+  name. Depending on your environment, the volume's ID in `/dev/disk/by-id/`
+  may contain a unique hardware serial number. Otherwise, a unique ID should be 
+  generated. The uniqueness of the symlink name will ensure that if another 
+  node with the same name is created, that any volumes on that node are 
+  unique and not mistaken for a volume on another node with the same name.
+
 
 ### Deleting/removing the underlying volume
 


local-volume/README.md
```diff
@@ -245,6 +245,13 @@ go run hack/e2e.go -- -v --test --test_args="--ginkgo.focus=\[Feature:LocalPersi
   Additionally, this practice will ensure that if another node with the
   same name is created, that any volumes on that node are unique and not
   mistaken for a volume on another node with the same name.
+* For raw block volumes without a filesystem, it's recommended to utilize their 
+  ID (e.g. the output from `ls -l /dev/disk/by-id`)
+  in the directory name of that mount point. Additionally, this practice will
```
msau42: s/in the directory name of that mount point/as the symlink name

mtanino:
@ianchakeres 
Quick question.
Is the name of symlink under /dev/disk/by-id always unique between multiple nodes?
I'm not sure it's ensured.

msau42:
It depends on the environment.  On baremetal, it typically should use the hardware serial number.  But in a VM environment, it may not be unique.

ianchakeres:
Here is one reference that describes the by-id naming: [link](https://wiki.archlinux.org/index.php/persistent_block_device_naming#by-id_and_by-path)

Here is another reference that is a little more specific: [link](https://superuser.com/questions/839667/how-is-by-id-created-on-linux)

"""
for the disk devices in `60-persistent-storage.rules`:

`KERNEL=="sd*|sr*|cciss*", ENV{DEVTYPE}=="disk", ENV{ID_SERIAL}=="?*", SYMLINK+="disk/by-id/$env{ID_BUS}-$env{ID_SERIAL}"`
"""

mtanino:
Agree. It depends on hardware environment, storage type, etc.

>> if another node with the same name is created, that any volumes 
>> +  on that node are unique and not mistaken for a volume on another node with 
>> +  the same name.

So, I thought we can't say this is ensured.

ianchakeres:
@msau42  @mtanino  - What do you think about adding bare-metal to the first sentence?

msau42:
Maybe you can say something like:

For raw block volumes, use a unique ID as the symlink name.  Depending on your environment, the id in /dev/disk/by-id may use the hardware's serial number and will be unique.  Otherwise, a unique ID may need to be generated.

mtanino:
I agree to use by-id naming, this is recommended way to refer physical device.

@msau42 
That's looks good to me.

mtanino:
>> Additionally, this practice will ensure that if another node with the same name is created, that any volumes on that node are unique and not mistaken for a volume on another node with the same name.

>> Additionally, this practice will ensure that if another node with the same name is created, that any volumes on that node are unique and not mistaken for a volume on another node with the same name.

Are there any background that we want to ensure volume-id as unique between multiple node?
May be I'm missing the background.

msau42:
See kubernetes/kubernetes#51431

mtanino:
Thanks. I understand.

mtanino:
lgtm.

PR #413, opened 2017-10-16
Adding support for volume.beta.kubernetes.io/mount-options for nfs-client from PVC or storage class

bonky42:
The purpose of this update is to gather mount-options annotations from storage-class or from pvc (if authorized). If both are available the one on the storage class will prevail.
The origin of this is a problem of compatibility between nfs version, and we didn't want to update each pv after automatic creation.

### DIFF
--- a/nfs-client/cmd/nfs-client-provisioner/provisioner.go
+++ b/nfs-client/cmd/nfs-client-provisioner/provisioner.go
@@ -18,7 +18,6 @@ package main
 
 import (
 	"errors"
-	"encoding/json"
 	"flag"
 	"fmt"
 	"os"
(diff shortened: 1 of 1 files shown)


wongma7:
Storageclass mount options is necessary but in 1.8 there is an official `StorageClass.mountOptions` and it will automatically be propagated to the PV, so we will get it for free without needing code change to nfs-client and we will not have to deal with the complication of reconciling the annotation with the field. Also, allowing PVC to specify mount options is deemed a security risk so we hjave to avoid that as well even if it is convenient :/

thanks.

bonky42:
Good news for 1.8, it will be handy for those of us stuck with nfsv3.
I agree for the security risk, that's why it's deactivated by default (to be honest, it's when I already added it from PVC that it comes to my mind, hey I don't want my users to have access to this!!!)
I understand your point, I will keep this modification in my job, because the vendor which gives us our kubernetes install is in 1.7, and I don't know when they will plane upgrade to 1.8.

thanks

PR #428, opened 2017-11-01
Fix WaitForAvailableCinderVolume() to polling

### LINKED ISSUE #427: WaitForAvailableCinderVolume only sleep 5 seconds instead of polling until volume in available
https://github.com/kubernetes-incubator/external-storage/blob/master/openstack/standalone-cinder/pkg/volumeservice/actions.go#L110-L117

chestack:
This PR fix function WaitForAvailableCinderVolume() to polling check volume status, get volume every 3 seconds and up to 10 times

This PR closes #427

### DIFF
--- a/openstack/standalone-cinder/pkg/volumeservice/actions.go
+++ b/openstack/standalone-cinder/pkg/volumeservice/actions.go
@@ -17,15 +17,19 @@ limitations under the License.
 package volumeservice
 
 import (
+	"strings"
 	"time"
 
 	"github.com/golang/glog"
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack/blockstorage/extensions/volumeactions"
 	volumes_v2 "github.com/gophercloud/gophercloud/openstack/blockstorage/v2/volumes"
+	"k8s.io/apimachinery/pkg/util/wait"
 )
 
 const initiatorName = "iqn.1994-05.com.redhat:a13fc3d1cc22"
+const interval = 3 * time.Second
+const timeout = 60 * time.Second
 
 // VolumeConnectionDetails represent the type-specific values for a given
 // DriverVolumeType.  Depending on the volume type, fields may be absent or
@@ -76,12 +80,10 @@ func CreateCinderVolume(vs *gophercloud.ServiceClient, options volumes_v2.Create
 // become available.  The connection information cannot be retrieved from cinder
 // until the volume is available.
 func WaitForAvailableCinderVolume(vs *gophercloud.ServiceClient, volumeID string) error {
-	// TODO: Implement proper polling instead of brain-dead timers
-	c := make(chan error)
-	go time.AfterFunc(5*time.Second, func() {
-		c <- nil
+	return wait.Poll(interval, timeout, func() (bool, error) {
+		status, err := GetCinderVolumeStatus(vs, volumeID)
+		return strings.ToLower(status) == "available", err
 	})
-	return <-c
 }
 
 // ReserveCinderVolume marks the volume as 'Attaching' in cinder.  This prevents
@@ -143,4 +145,14 @@ func DeleteCinderVolume(vs *gophercloud.ServiceClient, volumeID string) error {
 	}
 
 	return err
+}
+
+// GetCinderVolumeStatus get the given cinder volume status.
+func GetCinderVolumeStatus(vs *gophercloud.ServiceClient, volumeID string) (string, error) {
+	volume, err := volumes_v2.Get(vs, volumeID).Extract()
+	if err != nil {
+		glog.Errorf("Failed to get volume:%v ", volumeID)
+		return "", err
+	}
+	return volume.Status, nil
 }
--- a/openstack/standalone-cinder/pkg/provisioner/vsbroker.go
+++ b/openstack/standalone-cinder/pkg/provisioner/vsbroker.go
@@ -31,6 +31,7 @@ type volumeServiceBroker interface {
 	disconnectCinderVolume(vs *gophercloud.ServiceClient, volumeID string) error
 	unreserveCinderVolume(vs *gophercloud.ServiceClient, volumeID string) error
 	deleteCinderVolume(vs *gophercloud.ServiceClient, volumeID string) error
+	getCinderVolumeStatus(vs *gophercloud.ServiceClient, volumeID string) (string, error)
 }
 
 type gophercloudBroker struct {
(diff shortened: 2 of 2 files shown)


openstack/standalone-cinder/pkg/volumeservice/actions.go
```diff
@@ -108,11 +109,26 @@ func CreateCinderVolume(vs *gophercloud.ServiceClient, options controller.Volume
...
 	c := make(chan error)
-	go time.AfterFunc(5*time.Second, func() {
-		c <- nil
-	})
+	go func() error {
+		num :=0
+		for {
+			time.Sleep(3 * time.Second)
+			if num > 10 {
+				err := errors.New("Failed to get volume status after max retries")
```
aglitke: I would prefer at least a pair of constants to define the sleep interval and max retries

openstack/standalone-cinder/pkg/volumeservice/actions.go
```diff
@@ -108,11 +109,26 @@ func CreateCinderVolume(vs *gophercloud.ServiceClient, options controller.Volume
...
+				c <- err
+			}
+			volume, err := volumes_v2.Get(vs, volumeID).Extract()
+			if err != nil {
+				glog.Errorf("Failed to get volume:%v ", volumeID)
```
aglitke: This should either skip to the next iteration or bail out as we cannot continue to use the volume object in the next if statement if err != nil.

openstack/standalone-cinder/pkg/volumeservice/actions.go
chestack: absolutely, I will fix it in next patch

openstack/standalone-cinder/pkg/volumeservice/actions.go
chestack: agree, change to skip to the next iteration

chestack:
@aglitke , thanks for your comments. I had fixed it. please help review

openstack/standalone-cinder/pkg/volumeservice/actions.go
```diff
@@ -108,11 +111,27 @@ func CreateCinderVolume(vs *gophercloud.ServiceClient, options controller.Volume
...
 	c := make(chan error)
-	go time.AfterFunc(5*time.Second, func() {
-		c <- nil
-	})
+	go func() error {
+		num := 0
+		for {
+			time.Sleep(sleepTime * time.Second)
+			if num > maxRetry {
+				err := errors.New("Failed to get volume status after max retries")
```
aglitke: This error would be better written as: "Volume not available after max retries"

openstack/standalone-cinder/pkg/volumeservice/actions.go
```diff
@@ -108,11 +111,27 @@ func CreateCinderVolume(vs *gophercloud.ServiceClient, options controller.Volume
...
+				c <- err
+			}
+			volume, err := volumes_v2.Get(vs, volumeID).Extract()
```
aglitke: You might consider adding a vsbroker interface for volumes_v2.Get.   See #453.  We want to add unit tests for this method to ensure that it behaves as expected in the good case, when max retries are exceeded, and when we fail to get the volume information whatsoever.

openstack/standalone-cinder/pkg/volumeservice/actions.go
chestack: Done in next change

openstack/standalone-cinder/pkg/volumeservice/actions.go
chestack: Agree that It's better to add interface/function "getCinderVolumeState" into vsbroker/actions,  and reuse it in function "waitForAvailableCinderVolume"(could also be reused in other cases).  But it will be overlap with what did in #453

chestack:
@aglitke , we do not  have UT both for vsbroker.go and actions.go. So should I add UT in this PR or another one?

aglitke:
@chestack No need to test the vsbroker or actions directly.  What I mean is that we should test the provision flow in a few situations where polling for the volume may succeed or fail.  You can test these cases by overriding a new GetVolume interface just like we do for some of the other provisioner tests.

openstack/standalone-cinder/pkg/volumeservice/actions.go
```diff
@@ -108,11 +111,26 @@ func CreateCinderVolume(vs *gophercloud.ServiceClient, options controller.Volume
 // become available.  The connection information cannot be retrieved from cinder
 // until the volume is available.
 func WaitForAvailableCinderVolume(vs *gophercloud.ServiceClient, volumeID string) error {
-	// TODO: Implement proper polling instead of brain-dead timers
 	c := make(chan error)
-	go time.AfterFunc(5*time.Second, func() {
-		c <- nil
-	})
+	go func() error {
```
wenjun93: could we use util function `k8s.io/apimachinery/pkg/util/wait.poll()` instead? @chestack @aglitke

aglitke: @wenjun93 +1

hzxuzhonghu: +1

chestack: thanks for all your comments. Done in next patch. sorry for fixing it late.

openstack/standalone-cinder/pkg/volumeservice/actions.go
```diff
@@ -108,12 +111,10 @@ func CreateCinderVolume(vs *gophercloud.ServiceClient, options controller.Volume
 // become available.  The connection information cannot be retrieved from cinder
 // until the volume is available.
 func WaitForAvailableCinderVolume(vs *gophercloud.ServiceClient, volumeID string) error {
-	// TODO: Implement proper polling instead of brain-dead timers
-	c := make(chan error)
-	go time.AfterFunc(5*time.Second, func() {
-		c <- nil
+	return wait.Poll(interval, timeout, func() (bool, error) {
+		status, err := GetCinderVolumeStatus(vs, volumeID)
+		return status == "available", err
```
rootfs: i am paranoid but making strings.ToLower() to these strings is more robust

chestack: that make sense. Fix it in next change

rootfs:
/lgtm

cc @childsb for merge

rootfs:
@chestack please rebase and ping us when ready for merge

chestack:
@rootfs @childsb , has rebased, please review.

PR #429, opened 2017-11-01
Add openstack/manila as an external storage provisioner

chestack:
This PR is still in WIP status, pushed here to hope to get feedbacks 

update gophercloud to new version which has enabled sharedfilesystems
set needed parameters: share-networkid, share-protocol or anything else into cloud config file of cluster


openstack/manila/deployment.yaml
```diff
@@ -0,0 +1,28 @@
...
+      containers:
+      - name: manila-provisioner
+        image: "quay.io/external_storage/manila-provisioner:latest"
+        imagePullPolicy: IfNotPresent
+        env:
+        - name: OS_AUTH_URL
+          value: http://192.168.2.16:5000/v2.0
+        - name: OS_USERNAME
+          value: admin
+        - name: OS_PASSWORD
```
strigazi: Please consider authenticating using a trustid, it is supported by gophercloud.

chestack: @strigazi , glad to see you here, from OpenStack magnum to Kubernetes!

For the problem, absolutely I should use trustid. learn from magnum, it should be able to mount cloud config file which to manila provisioner. what it looks like now is what copied from standalone-cinder. I will fix it in next patch

chestack:
I think it's better to split this change into two:
1. upgrade gophercloud to new version which has enabled sharedfilesystems and to resolve compatible problems.
2. add openstack/manila as an external storage provisioner.

@aglitke @strigazi , what's your opinions? thanks.

childsb:
There's another effort to add an external manilla provisioner here: https://github.com/kubernetes-incubator/external-storage/pull/524

Can we get together and merge both efforts?  It's too bad both efforts were going on in parallel, we could have saved a lot of time by doing this as one..

@chestack have you had a chance to look at PR 524 in comparison with this?  This one is MUCH smaller, but I believe @pospispa updated the gopher cloud library in his PR to support different storage backends..

wongma7:
seems the openstack community has converged on maintaining one here https://github.com/kubernetes/cloud-provider-openstack , let's point to that.

PR #453, opened 2017-11-10
standalone-cinder: provision with existing cinderVolumeID

wenjun93:


openstack/standalone-cinder/pkg/provisioner/provisioner.go
```diff
@@ -36,7 +36,10 @@ const (
 	ProvisionerIDAnn = "standaloneCinderProvisionerIdentity"
 
 	// CinderVolumeID is an annotation to store the ID of the associated cinder volume
-	CinderVolumeID = "cinderVolumeId"
+	CinderVolumeID = "cinderVolumeID"
+
+	// ShouldProvisionVolumeAnn is an annotation to idetify whether to create new cinder volume or using existing volume
+	ShouldProvisionVolumeAnn = "pv.kubernetes.io/cinderVolumeID"
```
aglitke: How would you support provisioning via clone in the future?  Maybe add another annotation: "pv.kubernetes.io/clone" ?

openstack/standalone-cinder/pkg/provisioner/provisioner.go
```diff
@@ -74,19 +77,37 @@ func NewCinderProvisioner(client kubernetes.Interface, id, configFilePath string
...
+		if err != nil {
+			glog.Errorf("Failed to get volume %s, err: %v", volumeID, err)
+			goto ERROR
+		}
+		if volumeState != "available" {
+			glog.Errorf("Volume %s state is not available, current state %v", volumeID, volumeState)
+			err = fmt.Errorf("Volume %s state is not available, current state %v", volumeID, volumeState)
+			goto ERROR
+		}
+		goto CONNECT
```
aglitke: We should record the state that this is a pre-existing volume so if we encounter an error later (eg. connecting or reserving) we won't delete the volume.

openstack/standalone-cinder/pkg/provisioner/vsbroker.go
```diff
@@ -31,6 +31,7 @@ type volumeServiceBroker interface {
 	disconnectCinderVolume(vs *gophercloud.ServiceClient, volumeID string) error
 	unreserveCinderVolume(vs *gophercloud.ServiceClient, volumeID string) error
 	deleteCinderVolume(vs *gophercloud.ServiceClient, volumeID string) error
+	getCinderVolumeState(vs *gophercloud.ServiceClient, volumeID string) (string, error)
```
aglitke: This reminds me that we should have some tests for this functionality.  At least the positive flow, volume not found, and volume in wrong state.  We can also check that a found volume is not deleted in the event of a later error.

openstack/standalone-cinder/pkg/provisioner/provisioner.go
wenjun93: maybe I can rename this ShouldProvisionVolumeAnn annotation, could we have an annotation list for the future and arrange their name together.

openstack/standalone-cinder/pkg/provisioner/provisioner.go
wenjun93: yeah, I'm also thinking of this, I'll add it

openstack/standalone-cinder/pkg/provisioner/vsbroker.go
wenjun93: Do you mean unit test for this function?

wenjun93:
@aglitke I found when creating a number of PVCs, there will be multiple `provision()` operations, and sometimes error occurs in `reserveCinderVolume` step and `deleteCinderVolume` step in `provision` function, caused a lot of `available` cinder volumes not being used and cleaned, I have to clean the volumes manually, so I added retry here, how do you think?

aglitke:
@wenjun93 The reserve errors are usually due to the volume not yet becoming available.  Shouldn't this be solved by proper polling for the volume available status?

wenjun93:
@aglitke Right, I add wait.poll() for the volume available status and it works,  it' better than add wait poll in each volume operation, thanks.

wenjun93:
@aglitke would you mind review the changes? Thanks.

openstack/standalone-cinder/pkg/provisioner/provisioner.go
```diff
@@ -142,6 +164,9 @@ ERROR_UNRESERVE:
 	}
 	glog.V(3).Infof("Volume %s unreserved", volumeID)
 ERROR_DELETE:
+	if found {
+		goto ERROR
+	}
```
aglitke: I would prefer this to read:

if !found {
  // Delete volume
}

Instead of adding a goto.  The gotos are readable only if there is a single stack of operations that are unwound by the error cleanup labels at the end.  As soon as you add additional jumps it gets harder to follow.

openstack/standalone-cinder/pkg/provisioner/vsbroker.go
wenjun93: @aglitke Thanks, I've addressed other comments. For this one, I think might need to add another provisioner and deleting test case using different volume options (useExistingVolume=true), see [here](https://github.com/kubernetes-incubator/external-storage/blob/master/openstack/standalone-cinder/pkg/provisioner/provisioner_test.go#L33), right?

openstack/standalone-cinder/pkg/provisioner/provisioner.go
rootfs: follow k8s name convention [here](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/), `pv.kubernetes.io/cinder-volume-id`

openstack/standalone-cinder/pkg/provisioner/provisioner.go
```diff
@@ -74,19 +77,37 @@ func NewCinderProvisioner(client kubernetes.Interface, id, configFilePath string
...
 		return nil, fmt.Errorf("claim Selector is not supported")
 	}
 
+	if volumeID, useExistingVolume = options.PVC.Annotations[ShouldProvisionVolumeAnn]; useExistingVolume {
+		volumeState, err = p.vsb.getCinderVolumeState(p.VolumeService, volumeID)
+		if err != nil {
+			glog.Errorf("Failed to get volume %s, err: %v", volumeID, err)
+			goto ERROR
+		}
+		if volumeState != "available" {
```
rootfs: strings.ToLower

openstack/standalone-cinder/pkg/provisioner/provisioner.go
```diff
@@ -74,19 +77,37 @@ func NewCinderProvisioner(client kubernetes.Interface, id, configFilePath string
...
+			err = fmt.Errorf("Volume %s state is not available, current state %v", volumeID, volumeState)
+			goto ERROR
+		}
+		goto CONNECT
+	}
+
 	// TODO: Check access mode
 
-	volumeID, err := p.vsb.createCinderVolume(p.VolumeService, options)
+	volumeID, err = p.vsb.createCinderVolume(p.VolumeService, options)
```
rootfs: refactor `createCinderVolume` to pass volumeID to `source_volid`  in CreateOpts.

openstack/standalone-cinder/pkg/provisioner/provisioner.go
```diff
@@ -74,19 +77,37 @@ func NewCinderProvisioner(client kubernetes.Interface, id, configFilePath string
...
+		cleanupErr        error
+		useExistingVolume bool
+		volumeState       string
 	)
 
 	if options.PVC.Spec.Selector != nil {
 		return nil, fmt.Errorf("claim Selector is not supported")
 	}
 
+	if volumeID, useExistingVolume = options.PVC.Annotations[ShouldProvisionVolumeAnn]; useExistingVolume {
```
rootfs: for better readability, add a var called sourceVolumeID and use it here. Use volumeID for new volume.

openstack/standalone-cinder/pkg/provisioner/provisioner.go
rootfs: i am also interested to see how cinder create volume works if you pass both source volume id and capacity. Cinder API [spec](https://developer.openstack.org/api-ref/block-storage/v3/index.html#volumes-volumes) says `The API creates a new volume with the same size as the source volume.`, but doesn't specify the behavior if both capacity and source volume id are set.

wenjun93: @rootfs I've tried with cinder cli, must set size in createOpts, if both size and source volume id are set, cinder will still set according to size.

wenjun93: This is not to clone an existing volume, but use the existing volume and create pv with the volume information. No actual volume provisioning here.

stmcginnis: The only caveat to that is the newly requested size must be equal to or greater than the source volume size.

So if just source volume is provided, Cinder will create the new volume to match the source size. If both are specified and the requested size is at least the same size as the source volume, that new volume will be created from the source volume but sized as requested.

wongma7:
/area openstack/standalone-cinder
@aglitke are we good to go on merging this?

PR #458, opened 2017-11-14
update hostpath-provisioner.

sak0:
update hostpath-provisioner.

Signed-off-by: CuiHaozhi <cuihaozhi@chinacloud.com.cn>


wongma7:
thanks. it needs an overhaul (docs included) but let's merge it like this since a few people have asked.

PR #460, opened 2017-11-15
Improve script verify-gofmt.sh

chestack:
It need flag "-w" to format bad files. So it's better to add this message into script.


wongma7:
/lgtm

thank you

PR #461, opened 2017-11-15
Don't try to update ceph.conf if file is read-only

JrCs:
Don't try to update the ceph.conf file if it is a read only file (like a projection of a k8s configmap).


rootfs:
lgtm but wait for CLA

JrCs:
CLA signed since yesterday

Issue #455, opened 2017-11-11
cephfs provisioner leaks Secret objects


farcaller:
The cephfs provisioner doesn't clean up Secret objects (doesn't re-use them either), and doesn't clean up the actual users from ceph, leaving them dangling with some permissions:

```
client.kubernetes-dynamic-user-8e229ae4-c4a5-11e7-ad64-0adc8221b842
	key: 
	caps: [mds] allow r
	caps: [mon] allow r
	caps: [osd]
```

Instead the provisioner should return the storage to the same state it was before the PVC was created.

wongma7:
/assign @rootfs 
kind of related, we should no longer be creating secrets in same namespace as pvc anyhow: https://github.com/kubernetes-incubator/external-storage/issues/309

rootfs:
the provisioner [deauthorize](https://github.com/kubernetes-incubator/external-storage/blob/master/ceph/cephfs/cephfs_provisioner/cephfs_provisioner.py#L229) the user when deleting the share. 

Maybe the cap is not all that is expected in the library [here](https://github.com/ceph/ceph/blob/master/src/pybind/ceph_volume_client.py#L1193), since mds is given `allow r` [here](https://github.com/kubernetes-incubator/external-storage/blob/master/ceph/cephfs/cephfs_provisioner/cephfs_provisioner.py#L120) to work around a kernel cephfs issue.

rootfs:
@cofyc do you have a chance to look into this? thanks

cofyc:
@rootfs
OK, I will take a look at it.

JrCs:
My PR #466 only remove ceph user if it's name start with `kubernetes-dynamic-user-`

Issue #309, opened 2017-08-22
Ceph provisioner requires creating secrets in PVC namespaces


rootfs:
Existing CephFS PV expects a secret exists in the Pod's (or Claim's) namespace, and thus tries to create secret in *any* namespace. This will bring risk if CephFS provisioner is compromised and overrides secrets used by kubernetes.

@liggitt

PR #466, opened 2017-11-16
Remove dynamic users when pv is remove

JrCs:
This PR will remove dynamic users create by the provisionner (k8s secret and ceph client auth).

Fixes: #455

### DIFF
--- a/ceph/cephfs/cephfs-provisioner.go
+++ b/ceph/cephfs/cephfs-provisioner.go
@@ -68,6 +68,10 @@ func newCephFSProvisioner(client kubernetes.Interface, id string) controller.Pro
 
 var _ controller.Provisioner = &cephFSProvisioner{}
 
+func generateSecretName(user string) string {
+	return "ceph-" + user + "-secret"
+}
+
 // Provision creates a storage asset and returns a PV object representing it.
 func (p *cephFSProvisioner) Provision(options controller.VolumeOptions) (*v1.PersistentVolume, error) {
 	if options.PVC.Spec.Selector != nil {
@@ -104,7 +108,7 @@ func (p *cephFSProvisioner) Provision(options controller.VolumeOptions) (*v1.Per
 	}
 	// create secret in PVC's namespace
 	nameSpace := options.PVC.Namespace
-	secretName := "ceph-" + user + "-secret"
+	secretName := generateSecretName(user)
 	secret := &v1.Secret{
 		ObjectMeta: metav1.ObjectMeta{
 			Namespace: nameSpace,
@@ -143,6 +147,7 @@ func (p *cephFSProvisioner) Provision(options controller.VolumeOptions) (*v1.Per
 					SecretRef: &v1.SecretReference{
 						Name: secretName,
 						// TODO https://github.com/kubernetes-incubator/external-storage/issues/309
+						Namespace: nameSpace,
 					},
 					User: user,
 				},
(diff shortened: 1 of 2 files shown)


JrCs:
Fixes: #455

ceph/cephfs/cephfs_provisioner/cephfs_provisioner.py
```diff
@@ -196,6 +196,15 @@ def cap_update(orig, want, unwanted):
         assert caps[0]['entity'] == client_entity
         return caps[0]
 
+    def _remove_ceph_user(self, user_id):
+        client_entity = "client.{0}".format(user_id)
+        ret = self._volume_client._rados_command(
+            'auth rm',
```
rootfs: interesting, `auth rm` works but according to [doc](http://docs.ceph.com/docs/jewel/rados/operations/user-management/), should be `auth del`

ceph/cephfs/cephfs_provisioner/cephfs_provisioner.py
```diff
@@ -196,6 +196,15 @@ def cap_update(orig, want, unwanted):
         assert caps[0]['entity'] == client_entity
         return caps[0]
 
+    def _remove_ceph_user(self, user_id):
+        client_entity = "client.{0}".format(user_id)
+        ret = self._volume_client._rados_command(
```
rootfs: add try/catch in case user already removed.

ceph/cephfs/cephfs-provisioner.go
```diff
@@ -196,6 +200,14 @@ func (p *cephFSProvisioner) Delete(volume *v1.PersistentVolume) error {
 		return cmdErr
 	}
 
+	// Remove dynamic user secret
+	secretName := generateSecretName(user)
+	err = p.client.Core().Secrets(volume.Spec.ClaimRef.Namespace).Delete(secretName, &metav1.DeleteOptions{})
```
rootfs: use SecretReference.Namespace
see https://github.com/kubernetes-incubator/external-storage/blob/master/vendor/k8s.io/api/core/v1/types.go#L921

rootfs:
thanks, can you look into https://github.com/kubernetes-incubator/external-storage/issues/309 as well?

cofyc:
Hi,

`cephfs_provisioner.py` implements its customized `_authorize_ceph` method instead of using `ceph_volume_client.py`'s [_authorize_ceph](https://github.com/ceph/ceph/blob/master/src/pybind/ceph_volume_client.py#L1017) method.

How about following the same logic to implement a customized `_deauthorize` method instead of using `ceph_volume_client.py`'s [_deauthorize](https://github.com/ceph/ceph/blob/master/src/pybind/ceph_volume_client.py#L1181) ?

In this customized `_deauthorize` we can remove caps first,  and if all caps are removed then remove the given user like `ceph_volume_client.py`'s [_deauthorize]((https://github.com/ceph/ceph/blob/master/src/pybind/ceph_volume_client.py#L1181) ) do.

Doing it in this way, we can avoid hard-coding in `cephfs_provisioner.py` script. And it's possible to share a ceph user for many pvs in future.

JrCs:
Modification done from @rootfs reviews.
I'm agree with @cofyc: it will be better to have a custom `_deauthorize` method. My python skill is limited so i prefer to let you code this method.

rootfs:
@JrCs how about just merge your golang part (i.e. secret) and wait for python fix from @cofyc

JrCs:
@rootfs we can merge this PR completly (it fixes #455) and create another PR to have an enhance method to deauthorize the user ?

rootfs:
ping @wongma7, can you merge? thanks

PR #467, opened 2017-11-17
Add snapshot functionality for GlusterFS.

humblec:

### DIFF
--- a/snapshot/pkg/apis/crd/v1/types.go
+++ b/snapshot/pkg/apis/crd/v1/types.go
@@ -204,6 +204,12 @@ type HostPathVolumeSnapshotSource struct {
 	Path string `json:"snapshot"`
 }
 
+// GlusterVolumeSnapshotSource is Gluster volume snapshot source
+type GlusterVolumeSnapshotSource struct {
+	// UniqueID represents a snapshot resource.
+	SnapshotID string `json:"snapshotId"`
+}
+
 // AWSElasticBlockStoreVolumeSnapshotSource is AWS EBS volume snapshot source
 type AWSElasticBlockStoreVolumeSnapshotSource struct {
 	// Unique id of the persistent disk snapshot resource. Used to identify the disk snapshot in AWS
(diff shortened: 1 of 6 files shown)


rootfs:
thanks
cc @tsmetana

snapshot/pkg/apis/crd/v1/types.go
```diff
@@ -235,6 +241,9 @@ type VolumeSnapshotDataSource struct {
 	// kubelet's host machine and then exposed to the pod.
 	// More info: https://kubernetes.io/docs/concepts/storage/volumes#awselasticblockstore
 	// +optional
+	//GlusterSnapshotSource represents a gluster snapshot resource
+	GlusterSnapshotVolume *GlusterVolumeSnapshotSource `json:"glusterSnapshotVolume,omitempty"`
```
tsmetana: It's not discoverable by the controller: you should also update the GetSupportedVolumeFromPVSpec and GetSupportedVolumeFromSnapshotDataSpec functions.

humblec: @tsmetana Thanks, Done. :)

snapshot/pkg/volume/gluster/snapshot.go
```diff
@@ -0,0 +1,217 @@
...
+
+func (h *glusterfsPlugin) SnapshotCreate(pv *v1.PersistentVolume, tags *map[string]string) (*crdv1.VolumeSnapshotDataSource, *[]crdv1.VolumeSnapshotCondition, error) {
+	spec := &pv.Spec
+	if spec == nil || spec.Glusterfs == nil {
+		return nil, nil, fmt.Errorf("invalid PV spec %v", spec)
+	}
+
+	volumePath := spec.Glusterfs.Path
+	snapshotName := volumePath + string(uuid.NewUUID())
+	cmd := exec.Command("sudo", glusterfsBinary, "snapshot", "create", snapshotName, volumePath)
```
tsmetana: I don't think "sudo" is a right thing to do... The controller must run under an user who can use the glusterfs utilities and sudo should not be required. The controller should also work when run in container.

humblec: @tsmetana sure, addressed the comment.

tsmetana:
Looking good. Could you also please update the documentation and examples?

humblec:
@tsmetana Done. The documentation is updated.  More improvements can be done by follow up PRs.

tsmetana:
Thank you!

humblec:
@wongma7 PTAL.

PR #469, opened 2017-11-17
move sc util from k8s pkg/api/v1/helper out

rootfs:
let's see what it breaks.

k8s master has moved pkg/api/v1/helper out. This patch allow CSI external-provisioner to pickup the latest k8s code before vendored in.


rootfs:
@jsafrane

jsafrane:
The methods were removed because storage class annotation has been deprecated. IMO, we should follow the suit.

PR #477, opened 2017-11-22
Update dependencies

tsmetana:
When working on the snapshot code I (and others...) started recently seeing this error on controller startup:

panic: unable to load in-cluster configuration, KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT must be defined
goroutine 1 [running]:
main.main()
/home/hchiramm/gopath/src/github.com/kubernetes-incubator/external-storage/snapshot/cmd/snapshot-controller/snapshot-controller.go:63 +0x360

This is caused by changes outside of the snapshot code so I'm afraid the only thing that helps is update of the dependencies to new versions (glide up -v && glide-vc --use-lock-file).


tsmetana:
The failure occurs less frequently but still... So the dependencies update is premature. I need to find out what changed and where.

Issue #479, opened 2017-11-23
[local-volume] Enhance LV provisioner to allow use of kubeconfig file for API access


dhirajh:
Currently, the PV provisioner only allows in-cluster config mechanism to access API server. This is a request to enhance the provisioner and give it the option to use a kubeconfig file based access to the API.

msau42:
Can this be closed?

dhirajh:
Yes. It has been fixed with the pr #480

PR #480, opened 2017-11-23
Enhance LV provisioner to have option to access API using kubeconfig 

dhirajh:
This addresses issue #479 
Apart from the default in-config mechanism of accessing the API server, this change provides the option of using  kubeconfig file to access the API.

### DIFF
--- a/local-volume/provisioner/cmd/main.go
+++ b/local-volume/provisioner/cmd/main.go
@@ -27,7 +27,6 @@ import (
 	"k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
-	"k8s.io/client-go/rest"
 )
 
 var provisionerConfig common.ProvisionerConfiguration
@@ -42,19 +41,6 @@ func init() {
 	glog.Infof("Configuration parsing has been completed, ready to run...")
 }
 
-func setupClient() *kubernetes.Clientset {
-	config, err := rest.InClusterConfig()
-	if err != nil {
-		glog.Fatalf("Error creating InCluster config: %v\n", err)
-	}
-
-	clientset, err := kubernetes.NewForConfig(config)
-	if err != nil {
-		glog.Fatalf("Error creating clientset: %v\n", err)
-	}
-	return clientset
-}
-
 func main() {
 	flag.Set("logtostderr", "true")
 	flag.Parse()
@@ -64,7 +50,7 @@ func main() {
 		glog.Fatalf("MY_NODE_NAME environment variable not set\n")
 	}
 
-	client := setupClient()
+	client := common.SetupClient()
 	node := getNode(client, nodeName)
 
 	glog.Info("Starting controller\n")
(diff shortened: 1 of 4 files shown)


dhirajh:
cc @msau42

local-volume/bootstrapper/deployment/kubernetes/bootstrapper.yaml
```diff
@@ -18,6 +18,27 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.namespace
+#       Specify volume mount for kubeconfig file (if booststrap container is using for API calls,
+#       instead of the default in-cluster config).
+#        volumeMounts:
+#        - mountPath: /path/to/kubeconfig
+#          name: kubeconfig
+#          readOnly: true
+#        - mountPath: /path/to/pki
```
msau42: When I look at the Kubernetes setup guide [here](https://kubernetes.io/docs/getting-started-guides/scratch/#preparing-credentials), it seems like most components only require the kubeconfig file, and the keys are actually included in there too.  Do we still need the pki folder?

dhirajh: @msau42 The scenario that I was trying to support is one where the keys are not embedded in the config file  and are rather referred to by file name (as seen [here](https://kubernetes.io/docs/tasks/access-application-cluster/configure-access-multiple-clusters/#define-clusters-users-and-contexts)). This approach is used in environments where keys are rotated frequently for security reasons and they are typically rotated by automated processes that only deal with keys/certs files  for large number of different applications in a typical datacenter (not just k8s) and have no specific knowledge of kubeconfig and its embedded format.

A secondary benefit for having this folder option is for a situation I ran into while working on block device cleanup. I wanted to have the option of  loading host utilities into the container, which could then be used for cleanups, instead of making the user re-build the container image (discussed [here](https://docs.google.com/document/d/1hRRzZpWtbHyJC1lotE8rqp2h1NEXuFv-rAW-vLuQa94/edit?disco=AAAAA_weNrc)). This folder option would enable that feature.

Let me know what you think. The other option here is that for such (advanced) needs, the user doesn't use the Bootstrap process and instead works directly with crafting the daemonset manifest as the bootstrap process is more a convenience than a requirement for running the provisioner.

dhirajh: I favor keeping keeping the bootstrap code simple and focussed on common use cases for quick setup (the latter option). Working directly with LV provisioner daemonset manifest seems reasonable for more advanced use cases and its actually fairly intuitive for that purpose.

msau42: Yeah, I think it's kind of awkward that you have to specify the kubeconfig + pki in the boostrapper yaml and the configmap.

I wonder if we should reconsider the bootstrapper and just make it a script that generates all the necessary provisioner yamls from some templates.

dhirajh: OK, In that case, I will limit the changes to just LV provisioner and leave the bootstrapper code unchanged.

+1 on using bootstrapper only as a code generator. The current Job based approach sort of hides the daemonset manifest from the user and that is not really a good thing.  A template approach is more transparent.

msau42: Opened up #487 to change the boostrapper

dhirajh:
@msau42 Updated PR to remove changes to bootstrapper and limit them to provisioner. PTAL

PR #481, opened 2017-11-23
Add an option to specify supported image features to allow

JrCs:


rootfs:
I haven't tested other image features yet, but opening up supported features to cli options may open door to allow unsupported features slip in, and we have to validate the the cli passed supported features.
/assign @cofyc

cofyc:
hi, @JrCs 

Not all image format 2 features can be turned on without causing problems.

For example, we cannot turn on exclusive-lock feature for now (and object-map, fast-diff, journaling which require exclusive-lock), because exclusive lock and advisory lock cannot work together. (See https://github.com/kubernetes/examples/blob/master/staging/persistent-volume-provisioning/README.md#ceph-rbd and https://github.com/kubernetes/kubernetes/pull/45805).

Kubernetes pre-1.9, we use advisory lock to lock images on the node.

If we are going to support more image format 2 features, we need to take care these issues (e.g. allow some features when we know it's safe to use). Otherwise, user may run into problems.

JrCs:
Ok, thanks @cofyc for the informations.
Why can we not turn on exclusive-lock now ? For me it's include in kernel 4.12.x and above ?

rootfs:
@JrCs we don't need exclusive lock - it doesn't live well with advisory lock, and rbd can use attach/detach controller to fend off multi attach now.

wongma7:
resolved, thanks everyone

Issue #482, opened 2017-11-24
[local-volume] How do you add a mount point atomically?


bryanlarsen:
If I do `sudo mkdir /mnt/disks/foo && sudo mount /mnt/disks/foo`, local-volume-provisioner is too quick, it finds the volume before it's mounted and sets its size to the free space of /mnt/disks rather than the free space of /mnt/disks/foo.    I have to prepend it with `kubectl delete pod local-volume-provisioner-4rk7v &&`.

I got it working, but it seems like an awkward workflow, so hopefully there's a better way, and it should be probably be documented because I imagine it would be an FAQ.

jsafrane:
/close

Fixed by #499

Issue #370, opened 2017-09-15
local storage provisioner reads wrong capacity


jsafrane:
Let's pretend we have mount propagation in place and provisioner can see mounts that were mounted while the provisioner was running.

There is a race between admin mounting a volume and provisioner creating a PV for it:

1. admin creates /mnt/disks/new-volume1
2. admin mounts /dev/sdc to /mnt/disks/new-volume1

Provisioner will create a PV with wrong capacity when it scans /mnt/disks after step 1 and before 2. Should the provisioner check that new-volume1 is a mount point before creating the PV? It won't be possible to create a PV for the root partition then. Should admin bind-mount the root partition when creating a PV for the root partition is really wanted?

@wongma7 @ddysher @msau42 @ianchakeres, WDYT?

ianchakeres:
Here are three possible solutions:

#1 - Admin use permissions

Have the provisioner ignore directories with specific restricted permissions and have the administrator set these permissions on any directory (/mnt/disks/new-volume1) that should be ignored by the provisioner.

When the mount operation is performed, the now mounted location's permissions (/dev/sdc) will take over the directory permissions.

If we wanted to further restrict the directories permissions that are considered ignored, then we could also use the CAP_LINUX_IMMUTABLE on the mount point directory (/mnt/disks/new-volume1). This parameter could be checked by the provisioner, and the provisioner could avoid creating local PV for any directory with this flag set. 

#2 - Admin uses lockfile 

We could have a lock file that the administrator can set/clear to avoid the provisioner from scanning a directory for potential PV or a specific lock file to exclude a particular directory (e.g. new-volume1.lock).

#3 - Admin stops provisioner

The administrator could be expected to stop the provisioner during these types of operations.

wongma7:
IMO yes and yes, it would be good to mandate that every directory in /mnt/disks is a mountpoint. That requirement is clear, it doesn't require the admin to understand any new mechanism, and forcing the admin to do a bind mount to explicitly express that "this dir is ready to have a pv created for it" is less work than any of the other solutions.

msau42:
Unfortunately as part of our user requirements, we also need to support directories, and not just mount points.  Another possibility is to only allow specific types of volumes per directory and add it to the configuration:

storageClass = "fast", hostDir = "/mnt/ssds", type = "mounts"
storageClass = "shared", hostDir = "/something/in/root", type = "directories"
[...]

PR #486, opened 2017-11-27
[local-volume] Updated README with known discovery issues and workaround.

ianchakeres:
Multiple discovery issues exists while using the local-volume provisioner.

Some situations are described here: https://github.com/kubernetes-incubator/external-storage/issues/370; and another related issue has been filed: https://github.com/kubernetes-incubator/external-storage/issues/482.

This PR describes the known issue and a potential workaround.

### DIFF
--- a/local-volume/README.md
+++ b/local-volume/README.md
@@ -37,11 +37,15 @@ What doesn't work and workarounds:
         * Give your pods high priority.
         * Run a workaround controller that unbinds PVCs for pods that are
           stuck pending. TODO: add link
-* External provisioner cannot correctly detect capacity of mounts added after it
-  has been started.
-    * This requires mount propagation to work, which is targeted for 1.9.
-    * Workaround: Before adding any new mount points, stop the daemonset, add
-      the new mount points, start the daemonset.
+* The provisioner will not correctly detect mounts added after it
+  has been started due to lack of mount propogation.
+  * The provisioner will detect the existance of a new directory in the
+  discovery directory. Then, it will incorrectly create a local PV with
+  the root filesystem capacity. 
+  * Planned resolution of new mount point discovery requires mount propagation,
+  which is targeted for 1.9.
+  * Workaround: Before adding any new mount points, stop the provisioner
+  daemonset, add the new mount points, start the daemonset.
 
 Future features:
 * Local block devices as a volume source, with partitioning and fs formatting


local-volume/provisioner/README.md
```diff
@@ -48,3 +48,18 @@ The basic components of the provisioner are as follows:
 - Controller: The controller runs a sync loop that coordinates the other components.
   The discovery and deleter run serially to simplify synchronization with the cache
   and create/delete operations.
+
+## Known Issues and Workarounds
```
msau42: Can you actually add this to the top level README, which has all the known issues?

ianchakeres: Sure. Done.

local-volume/README.md
```diff
@@ -37,11 +37,19 @@ What doesn't work and workarounds:
...
         * Run a workaround controller that unbinds PVCs for pods that are
           stuck pending. TODO: add link
-* External provisioner cannot correctly detect capacity of mounts added after it
-  has been started.
-    * This requires mount propagation to work, which is targeted for 1.9.
-    * Workaround: Before adding any new mount points, stop the daemonset, add
-      the new mount points, start the daemonset.
+* The provisioner does not correctly detect mounts added after it
+  has been started, and may fail or incorrectly creates local PVs with
+  zero capacity. This situation manifests as a race condition between discovery
```
msau42: zero capacity -> the root filesystem capacity

msau42: It's actually not a race condition right now because there is no mount propagation enabled.  It always happens.

local-volume/README.md
```diff
@@ -17,7 +17,7 @@ directories by creating and cleaning up PersistentVolumes for each volume.
 
 ## Feature Status
 
-Current status: 1.7 & 1.8 - Alpha
+Current status: 1.7, 1.8, & 1.9 - Alpha
```
msau42: Can you leave out 1.9 for now.  I'm going to update this soon with 1.9 status; a lot has changed.

local-volume/README.md
ianchakeres: Ok. Updated.

local-volume/README.md
```diff
@@ -37,11 +37,19 @@ What doesn't work and workarounds:
...
           stuck pending. TODO: add link
-* External provisioner cannot correctly detect capacity of mounts added after it
-  has been started.
-    * This requires mount propagation to work, which is targeted for 1.9.
-    * Workaround: Before adding any new mount points, stop the daemonset, add
-      the new mount points, start the daemonset.
+* The provisioner does not correctly detect mounts added after it
+  has been started, and it incorrectly creates local PVs with
+  the root filesystem capacity. This situation consistently manifests due to
+  the non-atomic nature of discovery of the directory for the new mountpoint
```
msau42: It's because mount propagation is not enabled, so the provisioner cannot see new mount points.

ianchakeres: Thanks for the fast feedback. I've updated the description accordingly.

local-volume/README.md
```diff
@@ -37,11 +37,20 @@ What doesn't work and workarounds:
...
         * Run a workaround controller that unbinds PVCs for pods that are
           stuck pending. TODO: add link
-* External provisioner cannot correctly detect capacity of mounts added after it
-  has been started.
-    * This requires mount propagation to work, which is targeted for 1.9.
-    * Workaround: Before adding any new mount points, stop the daemonset, add
-      the new mount points, start the daemonset.
+* The provisioner does not correctly detect mounts added after it
+  has been started due to lack of mount propogation.
+  * The provisioner may also incorrectly create local PVs with the root
```
msau42: I would keep the first sentence and  s/may/will.

The rest of this paragraph I would remove because the race condition doesn't exist in current code.

PR #489, opened 2017-11-29
Flexprovisoner use k8s.io/util/exec and Passing options parameters when provisioner/delete

fisherxu:
Flexprovisoner should use k8s.io/util/exec and Passing options parameters when provisioner/delete to the plugin, and alse add some new command to the example.

### DIFF
--- a/flex/deploy/docker/flex-provision.sh
+++ b/flex/deploy/docker/flex-provision.sh
@@ -26,13 +26,12 @@ usage() {
 }
 
 err() {
-	echo -ne $* 1>&2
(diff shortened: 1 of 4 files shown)


childsb:
Thanks for the PR @fisherxu .. these changes look OK to me, but i'm going to wait for @wongma7 to give final review.

fisherxu:
Kindly ping @wongma7 :)

wongma7:
/lgtm

thanks, using the same flex driver call wrapping code as kubernetes makes sense to me.

PR #493, opened 2017-11-30
Use snap UUID string instead of time format string.

humblec:

### DIFF
--- a/snapshot/pkg/volume/gluster/snapshot.go
+++ b/snapshot/pkg/volume/gluster/snapshot.go
@@ -19,12 +19,12 @@ package gluster
 import (
 	"fmt"
 	"os/exec"
-	"time"
 
 	"github.com/golang/glog"
 	crdv1 "github.com/kubernetes-incubator/external-storage/snapshot/pkg/apis/crd/v1"
 	"github.com/kubernetes-incubator/external-storage/snapshot/pkg/cloudprovider"
 	"github.com/kubernetes-incubator/external-storage/snapshot/pkg/volume"
+	"github.com/pborman/uuid"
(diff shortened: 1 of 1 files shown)


snapshot/pkg/volume/gluster/snapshot.go
```diff
@@ -60,9 +60,12 @@ func (h *glusterfsPlugin) SnapshotCreate(pv *v1.PersistentVolume, tags *map[stri
 	}
 
 	volumePath := spec.Glusterfs.Path
-	t := time.Now()
-	snapshotName := volumePath + "_" + t.Format(snapTimeFormat)
-
+	uuid := make([]byte, 16)
+	_, err := rand.Read(uuid)
```
tsmetana: This looks rather wild... Wouldn't it be better to use "proper" UUID? We already have https://github.com/pborman/uuid among the dependencies anyway so I would definitely prefer using that.

humblec: @tsmetana Sure.. Used `uuid` from referenced repo. Thanks.

humblec:
@tsmetana Addressed the comment. Please review.

humblec:
@tsmetana Can you please review this PR ?  @wongma7 the test failure is not related to this PR. 
```
[ERROR]	Failed to install: https://cloud.google.com/go Not Found

make[1]: *** [test-e2e] Error 1

make[1]: Leaving directory `/home/travis/gopath/src/github.com/kubernetes-incubator/external-storage/nfs'

make: *** [test-nfs] Error 2


```

Can you please take a look at it ?

humblec:
@wongma7 PTAL :)

PR #496, opened 2017-11-30
gluster:block: fix allocation size calculation

obnoxxx:
Heketi interprets sizes as GiB, not GB.

### DIFF
--- a/gluster/block/cmd/glusterblock-provisioner/glusterblock-provisioner.go
+++ b/gluster/block/cmd/glusterblock-provisioner/glusterblock-provisioner.go
@@ -182,7 +182,7 @@ func (p *glusterBlockProvisioner) Provision(options controller.VolumeOptions) (*
 	// Calculate the size
 	volSize := options.PVC.Spec.Resources.Requests[v1.ResourceName(v1.ResourceStorage)]
 	volSizeBytes := volSize.Value()
-	volszInt := int(util.RoundUpSize(volSizeBytes, 1000*1000*1000))
+	volszInt := int(util.RoundUpToGiB(volSizeBytes))
 
 	// Create gluster block Volume
 	blockVolName := ""
--- a/lib/util/util.go
+++ b/lib/util/util.go
@@ -35,6 +35,11 @@ func RoundUpSize(volumeSizeBytes int64, allocationUnitBytes int64) int64 {
 	return (volumeSizeBytes + allocationUnitBytes - 1) / allocationUnitBytes
 }
 
+// RoundUpToGiB rounds up given quantity upto chunks of GiB
+func RoundUpToGiB(sizeBytes int64) int64 {
+	return RoundUpSize(sizeBytes, GiB)
+}
+
 // AccessModesContains returns whether the requested mode is contained by modes
 func AccessModesContains(modes []v1.PersistentVolumeAccessMode, mode v1.PersistentVolumeAccessMode) bool {
 	for _, m := range modes {


obnoxxx:
@wongma7, @humblec - please take a look!

obnoxxx:
fixed build error (brownpaperbag)

obnoxxx:
@jsafrane - ptal :-)

gluster/block/cmd/glusterblock-provisioner/glusterblock-provisioner.go
```diff
@@ -182,7 +182,7 @@ func (p *glusterBlockProvisioner) Provision(options controller.VolumeOptions) (*
 	// Calculate the size
 	volSize := options.PVC.Spec.Resources.Requests[v1.ResourceName(v1.ResourceStorage)]
 	volSizeBytes := volSize.Value()
-	volszInt := int(util.RoundUpSize(volSizeBytes, 1000*1000*1000))
+	volszInt := int(util.RoundUpToGiB(volSizeBytes))
```
ianchakeres: We could just use RoundUpSize and reference GiB, instead of creating a new method.

Here is an example - https://github.com/kubernetes-incubator/external-storage/blob/master/local-volume/provisioner/pkg/discovery/discovery.go#L213

obnoxxx: Sure, we could do that. If that's preferred, I'm happy to change the PR.
I was mainly following the pattern of https://github.com/kubernetes/kubernetes/pull/56581 that introduced the corresponding function in Kubernetes proper.

humblec:
@obnoxxx the PR justification is not correct. Heketi does not do any size calculation when it create gluster block volumes  https://github.com/heketi/heketi/blob/9d2ef4bbaf547f5b86055e935f1d583196a069f1/executors/sshexec/block_volume.go#L45. It always append "G" String with the command. That said, this case is different from the referenced kube issue. Ideally gluster block should have the capability of treating GB and GiB sizes properly. Accordingly we should have the proper calling from each layer . ie `heketi` and `gluster-block` has to be fixed first.

obnoxxx:
@humblec oops, that's true, I did not check thoroughly enoug. Sorry for the noise. :-)

obnoxxx:
@humblec as you say, heketi does not calculate but only adds `G` to the `gluster-block` cli for the size parameter. But `gluster-block` actually interprets this as GiB: https://github.com/gluster/gluster-block/blob/master/utils/common.c#L78 . Hence I still think my patch is currently correct (and required for the provisioner to behave correctly).

humblec:
@obnoxxx thats what exactly I was pointing out. Before I put my comment, I went through gluster block code and noticed it treat both `G` and Gi` same way which is not correct. ie if we request `G` we get `Gi` space. Thats why I said, first this has to be fixed in  gluster block. Once we have it then we need to fix the other layers. Because provisioner only know about Heketi and thats `G` for now.

humblec:
An update: The size calculation will be fixed in gluster-block by following issue https://github.com/gluster/gluster-block/issues/44.

obnoxxx:
@humblec yeah, `gluster-block` currently treats `G` and `Gi` the same, but also `Grrr` and `Goofy` would have the same effect. In that, it's a little imprecise. Saying this is not correct is stretching it a little bit. (Afaik, there is a standard meaning of GB and GiB, but not of G or Gi.) The manpage for `gluster-block` makes it clear that only units in powers of 1024 (GiB etc) are supported:
```
size in B|KiB|MiB|GiB|TiB|PiB ...
```

So **fact number 1** is that if a cli specifies 1G for the gluster-block cli, it gets 1GiB.

**Fact number 2** is that heketi uses `sizeG` as size parameter for `gluster-block`.

**Hence** (combining fact 1 and fact 2), the `size` parameter of a BlockVolumeCreateRequest is interpreted in units of (1024)^3 (GiB) by heketi. This is all a user of heketi (e.g. the gluster-block provisioner) should care about.

So a user of heketi's block-volume create api should input size in units of GiB. Hence the gluster-block provisioner is currently processing its input (which is in bytes) wrongly, because it converts it to units of (1000)^3 instead of (1024)^3. So the only patch strictly needed to fix the current wrong behavior is this proposed one.

Now we could stop here, but you are correct in that both the authors of gluster-block and heketi have been a little imprecise and mostly used the non-standard/colloquial meaning of gigabytes (GB) which is (1024)^3, and by more of a standard called gibibytes (GiB). In the heketi API description, this was already pointed out by https://github.com/kubernetes/kubernetes/pull/56581 (for volume create). And for gluster-block, it can be seen in the error message in https://github.com/gluster/gluster-block/blob/master/utils/common.c#L95.

So what we could and should do in addition to this patch to cosmetically improve the situation is:
- fix the heketi API comments to state GiB instead of GB.
- improve the gluster-block error message
- possibly change gluster-block to be more strict about its unit parsing, e.g. only allow GiB etc as units (but there's a potential problem here since that would change the behavior that's already released).

obnoxxx:
FYI: proposed an update of heketi documentiation and code to mention GiB:  https://github.com/heketi/heketi/pull/935

humblec:
As the comments are addressed and both the other layers ( `heketi` and `gluster-block` ) are fixed/adjusted wrt to this change , this patch is good to go!. Thanks @obnoxxx . 
/lgtm
@wongma7 can you please merge this PR ?

obnoxxx:
thanks @humblec !

PR #497, opened 2017-11-30
Allow docker image for cephfs provisioner to build the provisioner itself

farcaller:
I've added a pre-step to build the cephfs-provisioner binary within docker, so this dockerfile could be used in e.g. quay automated builds.

Due to how vendoring is set up, the context of docker build has to be the root of the repository itself.


rootfs:
can you make a separate dockerfile for this? it is good thing to have a portable way to build but bad if we just want to deploy the binary when we already have a good build.

farcaller:
Done.

PR #498, opened 2017-12-01
local-volume: No need to pass MY_NAMESPACE to provisioner

redbaron:


wongma7:
it seems it's used here https://github.com/kubernetes-incubator/external-storage/blob/fd1946598363284474dd664cc75eaa905f9c0679/local-volume/provisioner/cmd/main.go#L63 , let's keep it for now

PR #499, opened 2017-12-03
[local-volume] Check that discovered directories are mountpoints before creating local PV.

ianchakeres:
This PR checks whether discovered directories are actual mountpoints before creating local PVs.

This PR is part of a fix for https://github.com/kubernetes-incubator/external-storage/issues/370 & https://github.com/kubernetes-incubator/external-storage/issues/482. In a subsequent PR, I intend to provide instructions for enabling mount propagation in Kubernetes 1.8+ to fix these two issues.

### DIFF
--- a/local-volume/provisioner/pkg/discovery/discovery.go
+++ b/local-volume/provisioner/pkg/discovery/discovery.go
@@ -111,6 +111,13 @@ func (d *Discoverer) discoverVolumesAtPath(class string, config common.MountConf
 		return
 	}
 
+	// Retreive list of mount points to iterate through discovered paths (aka files) below
+	mountPoints, mountPointsErr := d.RuntimeConfig.Mounter.List()
+	if mountPointsErr != nil {
+		glog.Errorf("Error retreiving mountpoints: %v", err)
+		return
+	}
+
 	for _, file := range files {
 		// Check if PV already exists for it
 		pvName := generatePVName(file, d.Node.Name, class)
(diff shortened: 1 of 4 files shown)


local-volume/provisioner/pkg/discovery/discovery.go
```diff
@@ -111,6 +111,13 @@ func (d *Discoverer) discoverVolumesAtPath(class string, config common.MountConf
 		return
 	}
 
+	// Cache mount points to iterate through discovered paths (aka files) below
+	mountPoints, mountPointsErr := d.RuntimeConfig.Mounter.List()
```
msau42: Did you check that this also returns bind mounts?

msau42: After this change goes in, I'll need to tag a new version of the provisioner, and then the local e2e tests need to be updated to do bind mounts for directories.

local-volume/provisioner/pkg/discovery/discovery.go
```diff
@@ -111,6 +111,13 @@ func (d *Discoverer) discoverVolumesAtPath(class string, config common.MountConf
 		return
 	}
 
+	// Cache mount points to iterate through discovered paths (aka files) below
```
dhirajh: Won't this permanent cache make it harder for provisioner to discover newly mounted volumes? I think in 1.9 we will finally get mount propagation which will allow provisioner to discover new mount points on the fly and without a container restart.

ianchakeres: This PR isn't caching them across multiple runs, but within a single call of discoverVolumesAtPath. Instead of gathering the mountpoints when checking every mountpoint. I'll modify the text, and remove the word Cache.

dhirajh: Ah, my bad.

local-volume/provisioner/pkg/discovery/discovery.go
ianchakeres: @msau42 - I ran the following test to verify that bind mounts are also returned inside the provisioner container.
```
1. mkdir /tmp/var2
2. mkdir -p /var/log/var2
3. mount --bind /tmp/var2 /var/log/var2
```
Next, I executed Mounter.List() and printed the following tuple for each mount point: `<mountPoint.Device:mountPoint.Path:mountPoint.Type>`

The discovered mount point list that was enumerated showed the following tuple:
`overlay:/var/log/var2:overlay`

Let me know if you have any questions.

ianchakeres: @msau42 - I'll open an issue and start working on this change to the e2e tests.

PR #505, opened 2017-12-05
Flex improvements

sathieu:


sathieu:
I've done only one pull-request. I can split it if needed.

sathieu:
@childsb @wongma7 @kokhang Gentle ping

Monnoroch:
@sathieu hi, I've read the code and I have a question. Right now the flex volume driver script doesn't actually mount anything neither here nor in the master. Now suppose I would take this PR, generate an image from it, deploy it on my cluster which has an actual real driver deployed that actually mounts volumes. A volume -- is a directory on the host named `/var/lib/kubelet/pods/$POD_ID/volumes/$DRIVER_VENDOR~$DRIVER_NAME/$VOLUME_NAME`. But the script will be running inside a container, even though the script itself is mounted from the host. Then how do I mount a directory on a host machine from inside a container in this provider's pod?

sathieu:
@Monnoroch The mount phase is not run from the container but from the kubelet. In the container, only the provision and delete phase are run. After the provision phase, a Flex PersistentVolume is created which triggers the other phases from kubelet.

This is not well documented, and I'm not sure to understand the workflow completely. In my usecase, I only need to bind mount a [Rozofs](https://github.com/rozofs/rozofs/) directory in the container. We may improve this later, but currently it does the job.

Monnoroch:
@sathieu thanks! In that case, why does this change assumes the script is located on the host, when it's much more convenient to have the script in the pod's container image, as it is in the master branch?

sathieu:
@Monnoroch This allows a generic image which can be pushed to a registry for usage of everyone (i.e no need to "docker build", you just need to write a script and copy it to all nodes).

As the flex driver needs a script to be deployed on all nodes anyway this doesn't add any complexity.

Monnoroch:
@sathieu But in the master branch they don't deploy the script to nodes. Does the master not work right now?

sathieu:
"don't deploy the script to nodes", since when? And master branch of which project: k8s or external-storage?

Monnoroch:
Here: https://github.com/kubernetes-incubator/external-storage/tree/master/flex. And here's the documentation:

> The provisioner runs in a pod, so the shell script which provisions/deletes the volumes must also be included in the POD's container (and not on the host).

sathieu:
Yes. The provisioner script runs in the pod, but the flex driver script runs from the node (kubelet). See: https://github.com/kubernetes/community/blob/master/contributors/devel/flexvolume.md:
> Install the vendor driver on all nodes

So you already need to deploy a script on all node. You probably need a config management tool to do it (Puppet, Salt, Ansible, Chef, ...) or just rsync it. Bonus: the same script can be used for both the provisioner and as flex vendor driver. This is what we do.

Monnoroch:
Okay, I'm starting to understand. Last question then. [Here](https://github.com/kubernetes-incubator/external-storage/blob/master/flex/deploy/docker/Dockerfile#L21) they have BOTH the provisioner binary and the flex driver script inside the container image. This is what made me think it's possible to not deploy the driver on all nodes. Are you saying that if I would add an actual mount command to the current master branch it would not work because the driver script is inside the container, not on the host?

sathieu:
in [the example](https://github.com/kubernetes-incubator/external-storage/blob/master/flex/deploy/docker/Dockerfile#L21) they are installing the provisionner and the provisionner script which happen to be also a flex driver script.
IMO this is not possible to not deploy the driver on all nodes (sorry for the double not, read: you *need* to deploy on all nodes).

Monnoroch:
Ok, thanks a lot, this is super helpful! I really think this should be better documented though.

sathieu:
Yes I agree that this is not documented enough, and I hope y PR will be reviewed soon to improve this.

Monnoroch:
@sathieu I've just tested it and this:
> The provisioner script runs in the pod, but the flex driver script runs from the node (kubelet). 

is not true. The provisioner script is mounted into the pod and it does indeed run from within the container, not on the node. I.e. `uname -a` gives me the pod name as a hostname.

sathieu:
@wongma7 PR updated. The CI fails for an unrelated reason.

sathieu:
Thanks !

Next step would be to push this image  (which is now generic) to some public docker registry...

wongma7:
@sathieu okay, I am working on it. The images are now at https://quay.io/repository/external_storage/flexprovisioner?tab=tags but it seems the base image is now outdated, so I'm going to pick launcher.gcr.io/google/debian9 from https://github.com/GoogleContainerTools/base-images-docker

PR #514, opened 2017-12-18
[glusterfs] Set permission correctly

yuanying:


humblec:
@yuanying how did you verify or whats the source of this permission ?

yuanying:
@humblec The directory which will be shared by glusterfs should be writable from user who has same GID of this directory. So I think we should provide access permission to GID.

wongma7:
makes sense to me, what is the point in giving group ownership if group does not have full permissions

PR #525, opened 2017-12-23
Fix link in Digital Ocean Documentation

tianhuil:


tianhuil:
I have just signed the linux foundation CLA.

klausenbusk:
> @klausenbusk: changing LGTM is restricted to assignees, and only kubernetes-incubator org members may be assigned issues.

I see, that is how it works. Thanks @tianhuil  :)

PR #526, opened 2017-12-24
Added yum clean to rbd Dockerfiles to reduce docker image size

ianchakeres:
`yum clean all` reduces the size of the related docker images.


ianchakeres:
PTAL - @wongma7 

FYI: @cofyc, @kairen, & @mcronce

PR #532, opened 2017-12-29
aggregate errors and omit delete not found error

hzxuzhonghu:
1. aggregate errors in DeleteContents

2. when delete a non-exist pv,  just omit the error.

### DIFF
--- a/local-volume/provisioner/pkg/deleter/deleter.go
+++ b/local-volume/provisioner/pkg/deleter/deleter.go
@@ -24,6 +24,7 @@ import (
 
 	"bufio"
 	"k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	"os"
 	"os/exec"
 	"strings"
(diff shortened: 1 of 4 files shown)


ianchakeres:
@hzxuzhonghu - can you please add a description to this PR? For example, what's changing and why.

local-volume/provisioner/pkg/deleter/deleter.go
```diff
@@ -129,12 +130,13 @@ func (d *Deleter) asyncDeletePV(pv *v1.PersistentVolume, volType string, mountPa
 
 	// Remove API object
 	if err := d.APIUtil.DeletePV(pv.Name); err != nil {
-		// TODO: Does delete return an error if object has already been deleted?
-		deletingLocalPVErr := fmt.Errorf("Error deleting PV %q: %v", pv.Name, err.Error())
-		d.RuntimeConfig.Recorder.Eventf(pv, v1.EventTypeWarning, common.EventVolumeFailedDelete,
-			deletingLocalPVErr.Error())
-		glog.Error(deletingLocalPVErr)
-		return
+		if !errors.IsNotFound(err) {
```
ianchakeres: Are there any other types of errors (besides IsNotFound) that could occur? 

If yes, should we continue for those errors or log an error/event & return?

hzxuzhonghu: No, I try with kubectl delete and find kube-apiserver just retuen 404.

ianchakeres: If this is the only type of error ever returned by DeletePV, I think that we can remove the if statement check for IsNotFound, aka `if !errors.IsNotFound(err) {`.

hzxuzhonghu: I mean if no pv exists, delete will only fail with `NotFound error`, there are still other cases like `timeout`, etc. So This if statement should not be removed.

msau42: Can you add a unit test for this?

hzxuzhonghu: @msau42 add a ut case for deleting non-exist pv

local-volume/provisioner/pkg/util/api_util.go
```diff
@@ -98,8 +99,9 @@ func (u *FakeAPIUtil) DeletePV(pvName string) error {
 		u.deletedPVs[pvName] = pv
 		delete(u.createdPVs, pvName)
 		u.cache.DeletePV(pvName)
+		return nil
 	}
-	return nil
```
ianchakeres: Can we please create an errList with multiple NewNotFound errors, and return utilerrors.NewAggregate(errList)? This change will mirror the return in volume_util.go:L123.

Additionally, it might be worth creating a test case where only one of the errors in the list is a NewNotFound error.

hzxuzhonghu: @ianchakeres I donot think a errList is needed, since here we only result at most one error.

PR #533, opened 2017-12-29
Move from glide to dep

hzxuzhonghu:
> dep is production-ready and supposedly better. it also has vendor pruning built in so we can use that instead of having people use the third-party glide-vc tool to keep the vendor dir size down

1. `dep init`

addition we need to do:
 https://github.com/golang/dep/blob/master/docs/Gopkg.toml.md#ignored

>Use wildcard character (*) to define a package prefix to be ignored. Use this to ignore any package and their subpackages.

`ignored = ["github.com/kubernetes-incubator/external-storage/repo-infra*"]`


hzxuzhonghu:
ping @msau42 @wongma7

rootfs:
Been struggled with glide long time. If dep solves the glide problems, i am all for it ...

dmick:
This commit fails in "dep ensure" for me: 
```
$ ~/go/bin/dep ensure
Solving failure: No versions of github.com/bazelbuild/buildifier met constraints:
	0.6.0: Could not introduce github.com/bazelbuild/buildifier@0.6.0, as its subpackage github.com/bazelbuild/buildifier/core is missing. (Package is required by (root).)
	0.4.5: Could not introduce github.com/bazelbuild/buildifier@0.4.5, as its subpackage github.com/bazelbuild/buildifier/core is missing. (Package is required by (root).)
	0.4.3: Could not introduce github.com/bazelbuild/buildifier@0.4.3, as its subpackage github.com/bazelbuild/buildifier/core is missing. (Package is required by (root).)
	master: Could not introduce github.com/bazelbuild/buildifier@master, as its subpackage github.com/bazelbuild/buildifier/core is missing. (Package is required by (root).)
	fix-ci-badge: Could not introduce github.com/bazelbuild/buildifier@fix-ci-badge, as its subpackage github.com/bazelbuild/buildifier/core is missing. (Package is required by (root).)

$ ~/go/bin/dep version
dep:
 version     : devel
 build date  : 
 git hash    : 
 go version  : go1.8.3
 go compiler : gc
 platform    : linux/amd64
```

dmick:
....and I apparently can't get past that error.  it appears that github.com/bazelbuild/buildifier has disappeared.

dmick:
https://github.com/bazelbuild/rules_go/issues/361  Has no one updated dependencies since last April?...why doesn't this affect glide?....so many questions

hzxuzhonghu:
@dmick remove repo-infra directory and exec  dep command, it is ok. And I think the repo-infra is from https://github.com/kubernetes/repo-infra

dmick:
OK, but, repo-infra is part of the project; how is it a solution to just remove it?  And if that's the solution, shouldn't this PR include a commit removing it?

wongma7:
if there is a way to make dep ignore repo-infra that would be ideal, like this https://github.com/kubernetes-incubator/external-storage/blob/68fd702f0c839d91cfabbe480e414706560ae386/glide.yaml#L5
otherwise we can delete the repo-infra directory, vendor it like so https://github.com/kubernetes/repo-infra#using-vendoring, and change these two lines https://github.com/kubernetes-incubator/external-storage/blob/1eafa705922fff62072c2a7876ba3335217ead3f/Makefile#L26
caveat to vendoring repo-infra: we will lose my small patch to make directories ignorable from (imo) naggy golint but that's no big loss since almost everything already complies with golint already

hzxuzhonghu:
@wongma7 I would like to do some research about `if there is a way to make dep ignore repo-infra that would be ideal, like this`

dmick:
well, strangely, https://github.com/kubernetes-incubator/external-storage/pull/533/commits/7e8bdea7e8a613ace7d780b3e7766a1673930f97#diff-836546cc53507f6b2d581088903b1785R25, but I don't know why it's apparently not working

hzxuzhonghu:
```
Use wildcard character (*) to define a package prefix to be ignored. Use this to ignore any package and their subpackages.

ignored = ["github.com/user/project/badpkg*"]
```

hzxuzhonghu:
@dmick Add * is ok.

dmick:
Agreed; with "repo-infra*", dep ensure completes.

cofyc:
@hzxuzhonghu Do you have time to continue on this? Using `glide` is painful. I hope we can migrate to `dep`.

hzxuzhonghu:
ping @wongma7 ptal

wongma7:
/lgtm
thank you very much @hzxuzhonghu this seems too good to be true :)

PR #543, opened 2018-01-05
nfs: Upgrade NFS-Ganesha (slightly)

afbjorklund:
The nfsd.ganesha seems to be crashing every once in a while, maybe a software update would help

Eventually it should probably be updated to the latest fedora version and the packaged nfs-ganesha ?


afbjorklund:
build error due to package server timeout

afbjorklund:
ping @wongma7

wongma7:
/lgtm
it might help, wouldn't hurt to try. 

note I tried updating to ganesha 2.5 and it was much less stable for I don't know what reason so we might be stuck here for a while.

PR #552, opened 2018-01-11
RBD Provisioner: Add userSecretNamespace parameter.

cofyc:
 
With https://github.com/kubernetes/kubernetes/commit/1411c2698eeb3cfe8a319fe0cdb1a7155249dc6e, RBD Persistent Volume now can reference secret from non-PVC/POD namespace. So we can support an additional `userSecretNamespace` parameter. With it, administrator do not need
configure use secret for each user (assuming each user may use a separate namespace).

This requires [k8s.io/api@1.9](https://github.com/kubernetes/api/tree/release-1.9) (v1. RBDPersistentVolumeSource added in 1.9). I will test when dependencies are updated to 1.9.


cofyc:
@rootfs Tests passed.

rootfs:
thanks, please also update rbac role - secrets now don't have be be a cluster role any more

cofyc:
@rootfs Updated.

rootfs:
/lgtm
/approve

cc @childsb for merge

PR #557, opened 2018-01-15
Check for all Digital Ocean Volumes

ZachGlassman:
The flexplugin was not checking for all volumes since the api returns
volumes by page where each page has a limited number of volumes.

This commit alters the getVolumeByName function so that it checks all volumes by continuing to query pages until no results are found.

/area digitalocean

### DIFF
--- a/digitalocean/cmd/digitalocean-flexplugin/main.go
+++ b/digitalocean/cmd/digitalocean-flexplugin/main.go
@@ -110,14 +110,27 @@ func (c *cloud) findNode(nodeName string) (int, error) {
 }
 
 func (c *cloud) getVolumeByName(volumeName string) (string, error) {
-	volumes, _, err := c.client.Storage.ListVolumes(c.ctx, nil)
-	if err != nil {
-		return "", err
-	}
-	for _, volume := range volumes {
-		if volume.Name == volumeName {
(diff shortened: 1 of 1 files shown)


ZachGlassman:
all signed

klausenbusk:
Looks good! I'm wondering if using `[ListVolumeParams](https://godoc.org/github.com/digitalocean/godo#ListVolumeParams)` would be a better solution? What do you think?

ZachGlassman:
Do you mean using the ListOptions to set the number of results to include per page to something large?

klausenbusk:
> Do you mean using the ListOptions to set the number of results to include per page to something large?

No, I mean the `ListVolumeParams` (not `ListOptions`), which allow DigitalOcean to do some filtering server side.

Nvm: I just tested it, it require the region of the volume which isn't available when attaching/detaching.

klausenbusk:
/approve

Please squash the commits to a single commit, everything else looks great :)

ZachGlassman:
@klausenbusk Thanks!  Squash done

klausenbusk:
Great, I have just pushed the updated docker images.

ZachGlassman:
Excellent thank you!

PR #559, opened 2018-01-16
Generate local provisioner specs for GCE

msau42:
GCE/GKE local SSD environment is fixed and typically does not need user modification, so the same provisioner spec can be used for most users.


msau42:
cc @davidz627

local-volume/provisioner/deployment/kubernetes/gce/provisioner_generated_gce_pre19.yaml
```diff
@@ -0,0 +1,56 @@
...
+ 
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: local-provisioner-config 
+  namespace: default 
+data:
+  storageClassMap: |     
+    local-scsi:
+       hostDir: /mnt/disks
```
davidz627: Could this yaml ever be run with post 1.9 and with local SSD volume configs? If so, these would pick up the entire directory including the new "by-uuid" directory, so we would have *double* the mounts.

Not sure if the provisioner only scans top level dir. Just want to make sure.

local-volume/README.md
```diff
@@ -109,27 +109,53 @@ $ export KUBE_FEATURE_GATES="PersistentLocalVolumes=true,VolumeScheduling=true,M
...
+
 # If running K8s 1.9+, also create the StorageClasses
-$ kubectl create -f examples/deployment/kubernetes/gce/class-local-ssds.yaml
+$ kubectl create -f provisioner/deployment/kubernetes/gce/class-local-ssds.yaml
+$ kubectl create -f provisioner/deployment/kubernetes/gce/provisioner_generated_gce_pre19.yaml
+```
+
+##### Using local-ssd-config option (available via whitelist only)
+``` console
+$ gcloud alpha container cluster create ... --local-ssd-config=<n>,<scsi|nvme>,fs --enable-kubernetes-alpha
```
davidz627: --local-ssd-volumes=

local-volume/README.md
```diff
@@ -109,27 +109,53 @@ $ export KUBE_FEATURE_GATES="PersistentLocalVolumes=true,VolumeScheduling=true,M
...
+$ gcloud alpha container node-pools create ... --local-ssd-config=<n>,<scsi|nvme>,fs
```
davidz627: --local-ssd-volumes

local-volume/README.md
```diff
@@ -109,27 +109,53 @@ $ export KUBE_FEATURE_GATES="PersistentLocalVolumes=true,VolumeScheduling=true,M
...
 $ gcloud container node-pools create ... --local-ssd-count=<n>
+$ kubectl create -f provisioner/deployment/kubernetes/admin_account.yaml
+
 # If running K8s 1.9+, also create the StorageClasses
-$ kubectl create -f examples/deployment/kubernetes/gce/class-local-ssds.yaml
+$ kubectl create -f provisioner/deployment/kubernetes/gce/class-local-ssds.yaml
+$ kubectl create -f provisioner/deployment/kubernetes/gce/provisioner_generated_gce_pre19.yaml
+```
+
+##### Using local-ssd-config option (available via whitelist only)
```
davidz627: local-ssd-volumes

davidz627:
besides comments above, lgtm

local-volume/provisioner/deployment/kubernetes/gce/provisioner_generated_gce_pre19.yaml
msau42: Ooh interesting, I will try this out.  Looking at the code, it only looks at the top level directory, and if the file is not a mount point, it will be ignored.  So the worst case is that the provisioner will just be periodically logging that this "by-uuid" directory is not a mountpoint.

msau42: Yeah just tried it out, so we will not create double the PVs, but there are 2 minor issues:
1. Provisioner will be logging "Path "/mnt/disks/by-uuid" is not an actual mountpoint" every 10 seconds
2. All the scsi and nvme PVs will be created under the "local-scsi" storageclass since they are all in the same directory.

If we eventually add these as addons, hopefully that can help eliminate these potential user errors.

local-volume/provisioner/deployment/kubernetes/gce/provisioner_generated_gce_19.yaml
```diff
@@ -0,0 +1,64 @@
...
+# Source: provisioner/templates/provisioner.yaml
+ 
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: local-provisioner-config 
+  namespace: default 
+data:
+  storageClassMap: |     
+    local-scsi:
```
msau42: One more thing I'm considering: Instead of naming these storage classes "local-scsi" and "local-nvme", maybe "local-fast" and "local-faster" to be more portable across environments.

local-volume/provisioner/deployment/kubernetes/gce/provisioner_generated_gce_pre19.yaml
davidz627: to me these seem like acceptable "failure modes" due to user error. Agree that making these addons would solve this issue.

local-volume/provisioner/deployment/kubernetes/gce/provisioner_generated_gce_19.yaml
davidz627: I can see how this would be useful for portability, but not liking the "local-fast" and "local-faster" names. Hard to disambiguate, and if we go down this route what happens when we need a "local-fasterer"

ianchakeres: Maybe "standard" or "local-ssd".

msau42: Yeah, the challenge with "local-ssd" is that on GCE, there are two kinds, scsi and nvme.  And who knows, maybe there will be more in the future.

msau42: I'm going to keep the naming the way it is for now since I can't think of a better name.  Users can still generate their own specs with different storage class names if they want.

davidz627: Agreed. I also can't think of anything better.

local-volume/README.md
```diff
@@ -109,27 +109,53 @@ $ export KUBE_FEATURE_GATES="PersistentLocalVolumes=true,VolumeScheduling=true,M
...
+##### Using local-ssd-count option
 ``` console
 $ gcloud container cluster create ... --local-ssd-count=<n> --enable-kubernetes-alpha
 $ gcloud container node-pools create ... --local-ssd-count=<n>
+$ kubectl create -f provisioner/deployment/kubernetes/admin_account.yaml
+
 # If running K8s 1.9+, also create the StorageClasses
-$ kubectl create -f examples/deployment/kubernetes/gce/class-local-ssds.yaml
+$ kubectl create -f provisioner/deployment/kubernetes/gce/class-local-ssds.yaml
+$ kubectl create -f provisioner/deployment/kubernetes/gce/provisioner_generated_gce_pre19.yaml
```
davidz627: should this be not "pre"19

davidz627:
besides 1 last comment in README.md. /lgtm

local-volume/README.md
msau42: If they're using the old --local-ssd-count, then they need to use the pre1.9 config which looks at /mnt/disks.

Yeah, the naming is confusing.  What do you think of naming the file "v1" or "v2" instead of "pre19" and "19"

davidz627: v1 or v2 doesn't seem great to me, they don't really mean anything on their own and it implies you can always use v2 because its newer/better.

Then again I can't think of anything better. I'm not against pre19 and 19. Maybe add a comment for that particular call?

msau42: Pre 1.9 is confusing because in GKE you could still use the old count option with 1.9+.

msau42:
/assign @wongma7 

Could you also add David to this repo?

wongma7:
/lgtm

okay, email requesting so has been sent 👍

PR #563, opened 2018-01-19
Support noauth connections from the environment

aglitke:
Noauth support was added by a recent commit but it only worked when used
with the config file approach.  Add support for parsing the environment.
If OS_AUTH_URL is set, we will use a Keystone approach.  If
OS_CINDER_ENDPOINT is set we will use noauth.  Add connection tests to
validate that this is working as intended.

Signed-off-by: Adam Litke <alitke@redhat.com>

### DIFF
--- a/openstack/standalone-cinder/pkg/volumeservice/volumeservice_test.go
+++ b/openstack/standalone-cinder/pkg/volumeservice/volumeservice_test.go
@@ -1,0 +1,107 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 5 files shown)


aglitke:
cc: @j-griffith

aglitke:
@wongma7 @jsafrane And merge this one too?

openstack/standalone-cinder/pkg/volumeservice/volumeservice_test.go
```diff
@@ -0,0 +1,100 @@
...
+var _ = Describe("Volume Service", func() {
+	Describe("Connection configuration parsing", func() {
+		var (
+			config cinderConfig
+			err    error
+		)
+
+		Context("when config file path is empty", func() {
+			BeforeEach(func() {
+				os.Setenv("OS_AUTH_URL", "AUTH_URL")
```
rootfs: make AUTH_URL and CINDER_ENDPOINT const and refer them by name

rootfs:
some nit, please also update deployment or readme with cinder endpoint param.

PR #564, opened 2018-01-19
[standalone-cinder] Determine volume create options in provisioner

aglitke:
Cinder volume creation parameters need to be determined by looking at
the PVC.  Move this logic out of the volumeservice package and into the
provisioner.  This allows volumeservice to be cleaner and completely
ignorant of kubernetes details.

Signed-off-by: Adam Litke <alitke@redhat.com>

### DIFF
--- a/openstack/standalone-cinder/pkg/provisioner/provisioner.go
+++ b/openstack/standalone-cinder/pkg/provisioner/provisioner.go
@@ -19,13 +19,17 @@ package provisioner
 import (
 	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/golang/glog"
 	"github.com/gophercloud/gophercloud"
 	"github.com/kubernetes-incubator/external-storage/lib/controller"
 	"github.com/kubernetes-incubator/external-storage/openstack/standalone-cinder/pkg/volumeservice"
 	"k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/uuid"
 	"k8s.io/client-go/kubernetes"
+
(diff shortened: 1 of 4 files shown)


openstack/standalone-cinder/pkg/provisioner/provisioner.go
```diff
@@ -19,13 +19,17 @@ package provisioner
...
 
 	"github.com/golang/glog"
 	"github.com/gophercloud/gophercloud"
 	"github.com/kubernetes-incubator/external-storage/lib/controller"
 	"github.com/kubernetes-incubator/external-storage/openstack/standalone-cinder/pkg/volumeservice"
 	"k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/uuid"
 	"k8s.io/client-go/kubernetes"
+
+	volumes_v2 "github.com/gophercloud/gophercloud/openstack/blockstorage/v2/volumes"
```
stmcginnis: Maybe better left for another patch, but the Cinder v2 API is deprecated and all clients should be using v3. New deployments (as of the Pike release) do not have the v2 API enabled by default.

aglitke: Thanks for reminding me of this!  I'd like to adopt v3 but this will be a larger effort (especially due to some changing auth semantics.

aglitke:
@wongma7 @jsafrane can we merge this?

rootfs:
/lgtm

cc @childsb for merge

PR #565, opened 2018-01-19
standalone-cinder: Use cinder attach and detach

aglitke:
Prior to this patch, provisioned cinder volumes were left in 'attaching'
state in the cinder DB.  This is an interim state that is supposed to
resolve to 'attached' or (in the case of failure) 'available'.  This
patch completes the attachment process on the cinder side so that
volumes will appear attached to cinder.
    
Note that the volume is not really attached to a user until a pod
consumes the PV and the kubelet attaches and mounts the volume.  This
descrepency between k8s and cinder is OK as each system has an
appropriate view of the volume state from its own perspective.

### DIFF
--- a/openstack/standalone-cinder/pkg/provisioner/provisioner.go
+++ b/openstack/standalone-cinder/pkg/provisioner/provisioner.go
@@ -148,25 +148,36 @@ func (p *cinderProvisioner) Provision(options controller.VolumeOptions) (*v1.Per
 		goto ERROR_UNRESERVE
 	}
 
+	err = p.vsb.attachCinderVolume(p.VolumeService, volumeID)
+	if err != nil {
+		glog.Errorf("Failed to attach volume %s: %v", volumeID, err)
+		goto ERROR_DISCONNECT
+	}
+
 	mapper, err = p.mb.newVolumeMapperFromConnection(connection)
 	if err != nil {
 		glog.Errorf("Unable to create volume mapper: %v", err)
-		goto ERROR_DISCONNECT
+		goto ERROR_DETACH
 	}
 
 	err = mapper.AuthSetup(p, options, connection)
 	if err != nil {
 		glog.Errorf("Failed to prepare volume auth: %v", err)
-		goto ERROR_DISCONNECT
+		goto ERROR_DETACH
 	}
 
 	pv, err = p.mb.buildPV(mapper, p, options, connection, volumeID)
 	if err != nil {
 		glog.Errorf("Failed to build PV: %v", err)
-		goto ERROR_DISCONNECT
+		goto ERROR_DETACH
 	}
 	return pv, nil
 
+ERROR_DETACH:
+	cleanupErr = p.vsb.detachCinderVolume(p.VolumeService, volumeID)
+	if cleanupErr != nil {
+		glog.Errorf("Failed to detach volume %s: %v", volumeID, cleanupErr)
+	}
 ERROR_DISCONNECT:
 	cleanupErr = p.vsb.disconnectCinderVolume(p.VolumeService, volumeID)
 	if cleanupErr != nil {
@@ -216,6 +227,12 @@ func (p *cinderProvisioner) Delete(pv *v1.PersistentVolume) error {
 	}
 
 	mapper.AuthTeardown(p, pv)
+
+	err = p.vsb.detachCinderVolume(p.VolumeService, volumeID)
+	if err != nil {
+		glog.Errorf("Failed to detach volume %s: %v", volumeID, err)
+		return err
+	}
 
 	err = p.vsb.disconnectCinderVolume(p.VolumeService, volumeID)
 	if err != nil {
--- a/openstack/standalone-cinder/pkg/provisioner/testutils_test.go
+++ b/openstack/standalone-cinder/pkg/provisioner/testutils_test.go
@@ -17,7 +17,10 @@ limitations under the License.
 package provisioner
 
 import (
+	"bytes"
 	"errors"
+	"fmt"
+
 	"github.com/golang/glog"
 	"github.com/kubernetes-incubator/external-storage/lib/controller"
 	"github.com/kubernetes-incubator/external-storage/openstack/standalone-cinder/pkg/volumeservice"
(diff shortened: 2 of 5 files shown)


aglitke:
@j-griffith Thoughts on this one?  It is much better to move the cinder state to attached than to leave it in attaching.

openstack/standalone-cinder/pkg/provisioner/provisioner.go
```diff
@@ -179,6 +190,12 @@ func (p *cinderProvisioner) Delete(pv *v1.PersistentVolume) error {
 
 	mapper.AuthTeardown(p, pv)
 
+	err = p.vsb.detachCinderVolume(p.VolumeService, volumeID)
+	if err != nil {
+		glog.Errorf("Failed to detach volume %s: %v", volumeID, err)
+		return err
+	}
+
```
j-griffith: These are great!  One thing I was thinking though; what about condensing these into a single recovery function for Detach related failures?  So just have a general "_handleDetachErr" and check where in the process the volume appears to be and do basically the same steps you have here?

I think it might be cleaner?

One reason I think of this is I've been looking at adding the new Attach API's to Gophercloud and *some day* would like to add them here as well; it would make things easier in the future to add new support.

aglitke: Thanks @j-griffith for the review.  I agree with you that a detach handler could be cleaner and it is something I have considered adding in the future.  This would help handle cases where one step in the cleanup fails and causes subsequent attempts at cleanup to fail.  eg. Detach succeeds but disconnect fails.  Next time we retry detach but it fails because it is already detached.

That being said, I would prefer to merge this batch of PRs first and then address this next.

j-griffith:
👍

aglitke:
@j-griffith Thanks!  How do I go about getting this merged?

j-griffith:
@aglitke Unfortunately that's not something that I can do; I think @rootfs and @wongma7 are the only ones that can do that; but haven't looked lately.

rootfs:
@aglitke can you rebase to master branch? the commits look messed up, thanks

openstack/standalone-cinder/pkg/provisioner/testutils_test.go
```diff
@@ -79,31 +81,41 @@ func createCinderProvisioner() *cinderProvisioner {
...
 		return false
 	}
 	return value
 }
 
-func (vsb *failureInjector) ret(method string) error {
-	if vsb.isSet(method) {
+func (fi *failureInjector) ret(method string) error {
+	if fi.isSet(method) {
+		return errors.New("injected error for testing")
```
rootfs: nit: add method to error

openstack/standalone-cinder/pkg/volumeservice/actions.go
```diff
@@ -109,6 +112,22 @@ func ConnectCinderVolume(vs *gophercloud.ServiceClient, volumeID string) (Volume
 	return rcv.ConnectionInfo, nil
 }
 
+// AttachCinderVolume marks the volume as attached in the cinder database.
+func AttachCinderVolume(vs *gophercloud.ServiceClient, volumeID string) error {
+	opts := volumeactions.AttachOpts{
+		MountPoint: attachMountPoint,
+		HostName:   attachHostName,
+		Mode:       volumeactions.ReadWrite,
```
rootfs: not sure if mode is needed, cannot find it in cinder api https://developer.openstack.org/api-ref/block-storage/v3/#attach-volume-to-a-server

openstack/standalone-cinder/pkg/provisioner/provisioner.go
```diff
@@ -148,25 +148,36 @@ func (p *cinderProvisioner) Provision(options controller.VolumeOptions) (*v1.Per
...
 	pv, err = p.mb.buildPV(mapper, p, options, connection, volumeID)
 	if err != nil {
 		glog.Errorf("Failed to build PV: %v", err)
-		goto ERROR_DISCONNECT
+		goto ERROR_DETACH
 	}
 	return pv, nil
 
+ERROR_DETACH:
+	cleanupErr = p.vsb.detachCinderVolume(p.VolumeService, volumeID)
```
rootfs: what if 2 volume attach calls happen at the same time, only one succeeds, leaving the 2nd fail. Now would detach call here detach the volume from the 1st (i.e. the successful) attachment?

openstack/standalone-cinder/pkg/volumeservice/actions.go
```diff
@@ -109,6 +112,22 @@ func ConnectCinderVolume(vs *gophercloud.ServiceClient, volumeID string) (Volume
 	return rcv.ConnectionInfo, nil
 }
 
+// AttachCinderVolume marks the volume as attached in the cinder database.
+func AttachCinderVolume(vs *gophercloud.ServiceClient, volumeID string) error {
+	opts := volumeactions.AttachOpts{
```
rootfs: should we check volume status here? This could help avoid (but not eliminate) race condition
https://developer.openstack.org/api-ref/block-storage/v3/#attach-volume-to-a-server
```
Preconditions
Volume status must be available.
You should set instance_uuid or host_name.
```

rootfs:
some (open) questions. Can be addressed in future PRs. An ack will do.

openstack/standalone-cinder/pkg/provisioner/provisioner.go
stmcginnis: That looks safe here with the failure in the attach call going to ERROR_DISCONNECT, and anything after that point going to ERROR_DETACH.

openstack/standalone-cinder/pkg/volumeservice/actions.go
aglitke: Actually the volume must be Available even for the reserve step.  It's why we do the polling there.

openstack/standalone-cinder/pkg/volumeservice/actions.go
aglitke: From my testing it seemed to be required.

openstack/standalone-cinder/pkg/provisioner/testutils_test.go
aglitke: good plan, will do.

openstack/standalone-cinder/pkg/provisioner/provisioner.go
rootfs: I am not clear now disconnect works in cinder. The connection info [here](https://github.com/kubernetes-incubator/external-storage/pull/565/files/026a89e75eac4544db467cd7fcdcafd819a1b63d#diff-d8837f89cf53ba71cc227ee80877f0f5R134) passed to `TerminateConnection` is the same for all connections.

aglitke:
@rootfs I've addressed your nit about the log message and have rebased to master.  Would you mind if we move forward with this PR (merge) and I can incorporate other suggestions (ie. From @j-griffith ) in a future patch?

wongma7:
/lgtm 
merging.

PR #566, opened 2018-01-21
Add yaml for hostpath snapshot restore

xing-yang:
This patch modifies README for hostpath and adds the yaml file
to restore a snapshot to a new PV.


xing-yang:
Need to fix CLA issue first.

Issue #577, opened 2018-01-26
local-volume missing serviceaccount and serviceroles in helm/templates directory


nseyvet:
The helm chart for the local-volume provisioner refers to service accounts that are not part of the templates.

See https://github.com/kubernetes-incubator/external-storage/tree/master/local-volume/helm/provisioner/templates.

Add service-account and cluster-role-bindings as defined under https://github.com/kubernetes-incubator/external-storage/blob/master/local-volume/provisioner/deployment/kubernetes/admin_account.yaml

msau42:
This can be closed now?

nseyvet:
Yes

PR #579, opened 2018-01-26
RBAC: Add service account and cluster roles

nseyvet:
Fixes issue #577 

Add rbac as an option in the Values, and generates the service account and role bindings.

### DIFF
--- a/local-volume/helm/provisioner/templates/provisioner-cluster-role-binding.yaml
+++ b/local-volume/helm/provisioner/templates/provisioner-cluster-role-binding.yaml
@@ -1,0 +1,29 @@
+{{- if .Values.rbac.create }}
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRoleBinding
+metadata:
+  name: local-storage-provisioner-pv-binding
+  namespace: {{ .Values.common.namespace }}
+subjects:
+- kind: ServiceAccount
+  name: {{ .Values.daemonset.serviceAccount }}
+  namespace: {{ .Values.common.namespace }}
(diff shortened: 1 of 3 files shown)


nseyvet:
Signed CLA.

local-volume/helm/provisioner/templates/provisioner-cluster-role-binding.yaml
```diff
@@ -0,0 +1,29 @@
+{{- if .Values.rbac.create }}
```
sbezverk: I would suggest to place rbac variable in common section of values.yam.  Also since we only create rbac, not sure if we really need two levels, rbac.create.  I would suggest suggest .Values.common.rbac.

Please add this into values.yaml file and also provide short description, so users would know that this knob exists.

nseyvet:
Ok.  Will do.

For rbac, I followed how Prometheus stable and grafana stable do it. Seems
to be the common approach. Please confirm?

sbezverk:
@nseyvet the logic here is: we have common configuration and specific. RBAC can be considered as a common component, so it is should go into .Values.common, whether to use just rbac or rbac.create, I do not see any technical reason for two levels, and just "rbac" is shorted if you want to overwrite from the command line.

msau42:
Can you update the readme too?  Step 3, option 1, currently uses a hardcoded admin account yaml.

msau42:
And you can squash the commits

nseyvet:
Created new pull request with squashed commits https://github.com/kubernetes-incubator/external-storage/pull/586

PR #580, opened 2018-01-26
Check strconv.ParseUint error.

cofyc:
`strconv.ParseUint` returns 0 and an error if argument is invalid. If we
don't check error, the following `chgrd` will succeed to change the
group of file to 0, and reports no error if the process is
running as root user.

### DIFF
--- a/nfs/pkg/volume/provision.go
+++ b/nfs/pkg/volume/provision.go
@@ -450,7 +450,11 @@ func (p *nfsProvisioner) createDirectory(directory, gid string) error {
 	}
 
 	if gid != "none" {
-		groupID, _ := strconv.ParseUint(gid, 10, 64)
+		groupID, err := strconv.ParseUint(gid, 10, 64)
+		if err != nil {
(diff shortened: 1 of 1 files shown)


nfs/pkg/volume/provision.go
```diff
@@ -450,7 +450,11 @@ func (p *nfsProvisioner) createDirectory(directory, gid string) error {
 	}
 
 	if gid != "none" {
-		groupID, _ := strconv.ParseUint(gid, 10, 64)
+		groupID, err := strconv.ParseUint(gid, 10, 64)
+		if err != nil {
+			os.RemoveAll(path)
+			return fmt.Errorf("chgrp failed with error: %v", err)
```
rootfs: chgrp -> ParseUnit

cofyc: fixed

cofyc:
@rootfs fixed

rootfs:
/lgtm

cc @childsb for merge

Issue #558, opened 2018-01-16
Validation of AWS Snapshot example failing


screeley44:
Env: single node K8 cluster created from local-up-cluster.sh (1.10 dirty) running on AWS us-east-1d

Following the Snapshot AWS example, using the latest deployment.yaml, the controller and provisioner from the deployment seem to run fine.

```
kubectl get pods
NAME                                   READY     STATUS    RESTARTS   AGE
snapshot-controller-7bb857f5dc-s9hvk   2/2       Running   0          47s
```

But when I get to the step where I actually create the snapshot (snapshot.yaml) from my PV/PVC, I don't think it is working and didn't find anything obvious in the logs to indicate why it is failing:

```
kubectl get volumesnapshot,volumesnapshotdata -o yaml --namespace=myns

apiVersion: v1
items:
- apiVersion: volumesnapshot.external-storage.k8s.io/v1
  kind: VolumeSnapshot
  metadata:
    clusterName: ""
    creationTimestamp: 2018-01-11T01:34:22Z
    generation: 0
    labels:
      SnapshotMetadata-PVName: pvc-351f5fc3-f66f-11e7-9ee7-0eef3fbd31d8
      SnapshotMetadata-Timestamp: "1515634526910793155"
    name: snapshot-demo
    namespace: myns
    resourceVersion: "1178"
    selfLink: /apis/volumesnapshot.external-storage.k8s.io/v1/namespaces/myns/volumesnapshots/snapshot-demo
    uid: 8d39e74c-f66f-11e7-9ee7-0eef3fbd31d8
  spec:
    persistentVolumeClaimName: ebs-pvc
    snapshotDataName: ""
  status:
    conditions: null
    creationTimestamp: null
kind: List
metadata:
  resourceVersion: ""
  selfLink: ""
```

Of course when I try to promote the snapshot, it fails, probably because it doesn't exist:

```
kubectl create -f snapshot-claim.yaml

# kubectl get pvc -n myns
NAME                            STATUS    VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS        AGE
ebs-pvc                         Bound     pvc-351f5fc3-f66f-11e7-9ee7-0eef3fbd31d8   4Gi        RWO            slow                25m
snapshot-pv-provisioning-demo   Pending                    
```

screeley44:
cc @tsmetana

tsmetana:
[...]

PR #583, opened 2018-01-31
Copy CA certificates to deployment containers

tsmetana:
This fixes a problem with AWS cloud provider complaining about missing root certificates and resulting failed API calls.

cc: @rootfs, @jsafrane


jsafrane:
We should not copy certificates from the machine where we build images - it can be some random bot out of our control with keys at weird places. IMO you should pick the certificates you think are needed and push them to git.

snapshot/Makefile
```diff
@@ -46,6 +50,15 @@ container: controller provisioner container-quick
 container-quick:
 	cp _output/bin/snapshot-controller deploy/docker/controller
 	cp _output/bin/snapshot-provisioner deploy/docker/provisioner
+	# Copy the root CA certificates -- cloudproviders need them
+	cp --parent -r /etc/pki/ca-trust deploy/docker/controller
```
rootfs: this does not always work especially if ca-certificates pkg is not locally installed

rootfs: this Dockerfile seems to work for me locally
```
FROM  busybox:glibc
CMD rpm -i https://www.rpmfind.net/linux/fedora/linux/development/rawhide/Everything/x86_64/os/Packages/c/ca-certificates-2017.2.20-6.fc28.noarch.rpm
CMD EXTRACT_UNSAFE_SYMLINKS=1 rpm -i ca-certificates-2017.2.20-6.fc28.noarch.rpm
```

snapshot/deploy/docker/controller/Dockerfile
```diff
@@ -14,4 +14,9 @@
 
 FROM busybox:glibc
 COPY snapshot-controller /bin/snapshot-controller
+# Root CA certificates
+COPY /etc/pki/ca-trust /etc/pki/ca-trust
```
rootfs: this Dockerfile seems to work for me locally
```
FROM  busybox:glibc
CMD rpm -i https://www.rpmfind.net/linux/fedora/linux/development/rawhide/Everything/x86_64/os/Packages/c/ca-certificates-2017.2.20-6.fc28.noarch.rpm
CMD EXTRACT_UNSAFE_SYMLINKS=1 rpm -i ca-certificates-2017.2.20-6.fc28.noarch.rpm
```

tsmetana: I'm not sure I want to rely on rpmfind.net to get the package from: I don't know what would happen to the link once the package is obsoleted. Also -- if I have rpm I can be reasonably sure I have the CA certificates on the system too.

rootfs: @tsmetana updated with fedora pkg.

tsmetana:
OK. I have capitulated: Adding the CA certificates to the repository and using the bundled copies to build the images. PTAL.

snapshot/deploy/docker/controller/Dockerfile
rootfs: need to update /etc/ :)

tsmetana: Eh... How did I... nm. Will fix this, thanks.

tsmetana:
Updated Dockerfiles to remove at least one layer. The CA files are stored in-tree, so no assumptions about the build system. Once merged I will push new Quay images, which should hopefully fix issue #558.

rootfs:
/lgtm

cc @childsb for merge

PR #588, opened 2018-02-02
Expose `shouldProvision` to provisioners via new controller interface

aka-bo:
Added new `controller.Qualifier` interface which is implemented by
any controller that has a `ShouldProvision` method. When implemented,
the provisioner's `ShouldProvision` method is called by lib controller's
`shouldProvision` with a return value of false from the former
resulting in the same from the latter.

This allows provisioners the opportunity to ignore individual claims
prior to attaining leadership for the claim, avoiding uneccessary
election cycles when multiple provisioners are serving the same storage
class.

### DIFF
--- a/lib/controller/controller.go
+++ b/lib/controller/controller.go
@@ -569,6 +569,12 @@ func (ctrl *ProvisionController) shouldProvision(claim *v1.PersistentVolumeClaim
 		return false
 	}
 
+	if qualifier, ok := ctrl.provisioner.(Qualifier); ok {
+		if !qualifier.ShouldProvision(claim) {
+			return false
+		}
+	}
+
 	// Kubernetes 1.5 provisioning with annStorageProvisioner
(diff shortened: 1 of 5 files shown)


jsafrane:
> This allows provisioners the opportunity to ignore individual claims prior to attaining leadership for the claim

Do you have a use case where this is useful? So far all instances of one provisioner were equal and each of them could provision all PVCs.

aka-bo:
So far the most applicable use cases we've encountered are centered around capacity. While all instances of a provisioner start out on equal footing, over time each instance has the potential to exhaust the capacity of its backing storage medium at different rates, e.g., available disk space or--in the case of NFS--max exports. When this happens today an instance will still attempt to gain leadership for a PVC even when it lacks the ability to provision, leading to significant delays in provisioning.

jsafrane:
Ok, that sound useful.
/lgtm

@childsb, please merge.

PR #590, opened 2018-02-06
Fixes issues #577

nseyvet:
Add rbac as an option in the Values file, add templates for generating the service account and role bindings.

Pull request based on https://github.com/kubernetes-incubator/external-storage/pull/579.

### DIFF
--- a/local-volume/README.md
+++ b/local-volume/README.md
@@ -139,6 +139,7 @@ deployment spec and skip to [step 4](#step-4-create-local-persistent-volume-clai
 unless you want to customize the provisioner spec or storage classes.
 
 ##### Using local-ssd-count option
+
 ``` console
 $ gcloud container cluster create ... --local-ssd-count=<n> --enable-kubernetes-alpha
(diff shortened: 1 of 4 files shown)


local-volume/README.md
```diff
@@ -229,15 +227,17 @@ helm template ./helm/provisioner --set engine=gke > ./provisioner/deployment/kub
 ```
 Parameter **--set engine=** canbe used in conjunction with custom vlues.yaml file in the same command line.
 
-3. Deploy Provisioner 
+Note: By default, common.rbac is set to "true" which generates the necessary ServiceAccount, ClusterRole and ClusterRoleBinding for an RBAC enabled kubernetes cluster.
```
sbezverk: I think we should add a comment that, common.rbac can be overridden in the command line if creating those objects is not required. Maybe adding an example could be useful here too.

nseyvet:
Is not that the case for any/all parameters?

Not sure there is much value in describing it further.

nseyvet:
Added example in README.md.

nseyvet:
And squashed commits into 1.

PR #595, opened 2018-02-08
Fix error message when a plain file is discovered.

jsafrane:
Provisioner logs ugly `Block device check for "/mnt/local-storage/fast/vol3" failed: DirErr - <nil> BlkErr - <nil>` when it finds a plain file in scanned directory.

It should report something less confusing, e.g. `Skipping file "/mnt/local-storage/fast/vol3": not a directory nor block device`

@msau42 @dhirajh  @ianchakeres

### DIFF
--- a/local-volume/provisioner/pkg/discovery/discovery.go
+++ b/local-volume/provisioner/pkg/discovery/discovery.go
@@ -181,13 +181,21 @@ func (d *Discoverer) getVolumeType(fullPath string) (string, error) {
 	if isdir {
 		return common.VolumeTypeFile, nil
 	}
+	// check for Block before returning errdir
 	isblk, errblk := d.VolUtil.IsBlock(fullPath)
 	if isblk {
 		return common.VolumeTypeBlock, nil
 	}
 
-	return "", fmt.Errorf("Block device check for %q failed: DirErr - %v BlkErr - %v", fullPath, errdir, errblk)
+	if errdir == nil && errblk == nil {
+		return "", fmt.Errorf("Skipping file %q: not a directory nor block device", fullPath)
+	}
 
+	// report the first error found
+	if errdir != nil {
+		return "", fmt.Errorf("Directory check for %q failed: %s", fullPath, errdir)
+	}
+	return "", fmt.Errorf("Block device check for %q failed: %s", fullPath, errblk)
 }
 
 func generatePVName(file, node, class string) string {


local-volume/provisioner/pkg/discovery/discovery.go
```diff
@@ -181,12 +181,22 @@ func (d *Discoverer) getVolumeType(fullPath string) (string, error) {
...
+	// report the first error found
+	err := errdir
+	if err == nil {
+		err = errblk
+	}
+	if err != nil {
+		return "", fmt.Errorf("Error checking file %q: %s", fullPath, err)
+	}
+
+	return "", fmt.Errorf("Skipping file %q: not a directory nor block device", fullPath)
```
ianchakeres: nit: /nor/or/

ianchakeres: How about flipping the logic around...

if errdir ==nil && errblk == nil, then return "skipping"

and leaving the existing error message for both direrr and blkerr?

local-volume/provisioner/pkg/discovery/discovery.go
```diff
@@ -181,12 +181,22 @@ func (d *Discoverer) getVolumeType(fullPath string) (string, error) {
...
 	}
 
-	return "", fmt.Errorf("Block device check for %q failed: DirErr - %v BlkErr - %v", fullPath, errdir, errblk)
+	// report the first error found
+	err := errdir
+	if err == nil {
+		err = errblk
+	}
+	if err != nil {
+		return "", fmt.Errorf("Error checking file %q: %s", fullPath, err)
```
dhirajh: Can the log message here still refer to the call that generated the error message? Something like this:
if errdir != nil {
   return "", fmt.Errorf("Error checking isDir for path %q: %s", fullPath, err)
}
if errblk != nil {
   return "", fmt.Errorf("Error checking isBlock for path %q: %s", fullPath, err)
}
The extra detail would help with debugging.

msau42: If you have an errdir, is there still a possibility that it is a block device?  Can you just return immediately if you have an errdir?

dhirajh: For some quick tests I just did on CentOS 7 Linux and OS X, it does look like isDir() will not return err just because its a symbolic link to a device. I am trying to recall why I kept error returns independent. I think it might have been because isBlock() was returning errors where it could have been a dir.  In any case, its fine to return immediately after isDir() fails with an err.

jsafrane: `IsDir()` returns error only on serious error, like permission issues, file does not exist and such. IsBlock is very likely to return the same error.

jsafrane:
I reworked the PR as @ianchakeres suggested, PTAL

dhirajh:
LGTM

PR #601, opened 2018-02-13
flex.provision.sh improvements

sathieu:
This is a subset of #505.

@wongma7 please review.


wongma7:
I don't mind it being all one PR, thank you for the improvements and apologies for the delay

PR #603, opened 2018-02-13
flex: Use  gcr.io/google-containers/debian-base-amd64 as base image

sathieu:
Subset of #505.


sathieu:
@wongma7 please review.

PR #612, opened 2018-02-20
Add pvc namer interface

sanjid133:
For some providers, there have limitations on volume name. e.g. for `Linode`,  maximum length of the volume name is 32 characters (Length of default namer is 40 characters). So, this interface will help those providers to use own namer.

### DIFF
--- a/aws/efs/cmd/efs-provisioner/efs-provisioner.go
+++ b/aws/efs/cmd/efs-provisioner/efs-provisioner.go
@@ -297,6 +297,7 @@ func main() {
 		provisionerName,
 		efsProvisioner,
 		serverVersion.GitVersion,
+		controller.NewDefaultNamer(),
 	)
 
 	pc.Run(wait.NeverStop)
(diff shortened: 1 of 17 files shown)


wongma7:
Please follow the pattern here https://github.com/kubernetes-incubator/external-storage/pull/612/files#diff-16daa5c3f02eee06add7f1af0a3fa901L234 to make it an optional arg e.g. something like 
```go
func Namer(namer Namer) func(*ProvisionController) error {
 	return func(c *ProvisionController) error {
 		if c.HasRun() {
 			return errRuntime
 		}
 		c.namer = namer
 		return nil
 	}
}
```

this is just so not everyone has to add `controller.NewDefaultNamer(),` to their args should they update. Thanks & apologies for months-long delay before review

Otherwise the code lgtm

I am actually unsure about whether we should let authors name the pvc however they want or enforce a suffix <PVC UID>, I will need to check spec

PR #578, opened 2018-01-26
Upgrade kubernetes dependencies to 1.9.2

cofyc:
**What this PR does / why we need it**:

Upgrade kubernetes dependencies to 1.9.2. Some features requires 1.9.x API definitions. See https://github.com/kubernetes-incubator/external-storage/pull/552, https://github.com/kubernetes-incubator/external-storage/issues/570.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #

**Special notes for reviewer**:

I splited the PR into two parts:

- Vendor files: https://github.com/kubernetes-incubator/external-storage/pull/578/commits/92504da63ee667cdb1ae72935638fd30ce758015 (`glide up -v` followed by `glide-vc --use-lock-file`)
- Non-vendor changes: https://github.com/kubernetes-incubator/external-storage/pull/578/commits/c7537ccfe68998da90ff3929bcb9d57744b751e3

Due to some packages depending to packages under `k8s.io/kubernetes` which changed a lot, a lot of conflicts need to be fixed:

- Package `k8s.io/kubernetes/pkg/api/v1/helper` moved to `k8s.io/kubernets/pkg/apis/core/v1/helper`
- Remove Scheme.DeepCopy, use `<obj>.DeepCopy()` directly, see https://github.com/kubernetes/kubernetes/pull/53525.
- Regenerate deepcopy files for snapshot CRD types (also added some hack scripts to generate it automatically)

PR #617, opened 2018-02-22
Upgrade kubernetes dependencies to v1.10.0-beta.1

NickrenREN:
Upgrade kubernetes dependencies to v1.10.0-beta.1

first part of https://github.com/kubernetes-incubator/external-storage/issues/638

/assign @msau42


NickrenREN:
@msau42  should we support both alpha and beta NodeAffinity ?

msau42:
Yes, I think we should have a flag to indicate if you want to use alpha or beta api.  Default can be beta.

NickrenREN:
ok, got it. will update

msau42:
I think to update vendor, glide files also need to be updated.  See https://github.com/kubernetes-incubator/external-storage/pull/578.  If it gets too big, maybe it will be better to do vendor changes as a separate pr

NickrenREN:
yeah, wait for a new cut of kubernetes tags as well as api repo which includes your commit, and then i will update this PR

local-volume/provisioner/pkg/common/common.go
```diff
@@ -149,8 +149,7 @@ func CreateLocalPVSpec(config *LocalPVConfig) *v1.PersistentVolume {
 			Name:   config.Name,
 			Labels: config.Labels,
 			Annotations: map[string]string{
-				AnnProvisionedBy:                      config.ProvisionerName,
-				v1.AlphaStorageNodeAffinityAnnotation: config.AffinityAnn,
```
msau42: I think instead of completely removing, it should be controlled by a configuration flag, with beta as the default.

NickrenREN: yes, i see. 
and as you suggested, i splited this task into several sub tasks,  this PR is just to upgrade kubernetes dependencies

NickrenREN: see: https://github.com/kubernetes-incubator/external-storage/issues/638

msau42:
@rootfs has a PR out to do the API update: #631

Issue #608, opened 2018-02-19
New Release for RBD


NicolaiSchmid:
Hey,
could we build and push a new release for Ceph RBD? 
I'm on Kubernetes 1.9.2 and I'd like to use it for provisioning my PVCs.
Also, are the images still pushed to quay.io/external_storage?
Cheers,
Nicolai

rootfs:
@childsb @wongma7 can you make a new release image?

AnatolyRugalev:
I really stuck because of this issue, so I decided to make temporary image in docker hub to use it in my cluster.

https://hub.docker.com/r/anatolyrugalev/rbd-provisioner/

Hope it helps.

NicolaiSchmid:
Thank you!
I'm currently using your image, but I still get an error mounting my PV to a pod:
```
Unable to mount volumes for pod "...": 
timeout expired waiting for volumes to attach/mount for pod "...".
list of unattached/unmounted volumes=[volume]
```
Do you think this is related to the provisioner or some other problem? I have absolutely no idea since the PVC gets satisfied, an image is created and tried to mount. Inside the RBD container, it can connect to Ceph and provision the image without any challenges. There are no error logs inside the provisioner after the timeout and I can't find any other logs regarding this problem.
Any suggestions are really appreciated!

timn:
I made the experience yesterday that the ClusterRole does not give all the permissions necessary. In particular, the rbd-provisioner needs to resolve the kube-dns service to pass on the IP of the ceph monitor. If that fails it seems to not do some important steps afterwards (check the output of rbd-provisioner, look for a permission denial). You could see this since the volume does not create in the rbd pool at all.

I fixed this by adding the following at the end of https://github.com/kubernetes-incubator/external-storage/blob/master/ceph/rbd/deploy/rbac/clusterrole.yaml (didn't get to file a PR, yet):
```
  - apiGroups: [""]
    resources: ["services"]
    verbs: ["list", "get"]
```
Maybe that helps.

I'm also still looking into another problem, where the volume is created, but mounting on the pod host fails. Still looking into this one.

timn:
Pushed #620 with more explicit restrictions now. @NicolaiSchmid, you could close your PR.

klynch:
Any way we could get a release cut?

timn:
@rootfs, anything you can do to get a release pushed?
[...]

PR #619, opened 2018-02-23
RBD image connection issue in Kubernetes v1.9.x

NicolaiSchmid:
This commit fixes the timeout, binding a new image from CEPH to a pod.
The RDB provisioner needs access to the service resource in order to connect the RBD image to the pod. Those permissions are granted with this additional rule
```
Unable to mount volumes for pod "...": 
timeout expired waiting for volumes to attach/mount for pod "...".
list of unattached/unmounted volumes=[volume]
```
based on the work of @timn in #608


timn:
This can actually be further constrained. I have added a PR doing so in #620.

NicolaiSchmid:
Closed in favor of #620

PR #620, opened 2018-02-23
rbd: add access to kube-dns service

timn:
The rbd-privisioner gets the kube-dns service to pass on the numeric IP
to the provisioning tools in order to avoid outside name resolution. On
a cluster with RBAC this requires appropriate permissions. Add those to
the cluster role (role does not suffice if provisioner is put into other
namespace than kube-system, for example ceph).


timn:
CLA signed.

rootfs:
/lgtm

cc @dmick

dmick:
I'll take your word for it; I'm no permissions guy

dmick:
(and thanks for finding/fixing this @timn)

PR #633, opened 2018-03-01
- updated version from v1alpha1 to v1 , else it breaks atleast on k8s…

initcron:
updated version for clusterrole and clusterrolebinding from v1alpha1 to v1. It broken on k8s 1.9.3 at least. 

added documentation to update namespace. does not work in custom namespaces otherwise


wongma7:
Thank you, you are right alpha is not around any more. Please sign the cla and I will merge

PR #635, opened 2018-03-02
Add binding error message for volumeMode:Block unsupported case

mkimuram:
This PR adds a binding error message for volumeMode:Block
unsupported case.
With this message, users can understand why PV is not bound to
PVC on dynamic provisioning.

fixes: #59942

Special notes for your reviewer:
/cc @mtanino
/sig storage

### DIFF
--- a/lib/controller/controller.go
+++ b/lib/controller/controller.go
@@ -950,9 +950,27 @@ func (ctrl *ProvisionController) provisionClaimOperation(claim *v1.PersistentVol
 			ctrl.eventRecorder.Event(claim, v1.EventTypeWarning, "ProvisioningCleanupFailed", strerr)
 		}
 	} else {
-		glog.Infof("volume %q provisioned for claim %q", volume.Name, claimToClaimKey(claim))
-		msg := fmt.Sprintf("Successfully provisioned volume %s", volume.Name)
-		ctrl.eventRecorder.Event(claim, v1.EventTypeNormal, "ProvisioningSucceeded", msg)
+		expectedVolumeMode := claim.Spec.VolumeMode
+		provisionedVolumeMode := volume.Spec.VolumeMode
+		// Report error if volumeMode for claim is different from volumeMode for provisioned volume,
+		// except if default volumeMode(Filesystem) is expected for claim and volumeMode is not set for provisioned volume.
+		if expectedVolumeMode != provisionedVolumeMode &&
+			!(expectedVolumeMode != nil && *expectedVolumeMode == v1.PersistentVolumeFilesystem && provisionedVolumeMode == nil) {
+			var expected, provisioned v1.PersistentVolumeMode
+			if expectedVolumeMode != nil {
+				expected = *expectedVolumeMode
+			}
+			if provisionedVolumeMode != nil {
+				provisioned = *provisionedVolumeMode
+			}
+			glog.Infof("volume %q is provisioned, but not bound to claim %q", volume.Name, claimToClaimKey(claim))
+			strerr := fmt.Sprintf("Volume %q is provisioned, but won't be bound to claim %q due to volumeMode mismatch(expected: %v, provisioned: %v)", volume.Name, claimToClaimKey(claim), expected, provisioned)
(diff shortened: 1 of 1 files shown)


lib/controller/controller.go
```diff
@@ -950,9 +950,27 @@ func (ctrl *ProvisionController) provisionClaimOperation(claim *v1.PersistentVol
...
-		glog.Infof("volume %q provisioned for claim %q", volume.Name, claimToClaimKey(claim))
-		msg := fmt.Sprintf("Successfully provisioned volume %s", volume.Name)
-		ctrl.eventRecorder.Event(claim, v1.EventTypeNormal, "ProvisioningSucceeded", msg)
+		expectedVolumeMode := claim.Spec.VolumeMode
+		provisionedVolumeMode := volume.Spec.VolumeMode
+		// Report error if volumeMode for claim is different from volumeMode for provisioned volume,
+		// except if default volumeMode(Filesystem) is expected for claim and volumeMode is not set for provisioned volume.
+		if expectedVolumeMode != provisionedVolumeMode &&
+			!(expectedVolumeMode != nil && *expectedVolumeMode == v1.PersistentVolumeFilesystem && provisionedVolumeMode == nil) {
+			var expected, provisioned v1.PersistentVolumeMode
```
msau42: Can you also set the string to filesystem when the mode is nil?

lib/controller/controller.go
```diff
@@ -950,9 +950,27 @@ func (ctrl *ProvisionController) provisionClaimOperation(claim *v1.PersistentVol
...
+			if expectedVolumeMode != nil {
+				expected = *expectedVolumeMode
+			}
+			if provisionedVolumeMode != nil {
+				provisioned = *provisionedVolumeMode
+			}
+			glog.Infof("volume %q is provisioned, but not bound to claim %q", volume.Name, claimToClaimKey(claim))
+			strerr := fmt.Sprintf("Volume %q is provisioned, but won't be bound to claim %q due to volumeMode mismatch(expected: %v, provisioned: %v)", volume.Name, claimToClaimKey(claim), expected, provisioned)
+			ctrl.eventRecorder.Event(claim, v1.EventTypeWarning, "ProvisioningFailed", strerr)
```
msau42: Should the PV be deleted?  Will the external provisioner keep trying over and over?

lib/controller/controller.go
mkimuram: O.K. I will change it and submit the code later.

lib/controller/controller.go
mkimuram: > Will the external provisioner keep trying over and over?
For as long as I see in my testing, no more volume is provisioned after one volume is provisioned.

> Should the PV be deleted? 
Ideally yes, I think. 

Let me check how it work after deleting the provisioned PV here and I will implement it and report you back.
My concerns currently are below two points:
1. Whether provision of the volume happen again after deleting the PV here, as you pointed out.
2. Both internal and external provisioner can be work in the same way.
> Will the external provisioner keep trying over and over?
For as long as I see in my testing, no more volume is provisioned after one volume is provisioned.

> Should the PV be deleted? 
Ideally yes, I think. 

Let me check how it work after deleting the provisioned PV here and I will report you back.
My concerns currently are below two points:
1. Whether provision of the volume happen again after deleting the PV here, as you pointed out.
2. Both internal and external provisioner can be work in the same way.

mkimuram: As a result of some tests, I found that a new volume continues to be provisioned if the old volume is deleted, and it continues over and over. So, it would be better not to delete it, until some check mechanism is ready in provisioner to skip provisioning. (Please see the URL below for detail.)
Also, I found that it would be better to put everything in k8s side of the code, not external provisioner side. Because, if we put the fix to external provisioner side, this won't be enabled until provisioner is updated. As a result, old provisioner without this code continues to have the same problem, even when it is used with the latest version of the k8s.

Please see the k8s side of the fix below for the idea of the fix in detail. It will solve both internal provisioner case and external provisioner case.
https://github.com/kubernetes/kubernetes/pull/59947

wongma7:
@mkimuram do we still need this, saw the kube fix has merged?

PR #637, opened 2018-03-03
Cleanup local provisioner example specs

### LINKED ISSUE #607: [local-volume] Need example configmap
Now that configmap is required, the daemonset spec needs to be updated to also include a configmap example.

/area local-volume

msau42:
* Merge admin_account yaml into generated yamls for gce
* Move example spec into examples directory
* Add configmap example

Fixes #607


msau42:
cc @davidz627

davidz627:
Should the `admin_account.yaml` be deleted or is it still used elsewhere?

msau42:
It's still there as a generic example.  I will create a new directory to try to avoid confusion.

msau42:
Also added an example configmap to fix #607

msau42:
Spec should probably use a specific version instead of latest.

local-volume/README.md
```diff
@@ -80,15 +80,13 @@ unless you want to customize the provisioner spec or storage classes.
 
 ``` console
 $ NODE_LOCAL_SSDS=<n> cluster/kube-up.sh
-$ kubectl create -f provisioner/deployment/kubernetes/admin_account.yaml
```
rootfs: do we still need to keep `admin_account.yaml`? The file is just relocated.

msau42: I moved it into a generic "examples" directory.  It's just there for reference if you don't use the helm chart to generate all the specs together.  I guess I could remove everything all together and put a sample spec using helm chart defaults to avoid confusion

msau42:
Updated to just use the default helm generated spec

local-volume/provisioner/deployment/kubernetes/example/default_example_provisioner_generated.yaml
```diff
@@ -0,0 +1,98 @@
...
+    matchLabels:
+      app: local-volume-provisioner 
+  template:
+    metadata:
+      labels:
+        app: local-volume-provisioner
+    spec:
+      serviceAccountName: local-storage-admin
+      containers:
+        - image: "quay.io/external_storage/local-volume-provisioner:latest"
```
ianchakeres: Should we change the tag here to a specific version (e.g. v2.0.0) also?

Issue #641, opened 2018-03-05
Snapshot controller can silently fail to start if cache sync takes time


disrani-px:
There is a 1 second wait in the snapshot controller startup sequence for the cache to sync. It can sometimes take longer to sync the cache causing the startup sequence to silently return midway.

/area snapshot

disrani-px:
Fix was merged, closing.

PR #642, opened 2018-03-05
Fix cache sync for snapshot controller

disrani-px:
The previous 1 second wait was too low causing the controller to
sometimes silently return midway during startup

Issue #641


disrani-px:
@rootfs  Can you cc someone who can merge this change?

Issue #646, opened 2018-03-06
Snapshot controller: Race in delete/create


disrani-px:
The cache uses the name of the snapshot as the key for the maps when maintaining the actual and desired states. This leads to a race when objects are deleted and then created quickly especially when the delete fails initially.

Observed behavior:
1. Delete a snapshot. volumesnapshot object is deleted.
2. Delete from the plugin fails for some reason and is being retried continuously with a backoff
3. Create a snapshot with the same name. volumesnapshot object is created.
4. The retry of the delete operations to the plugin stops at this point
5. The create of the new snapshot is never picked up by the controller

From the code it looks like using the snapshot name as the key is trampling the maps in the cache. Other controllers in kubernetes seem to be using a UID for the key, which will be unique and avoid the race.

/area snapshot

disrani-px:
Fixed with #647

PR #647, opened 2018-03-06
Use unique name for snapshot to avoid race

disrani-px:
Using just the snapshot name can cause a race when objects are being deleted
and created too quickly.

Also cleaned up some code and fixed error return from FindSnapshot() for
the drivers. FindSnapshot() was not being called previously because
GetNameAndNameSpaceFromSnapshotName() was called with the wrong name.

Issue #646

### DIFF
--- a/snapshot/pkg/controller/cache/actual_state_of_world.go
+++ b/snapshot/pkg/controller/cache/actual_state_of_world.go
@@ -70,7 +70,7 @@ func (asw *actualStateOfWorld) AddSnapshot(snapshot *crdv1.VolumeSnapshot) error
 	asw.Lock()
 	defer asw.Unlock()
 
-	snapshotName := MakeSnapshotName(snapshot.Metadata.Namespace, snapshot.Metadata.Name)
+	snapshotName := MakeSnapshotName(snapshot)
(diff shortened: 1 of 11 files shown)


snapshot/pkg/controller/cache/actual_state_of_world.go
```diff
@@ -70,7 +70,7 @@ func (asw *actualStateOfWorld) AddSnapshot(snapshot *crdv1.VolumeSnapshot) error
 	asw.Lock()
 	defer asw.Unlock()
 
-	snapshotName := MakeSnapshotName(snapshot.Metadata.Namespace, snapshot.Metadata.Name)
+	snapshotName := MakeSnapshotName(snapshot)
```
rootfs: this could add the same snapshot multiple times (since they have different UIDs)

disrani-px: If they have different UID they would be different snapshots (possible if the old is being deleted and new one being created?).  There can still only be one snapshot object for a given name+namespace in etcd

disrani-px:
Ping @tsmetana

disrani-px:
@tsmetana PTAL. Thanks!

disrani-px:
@wongma7 @childsb Can this be merged?

Issue #640, opened 2018-03-04
SnapshotCreate should pass in annotations from the spec


disrani-px:
Currently, the SnapshotCreate API only passes in the PV spec and some hardcoded tags. It would be helpful to pass in annotations from the Snapshot Spec (or even the entire spec) to the plugin as a way to pass in options to the plugin when taking snapshots.

/area snapshot

disrani-px:
I took a stab at this and sent out a PR

disrani-px:
Fixed with #650

PR #650, opened 2018-03-08
Pass in Snapshot spec to Create API

disrani-px:
Issue #640

### DIFF
--- a/snapshot/pkg/controller/snapshotter/snapshotter.go
+++ b/snapshot/pkg/controller/snapshotter/snapshotter.go
@@ -252,7 +252,11 @@ func (vs *volumeSnapshotter) waitForSnapshot(uniqueSnapshotName string, snapshot
 
 // This is the function responsible for determining the correct volume plugin to use,
 // asking it to make a snapshot and assigning it some name that it returns to the caller.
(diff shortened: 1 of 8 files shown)


rootfs:
do you have a use case for passing the entire snapshot spec?

/assign @tsmetana

disrani-px:
@rootfs Yes. Without this, there is no way to pass in additional options to the snapshot plugin. For example, in Portworx we can take different types of snaps, local snaps and cloud snaps. We can choose which one to take by passing in annotations in the spec.
Other drivers can also use this to pass in options similar to PVC specs.

disrani-px:
@tsmetana PTAL. Thanks!

tsmetana:
I understand the point. I would prefer to extend the existing mechanism: we already pass tags to the plugin. So far it's only the metadata created/updated during the snapshot creation. We may append the created tags with existing ones (created by the user in the VolumeSnapshot) and pass them all using the existing interface. Would that work for your use case?

disrani-px:
@tsmetana Sure that would work for this case but in the future if there are other fields/options added to the snapshot spec that need to be passed in to the plugins it'll create the same problem again.

tsmetana:
All right then.
/lgtm

disrani-px:
@wongma7 Can you please merge this change. Thanks!

harsh-px:
@wongma7 Can we get this merged?

disrani-px:
@rootfs @tsmetana Can someone else help merge this change?

disrani-px:
@childsb Can you please merge this change? Thanks!

PR #653, opened 2018-03-09
nfs-client: rbac no longer alpha, add serviceAccount to deployment

cdrage:
RBCA is no longer alpha and thus `v1` is only needed.

Service account has been added to deployment.yaml in order to get the
example to work correctly.


cdrage:
@wongma7 @musse

wongma7:
/lgtm
thanks :+1:

Issue #638, opened 2018-03-03
Using beta api: promote NodeAffinity


NickrenREN:
Since we promoted volume scheduling and local storage to beta in kubernetes core repo, see: https://github.com/kubernetes/kubernetes/pull/59391
Here we need to update local storage provisioner to use beta api;

tasks:

- [x] upgrade kubernetes dependencies (PR: https://github.com/kubernetes-incubator/external-storage/pull/631)
- [x] promote NodeAffinity (both alpha and beta, specified by configMap) (PR: https://github.com/kubernetes-incubator/external-storage/pull/654)
- [ ] one time controller or script (issue:https://github.com/kubernetes-incubator/external-storage/issues/618, PR: #695)

/area local-volume
/assign @NickrenREN
cc @msau42

PR #654, opened 2018-03-09
Promote NodeAffinity to beta

NickrenREN:
Promote NodeAffinity to beta

second part of #638

### DIFF
--- a/local-volume/provisioner/pkg/common/common.go
+++ b/local-volume/provisioner/pkg/common/common.go
@@ -268,8 +268,8 @@ func insertSpaces(original string) string {
 
 // LoadProvisionerConfigs loads all configuration into a string and unmarshal it into ProvisionerConfiguration struct.
 // The configuration is stored in the configmap which is mounted as a volume.
-func LoadProvisionerConfigs(provisionerConfig *ProvisionerConfiguration) error {
-	files, err := ioutil.ReadDir(ProvisionerConfigPath)
+func LoadProvisionerConfigs(configPath string, provisionerConfig *ProvisionerConfiguration) error {
+	files, err := ioutil.ReadDir(configPath)
 	if err != nil {
 		return err
 	}
@@ -277,9 +277,9 @@ func LoadProvisionerConfigs(provisionerConfig *ProvisionerConfiguration) error {
 	for _, file := range files {
 		if !file.IsDir() {
 			if strings.Compare(file.Name(), "..data") != 0 {
-				fileContents, err := ioutil.ReadFile(path.Join(ProvisionerConfigPath, file.Name()))
+				fileContents, err := ioutil.ReadFile(path.Join(configPath, file.Name()))
 				if err != nil {
-					glog.Infof("Could not read file: %s due to: %v", path.Join(ProvisionerConfigPath, file.Name()), err)
+					glog.Infof("Could not read file: %s due to: %v", path.Join(configPath, file.Name()), err)
 					return err
 				}
 				data[file.Name()] = string(fileContents)
--- a/local-volume/provisioner/pkg/common/common.go
+++ b/local-volume/provisioner/pkg/common/common.go
@@ -59,6 +59,8 @@ const (
 	ProvisonerStorageClassConfig = "storageClassMap"
 	// ProvisionerNodeLabelsForPV contains a list of node labels to be copied to the PVs created by the provisioner
 	ProvisionerNodeLabelsForPV = "nodeLabelsForPV"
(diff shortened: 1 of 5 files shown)


local-volume/provisioner/pkg/discovery/discovery_test.go
```diff
@@ -331,17 +331,26 @@ func findSCName(t *testing.T, targetDir string, test *testConfig) string {
 }
```
ianchakeres: @NickrenREN - can we add a test or tests for UseAlphaAPI?

local-volume/provisioner/pkg/common/common.go
```diff
@@ -199,12 +211,17 @@ func VolumeConfigToConfigMapData(config *ProvisionerConfiguration) (map[string]s
...
-		nodeLabels, err := yaml.Marshal(config.NodeLabelsForPV)
-		if err != nil {
-			return nil, fmt.Errorf("unable to Marshal node label: %v", err)
+		nodeLabels, nlErr := yaml.Marshal(config.NodeLabelsForPV)
+		if nlErr != nil {
+			return nil, fmt.Errorf("unable to Marshal node label: %v", nlErr)
 		}
 		configMapData[ProvisionerNodeLabelsForPV] = string(nodeLabels)
 	}
+	ver, err := yaml.Marshal(config.UseAlphaAPI)
```
msau42: What happens of configmap is missing the entry? Will it default to false? Can you add a unit test for this?

local-volume/provisioner/pkg/discovery/discovery_test.go
NickrenREN: yes, will do

NickrenREN: done

local-volume/provisioner/pkg/common/common.go
NickrenREN: Wanted to add unit test for `LoadProvisionerConfigs` to test the default value of `UseAlphaAPI` ( which will not execute `Marshal` functions)
But when i made dir for `/etc/provisioner/config/` which is the `ProvisionerConfigPath`, `permission denied` will occur, any thoughts ?

test like this:
```
func TestLoadProvisionerConfigs(t *testing.T) {
       if err := os.MkdirAll("/etc/provisioner/config/", 0777); err != nil {
               t.Fatal("create config dir error: ", err)
       }
       provisionerConfig := &ProvisionerConfiguration{
               StorageClassConfig: make(map[string]MountConfig),
       }
       LoadProvisionerConfigs(provisionerConfig)

       if provisionerConfig.UseAlphaAPI == true {
               t.Errorf("UseAlphaAPI should default to false")
       }
}
```

msau42: You could add provisioner config as a variable to the function.  Then in the unit test, it could be a tmpdir, and in production it could be /etc/provisioner/config.

NickrenREN: done, PTAL

local-volume/provisioner/pkg/common/common_test.go
```diff
@@ -69,3 +69,24 @@ func TestSetupClientByInCluster(t *testing.T) {
 		t.Errorf("InClusterConfig not invoked")
 	}
 }
+
+func TestLoadProvisionerConfigs(t *testing.T) {
+	tmpConfigPath := "/tmp/provisioner/config"
+	if err := os.MkdirAll(tmpConfigPath, 0777); err != nil {
```
msau42: There's a TempDir function you can use that will randomly generate a unique directory name, so you won't conflict with other tests.

NickrenREN: updated, PTAL, thanks

local-volume/provisioner/pkg/common/common_test.go
```diff
@@ -69,3 +70,27 @@ func TestSetupClientByInCluster(t *testing.T) {
 		t.Errorf("InClusterConfig not invoked")
 	}
 }
+
+func TestLoadProvisionerConfigs(t *testing.T) {
+	tmpConfigPath, err := ioutil.TempDir("", "local-provisioner-config")
+	if err != nil {
+		t.Fatalf("create temp dir error: %v", err)
+	}
+	if err = os.MkdirAll(tmpConfigPath, 0777); err != nil {
```
msau42: I think this line is not needed

NickrenREN: sorry for this. removed

msau42:
ping @childsb

Issue #345, opened 2017-09-08
CephFS Provisioner Input/Output Error


tmysl:
We are having issues with the cephfs provisioner against a few of our kube clusters: 

The Provisioner runs correctly and I am able to create a PVC and mount it in my container but when i perform any operations against the mount I am presented with an I/O error.

`kubectl exec -it test-pod touch /mnt/test --namespace=sre
touch: cannot touch '/mnt/test': Input/output error`

The behavior is visible outside the container as well, this is on a node with the cephfs pool mounted within the directories the provisioner created:

```[root@ceph-client001]# ls -la /data/cephfs/volumes/kubernetes/
ls: cannot access /data/cephfs/volumes/kubernetes/kubernetes-dynamic-pvc-3041dba2-9415-11e7-862f-169051d230b2: Input/output error
total 0
drwxr-xr-x 1 root root 1 Sep  7 21:40 .
drwxr-xr-x 1 root root 2 Sep  7 21:40 ..
d????????? ? ?    ?    ?            ? kubernetes-dynamic-pvc-3041dba2-9415-11e7-862f-169051d230b2


[root@ceph-client001 tmysliwiec]# ls -la /data/cephfs/volumes/kubernetes/kubernetes-dynamic-pvc-3041dba2-9415-11e7-862f-169051d230b2
ls: cannot access /data/cephfs/volumes/kubernetes/kubernetes-dynamic-pvc-3041dba2-9415-11e7-862f-169051d230b2: Input/output error
```

I'm able to perform operations agains the cephfs everywhere else, and the cluster has been healthy and running mounts on tons of other non-kubernetes nodes. Not sure what would cause this behavior. Let me know if theres anything i can do to resolve this, thanks in advance!

kairen:
I also got the same error, more details as below:
```sh
$ kubectl describe po/fs-mount-pod
...
Events:
  FirstSeen	LastSeen	Count	From			SubObjectPath	Type		Reason			Message
  ---------	--------	-----	----			-------------	--------	------			-------
  41s		41s		1	default-scheduler			Normal		Scheduled		Successfully assigned fs-mount-pod to node4
  41s		41s		1	kubelet, node4				Normal		SuccessfulMountVolume	MountVolume.SetUp succeeded for volume "default-token-txx5l"
[...]

PR #472, opened 2017-11-20
Implement a cutomized `_deauthorize` method to remove caps and client entities correctly.

cofyc:
**What this PR does / why we need it**:

See https://github.com/kubernetes-incubator/external-storage/pull/466#issuecomment-345128222.

~~This PR conflicts with https://github.com/kubernetes-incubator/external-storage/pull/466. I will rebase this after https://github.com/kubernetes-incubator/external-storage/pull/466 is merged.~~ (Rebased)

PR #674, opened 2018-03-16
CephFS Provisioner: Make ceph.dir.layout.pool_namespace optional.

cofyc:

Linux kernel cephfs only support Rados Pool Namespace in 4.8+. See
https://github.com/torvalds/linux/commit/779fe0fb8e1883d5c479ac6bd85fbd237deed1f7.

If security is not critical, we can configure cephfs provisioner not to enforce pool namespace, then we can use it with linux kernels before 4.8.

fixes https://github.com/kubernetes-incubator/external-storage/issues/345

cc @wongma7 @rootfs

### DIFF
--- a/ceph/cephfs/cephfs_provisioner/cephfs_provisioner.py
+++ b/ceph/cephfs/cephfs_provisioner/cephfs_provisioner.py
@@ -220,7 +220,7 @@ class CephFSNativeDriver(object):
         assert caps[0]['entity'] == client_entity
         return caps[0]
 
-    def _create_volume(self, volume_path, size=None, data_isolated=False, namespace_enforced=True):
+    def _create_volume(self, volume_path, size=None, namespace_enforced=True):
         """
         Linux kernel cephfs only support Rados Pool Namespace after 4.8:
         https://github.com/torvalds/linux/commit/779fe0fb8e1883d5c479ac6bd85fbd237deed1f7.
@@ -232,7 +232,7 @@ class CephFSNativeDriver(object):
         for an already-created volume, even if it is in use.
         :param volume_path: VolumePath instance
         :param size: In bytes, or None for no size limit
-        :param data_isolated: If true, create a separate OSD pool for this volume
+        :param namespace_enforced: If true, enforce security isolation, use seperate namespace for this volume.
         :return:
         """
         path = self.volume_client._get_path(volume_path)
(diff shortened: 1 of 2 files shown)


ceph/cephfs/cephfs_provisioner/cephfs_provisioner.py
```diff
@@ -199,13 +220,69 @@ def cap_update(orig, want, unwanted):
...
+        self.volume_client._mkdir_p(path)
+
+        if size is not None:
+            self.volume_client.fs.setxattr(path, 'ceph.quota.max_bytes', size.__str__(), 0)
+
+        # data_isolated means create a separate pool for this volume
+        if data_isolated:
+            pool_name = "{0}{1}".format(self.volume_client.POOL_PREFIX, volume_path.volume_id)
+            log.info("create_volume: {0}, create pool {1} as data_isolated =True.".format(volume_path, pool_name))
+            pool_id = self.volume_client._create_volume_pool(pool_name)
```
ictfox: Seems it's not suitable for each volume create one Ceph pool as it would increase the pgs on OSD. For performance recommmand, each Ceph OSD pgs should less than one number(ex. 300).

rootfs:
cc @jcsp

ceph/cephfs/cephfs_provisioner/cephfs_provisioner.py
cofyc: This func supports to isolate data (it's just copied from https://github.com/ceph/ceph/blob/v13.0.0/src/pybind/ceph_volume_client.py#L603 with some modifications for our needs), in this script `data_isolated` is always `False`.

jcsp:
I'm very concerned about how this code is accessing private methods from the Ceph python modules.  This will certainly break as soon as anyone makes any changes on the Ceph side, as there will be no attempt (let alone a guarantee) to preserve the behaviour of private methods.

The "data_isolated=True" stuff (which probably inspired the pool creation in this patch) is not suitable for general purpose systems -- it was only in there to enable certain carefully crafted configurations in OpenStack Manila, where the admin is aware that they can only create a few shares, because each share is getting a whole pool to itself.  Apologies that that isn't clearer in docs.

If you wanted to enable avoiding namespaces, then you could consider a patch for Ceph to add a flag like that to volume creation.  However, if you were still hoping to use data_isolated=True then this would probably not work out for you, as creating a pool per volume is unsustainable.  The only really clean solution for someone with this problem is to get the newer kernel.

cofyc:
@jcsp 

`data_isolated` parameter is not used in this script actually (it originates from https://github.com/ceph/ceph/blob/v13.0.0/src/pybind/ceph_volume_client.py#L623). Removed in https://github.com/kubernetes-incubator/external-storage/pull/674/commits/38ff01455d196006b1feae628eee55bb05bbddc8. cc @ictfox 

I admit this is a dirty hack to make namespace enforcement optional. Due to `CephFSVolumeClient` do not meet our needs anymore (see also https://github.com/kubernetes-incubator/external-storage/pull/472), how about refactoring `cephfs_provisioner.py` to use `cephfs`, `rados` python modules directly? Then we don't need to call `CephFSVolumeClient` methods (including private). That would be much flexible.

cc @rootfs

jcsp:
If you're implementing similar functionality to CephFSVolumeClient, please bear in mind that the ceph source is LGPL, so you can't copy anything into Apache licensed code.

I would still encourage you to make the changes you need to CephFSVolumeClient in the ceph tree, rather than reimplementing things.

rootfs:
@cofyc see http://tracker.ceph.com/issues/23695

cofyc:
@rootfs Thanks!

batrick:
@cofyc please review https://github.com/ceph/ceph/pull/21808 and let us know if that works for you.

cofyc:
Closed in favor of https://github.com/kubernetes-incubator/external-storage/pull/773.

PR #680, opened 2018-03-22
remove stale file

cwedgwood:
remove stale file


cwedgwood:
CLA done

wongma7:
thanks. The vendor directory is automatically generated so even though the file is worthless unfortunately we have no choice but to keep it around

Issue #691, opened 2018-03-28
Set quota size for dynamic provisioned cephfs volume


mlmhl:
Kubernetes v1.10 supports using ceph-fuse to mount cephfs volumes, see this [PR](https://github.com/kubernetes/kubernetes/pull/55866), so we can set the volume's quota to PVC's capacity to make it work.

PR #692, opened 2018-03-28
Set quota size for dynamic provisioned cephfs volume

mlmhl:
Kubernetes v1.10 supports using ceph-fuse to mount cephfs volumes, so we can set the volume's quota to PVC's capacity.

Fix: #691


wongma7:
what happens if it is not a fuse mount? `-s` is ignored?

wongma7:
there is a duplicate at https://github.com/kubernetes-incubator/external-storage/pull/770/files with more discussion let's go there.

PR #696, opened 2018-03-29
[cephfs] Added storage class parameters to specify a root path within the backing cephfs and, optionally, use deterministic directory and user names

### LINKED ISSUE #406: CephFS root path
I would like to share one cephfs with multiple k8s instances. Currently, it places all pv's under /volumes/kubernetes/xxx

Could we make '/volumes/kubernetes/' specified on the cephfs-provisioner cli to be overridable so we can make directories non overlapping?

This is kind of the opposite of issue #245

mcronce:
Think I got everything covered.  This is tested and working on my local (1.9.4) Kube with some changes to the RBAC stuff.  I'm going to submit those RBAC changes in a separate PR to keep changes compartmentalized.

By the way, I have no idea why that L is lower-case in `VOlUME_GROUP`, but I can change that too if desired :)

This fixes #406.

### DIFF
--- a/ceph/cephfs/cephfs-provisioner.go
+++ b/ceph/cephfs/cephfs-provisioner.go
@@ -112,7 +112,7 @@ func (p *cephFSProvisioner) Provision(options controller.VolumeOptions) (*v1.Per
 	if options.PVC.Spec.Selector != nil {
 		return nil, fmt.Errorf("claim Selector is not supported")
 	}
-	cluster, adminID, adminSecret, pvc_root, mon, err := p.parseParameters(options.Parameters)
+	cluster, adminID, adminSecret, pvcRoot, mon, err := p.parseParameters(options.Parameters)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


mcronce:
Note - with and without the changes in this PR, the RBAC changes in #698 along with a freshly built container image are necessary for the provisioner to work

mcronce:
I just added another commit to this, rather than putting it in its own PR, because it's pretty closely related code-wise and reasonably related functionality-wise.

It adds a "deterministicNames" option to the storage class; if set to false, it retains the current behavior; if set to true, it uses the PVC name rather than the PV name (the PV name being randomly generated) as the name of the directory and, prepended with "k8s.", the Ceph user.

Critically, the default is false (retain current behavior).

mcronce:
After giving it some more thought, yeah, namespaces in the directory tree and Ceph client name are kinda required to prevent naming collisions.  Went ahead and added them.

mcronce:
@rootfs Thoughts? :)

wongma7:
/lgtm
more options is always good :)
imo the share names should have been deterministic to begin with but okay

mcronce:
@wongma7 thanks :) I agree but figured rather than forcing a change making it optional would be preferable.  XKCD 1172 and all that.  If it were _my_ project I would've made it forced but it's not ;)

PR #699, opened 2018-03-30
Update to allow flex provisioner scripts to specify more volume options

KnicKnic:
The goal was to allow more range for script provisioners.
1. Different PersistentVolumeSources.

1. All options for flexvolume including the driver

1. Give more info to the Deleter

1. I deleted code that I didn't think was being used

I know that this is a breaking change. I want to use this as opening the conversation on how to accomplish some of these goals.


wongma7:
I agree the old iteration of this flex provisioner is very limited so I don't mind breaking changes. https://github.com/kubernetes-incubator/external-storage/pull/505 was just merged which had a similar goal of making it more usable so you'll have to rebase

PR #697, opened 2018-03-29
Various small cleanups to glusterfile

jarrpa:
This PR contains various cleanups to the glusterfile provisioner. These were mostly things I looked for as an exercise to familiarize myself with the project. Most of the changes are documentation or aesthetic/stylistic changes, though there are a few points where I tried to simplify and clarify bits of logic in the code.

PR #702, opened 2018-04-02
Change error message and make it consistent across file.

humblec:

### DIFF
--- a/gluster/file/cmd/glusterfile-provisioner/glusterfile-provisioner.go
+++ b/gluster/file/cmd/glusterfile-provisioner/glusterfile-provisioner.go
@@ -227,8 +227,8 @@ func (p *glusterfileProvisioner) CreateVolume(gid *int, config *provisionerConfi
 	volID = volume.Id
 	dynamicHostIps, err := getClusterNodes(cli, volume.Cluster)
 	if err != nil {
-		glog.Errorf("error [%v] when getting cluster nodes for volume %s", err, volume)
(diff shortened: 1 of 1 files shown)


humblec:
@childsb Please review. Thanks !

obnoxxx:
I feel this is kind of in response to PR #697.
I think we should first sort that PR out.
I am personally in favor of the `[...]` quoting.
IMHO it makes the error messages much more readable (for the human eye).

humblec:
Thanks for the review.
~~~
 `I am personally in favor of the [...] quoting.
IMHO it makes the error messages much more readable (for the human eye).
~~~
Thats *not* the practice in entire kube source tree, thats why it was removed in between and I mentioned it in review comments. I will remove `error` string and do other correction mentioned here.

obnoxxx:
> Thats not the practice in entire kube source tree, thats why it was removed in between and I mentioned it in review comments.

Hmm, we might still adopt it for our modules, since it makes log messages much more readable. (Easier to spot the relevant pieces.) That's what I was voting for.

>  I will remove error string and do other correction mentioned here in next iteration of the PR.

Thanks!

obnoxxx:
ping

wongma7:
closing for now, let me know if we still want this

obnoxxx:
I think we still want to make the log messages consistent. (There was just the discussion about the "how".) Meanwhile pr #697 has adapted to @humblec 's proposed removal of `[...]`, so I'll withdraw my comments about that.

The other small nitpicks about making the messages really consistent still remain.

@humblec do you want to update the PR?

Issue #710, opened 2018-04-04
Support Jobs based cleaning for block devices in local volume provisioner


dhirajh:
Currently we support cleaning of block devices using a child process that is forked from the local volume provisioner. This request is for optionally allowing a Jobs based execution of the same cleaning process. The advantages of this approach are as follows

1. Greater transparency as the admin can check on the status of these long running jobs simply by using kubectl commands. 
2. Admins can take corrective action on specific cleaning jobs (like deleting a hung job) without disturbing any other cleaning jobs or the provisioner, using just the kubectl commands.
3. We can ensure that only one instance of cleaning process is running for a given volume (something we could not completely ensure with the process exec mechanism as DaemonSet has "at least one" semantics). We can also make sure that if the PV is manually deleted by the admin, a new one will not be created while the job is running.

PR #711, opened 2018-04-04
Jobs based block volume cleaning

dhirajh:
Addresses  issue #710 

Currently we support cleaning of block devices using a child process that is forked from the local volume provisioner. This request is for optionally allowing a Jobs based execution of the same cleaning process. The advantages of this approach are as follows

Greater transparency as the admin can check on the status of these long running jobs simply by using kubectl commands.
Admins can take corrective action on specific cleaning jobs (like deleting a hung job) without disturbing any other cleaning jobs or the provisioner, using just the kubectl commands.
We can ensure that only one instance of cleaning process is running for a given volume (something we could not completely ensure with the process exec mechanism as DaemonSet has "at least one" semantics). We can also make sure that if the PV is manually deleted by the admin, a new one will not be created while the job is running.


local-volume/provisioner/pkg/deleter/deleter.go
```diff
@@ -86,23 +86,33 @@ func (d *Deleter) deletePV(pv *v1.PersistentVolume) error {
...
 	}
 
-	if d.ProcTable.IsRunning(pv.Name) {
-		// Run in progress, nothing to do,
+	if d.IsCleaningInProgress(pv.Name) {
 		return nil
 	}
 
-	err = d.ProcTable.MarkRunning(pv.Name)
+	if volMode == v1.PersistentVolumeBlock && d.RuntimeConfig.UseJobForCleaning {
```
lichuqiang: @msau42 To clarify, if later we decide to support both block and file path as input of volume path,
the `volumeMode` cannot be a basis to determine the path type here any more (I think the field is first designed for the final format of a volume)
So maybe call `Mounter.PathIsDevice` to check the type?

lichuqiang: oh, I just noticed your comments https://github.com/kubernetes/community/pull/1914#pullrequestreview-110627427, again I get confused, so you mean volumeMode in PVs represent the origin type of a volume, and volumeMode in PVCs represent the its purpose type, the two can be different, is it?

msau42: Yes, I think it would be nicer to be more explicit about the path type for cases like this.  Problem is current code only supports volumeMode in PVC and PV to be equal, so we would need to change that, but it was rejected in the initial design for the block feature.

local-volume/helm/provisioner/templates/provisioner-cluster-role-binding.yaml
```diff
@@ -37,3 +37,29 @@ roleRef:
   name: local-storage-provisioner-node-clusterrole
   apiGroup: rbac.authorization.k8s.io
 {{- end }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
```
msau42: I think we want a Role (permissions in a specific namespace), rather than a ClusterRole (permissions across all namespaces)

local-volume/provisioner/pkg/common/common.go
```diff
@@ -80,6 +85,12 @@ type UserConfig struct {
 	NodeLabelsForPV []string
 	// UseAlphaAPI shows if we need to use alpha API
 	UseAlphaAPI bool
+	// UseJobForCleaning indicates if Jobs should be spawned for cleaning block devices (as opposed to process),.
+	UseJobForCleaning bool
+	// Namespace of this Pod (optional)
+	Namespace string
```
msau42: Can the Job namespace be different from provisioner namespace?

local-volume/provisioner/pkg/deleter/deleter.go
msau42: Looks like the current state of the proposal will not have explicit volumeMode checks.  To workaround that, I think we need explicit modes in the MountConfig.   So we can check for those modes here.

local-volume/provisioner/pkg/deleter/jobcontroller.go
```diff
@@ -0,0 +1,354 @@
+/*
+Copyright 2017 The Kubernetes Authors.
```
msau42: Year

local-volume/provisioner/pkg/deleter/jobcontroller.go
```diff
@@ -0,0 +1,354 @@
...
+			// the job. In that case, a new cleaning job will eventually be launched and the whole cleaning cycle
+			// will be repeated. This is an acceptable inefficiency for this failure case, as making sure that
+			// the deletion of job and pv is done by the same processItem() call and in that fixed sequence,
+			// avoids a whole bunch of unwanted race conditions.
+			glog.Infof("Job %s has already been deleted.", key)
+			return nil
+		}
+
+		glog.Infof("Job %s has completed and been deleted. Deleting PV %s", key, pvName)
+		if err := c.APIUtil.DeletePV(pvName); err != nil {
```
msau42: Instead of deleting the PV in the job controller, can you signal that the job is done in the proc table, and have the deleter delete the API object?  Then the PV object deletion is handled the same regardless of if you use jobs or not.

local-volume/provisioner/pkg/deleter/jobcontroller.go
```diff
@@ -0,0 +1,354 @@
...
+		SecurityContext: &apiv1.SecurityContext{
+			Privileged: &priv,
+		},
+	}
+	var volumes []apiv1.Volume
+	if len(discoveryMap) > 0 {
+		jobContainer.VolumeMounts = make([]apiv1.VolumeMount, len(discoveryMap))
+		volumes = make([]apiv1.Volume, len(discoveryMap))
+		i := 0
+		for _, mount := range discoveryMap {
```
msau42: Does it need to mount everything?  Or can it only mount the single device that needs to be cleaned?

local-volume/provisioner/pkg/common/common.go
dhirajh: The "Namespace" field here is meant to be set via the downward API in the pod manifest to be the namespace of the provisioner. This is also what the helm templates set it to. The admin could go in and explicitly change the provisioner manifest to point to some other namespace if they really want to.

local-volume/provisioner/pkg/deleter/jobcontroller.go
dhirajh: Fixed.

local-volume/provisioner/pkg/deleter/deleter.go
dhirajh: OK. When this change is made, please make sure that the e2e tests in kubernetes are also updated. I believe they currently assume both block device and filesystem discovery folders are the same.

local-volume/provisioner/pkg/deleter/deleter.go
```diff
@@ -82,75 +98,110 @@ func (d *Deleter) deletePV(pv *v1.PersistentVolume) error {
...
 
-	err = d.ProcTable.MarkRunning(pv.Name)
+	// Check if cleaning was just completed.
+	state, err := d.CleanupStatus.RemoveStatus(pv.Name, runjob)
 	if err != nil {
 		return err
 	}
 
-	go d.asyncDeletePV(pv, volMode, mountPath, config)
+	if state == CSSucceeded {
```
msau42: Can a switch statement be used here instead?

msau42 (commented):
lgtm just a minor nit.

local-volume/provisioner/pkg/deleter/deleter.go
dhirajh: Yes, I was on the fence with my choice there. Fixed now.

dhirajh:
Just wanted to check and see if this could be merged @childsb . Thanks.

PR #716, opened 2018-04-06
Update CI Links

msau42:
Now that the feature is beta, the tests are run in normal CI suites.
/area local-volume

### DIFF
--- a/local-volume/README.md
+++ b/local-volume/README.md
@@ -269,9 +269,21 @@ go run hack/e2e.go -- -v --test --test_args="--ginkgo.focus=PersistentVolumes-lo
 ```
 
 ### View CI Results
-[GCE Alpha](https://k8s-testgrid.appspot.com/sig-storage#gce-alpha)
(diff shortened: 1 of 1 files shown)


local-volume/README.md
```diff
@@ -269,9 +269,14 @@ go run hack/e2e.go -- -v --test --test_args="--ginkgo.focus=PersistentVolumes-lo
 ```
 
 ### View CI Results
-[GCE Alpha](https://k8s-testgrid.appspot.com/sig-storage#gce-alpha)
-
-[GKE Alpha](https://k8s-testgrid.appspot.com/sig-storage#gke-alpha)
+[GCE](https://k8s-testgrid.appspot.com/sig-storage#gce&include-filter-by-regex=PersistentVolumes-local)
```
ianchakeres: @msau42 - all these results are showing up on a single line. https://github.com/msau42/external-storage/blob/32bdbceb6203a2192701194b3126be7da22fa21d/local-volume/README.md#view-ci-results

It might be better to add lines between them (like existing README.md), so that they are each on a separate line. 
https://github.com/kubernetes-incubator/external-storage/tree/master/local-volume#view-ci-results

PR #721, opened 2018-04-08
Expose prometheus metrics for local volume provisioner

cofyc:
Expose prometheus metrics for local volume provisioner. Then we can monitor its internal, actions running, finished, and latency, etc.

### DIFF
--- a/local-volume/provisioner/cmd/main.go
+++ b/local-volume/provisioner/cmd/main.go
@@ -18,18 +18,32 @@ package main
 
 import (
 	"flag"
+	"log"
+	"net/http"
 	"os"
 
 	"github.com/golang/glog"
 	"github.com/kubernetes-incubator/external-storage/local-volume/provisioner/pkg/common"
 	"github.com/kubernetes-incubator/external-storage/local-volume/provisioner/pkg/controller"
+	"github.com/kubernetes-incubator/external-storage/local-volume/provisioner/pkg/deleter"
+	"github.com/kubernetes-incubator/external-storage/local-volume/provisioner/pkg/metrics"
+	"github.com/kubernetes-incubator/external-storage/local-volume/provisioner/pkg/metrics/collectors"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
 
 	"k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 )
 
+var (
+	optListenAddress string
+	optMetricsPath   string
+)
+
 func main() {
+	flag.StringVar(&optListenAddress, "listen-address", ":8080", "address on which to expose metrics")
+	flag.StringVar(&optMetricsPath, "metrics-path", "/metrics", "path under which to expose metrics")
 	flag.Set("logtostderr", "true")
 	flag.Parse()
 
@@ -61,7 +75,8 @@ func main() {
 	node := getNode(client, nodeName)
 
 	glog.Info("Starting controller\n")
-	controller.StartLocalController(client, &common.UserConfig{
+	procTable := deleter.NewProcTable()
+	go controller.StartLocalController(client, procTable, &common.UserConfig{
 		Node:              node,
 		DiscoveryMap:      provisionerConfig.StorageClassConfig,
 		NodeLabelsForPV:   provisionerConfig.NodeLabelsForPV,
@@ -70,6 +85,21 @@ func main() {
 		Namespace:         namespace,
 		JobContainerImage: jobImage,
 	})
+
+	glog.Infof("Starting metrics server at %s\n", optListenAddress)
+	prometheus.MustRegister([]prometheus.Collector{
+		metrics.PersistentVolumeDiscoveryTotal,
+		metrics.PersistentVolumeDiscoveryDurationSeconds,
+		metrics.PersistentVolumeDeleteTotal,
+		metrics.PersistentVolumeDeleteDurationSeconds,
+		metrics.PersistentVolumeDeleteFailedTotal,
+		metrics.APIServerRequestsTotal,
+		metrics.APIServerRequestsFailedTotal,
+		metrics.APIServerRequestsDurationSeconds,
+		collectors.NewProcTableCollector(procTable),
+	}...)
+	http.Handle(optMetricsPath, promhttp.Handler())
+	log.Fatal(http.ListenAndServe(optListenAddress, nil))
 }
 
 func getNode(client *kubernetes.Clientset, name string) *v1.Node {
--- a/local-volume/provisioner/pkg/deleter/deleter.go
+++ b/local-volume/provisioner/pkg/deleter/deleter.go
@@ -23,12 +23,14 @@ import (
 	"os/exec"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/golang/glog"
+	"github.com/kubernetes-incubator/external-storage/local-volume/provisioner/pkg/common"
+	"github.com/kubernetes-incubator/external-storage/local-volume/provisioner/pkg/metrics"
+
 	"k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
-
-	"github.com/kubernetes-incubator/external-storage/local-volume/provisioner/pkg/common"
 )
 
 // CleanupState indicates the state of the cleanup process.
@@ -72,6 +74,12 @@ func (d *Deleter) DeletePVs() {
 			// Cleanup volume
 			err := d.deletePV(pv)
 			if err != nil {
+				mode, runjob := d.getVolModeAndRunJob(pv)
+				deleteType := metrics.DeleteTypeProcess
+				if runjob {
+					deleteType = metrics.DeleteTypeJob
+				}
+				metrics.PersistentVolumeDeleteFailedTotal.WithLabelValues(string(mode), deleteType).Inc()
 				cleaningLocalPVErr := fmt.Errorf("Error cleaning PV %q: %v", name, err.Error())
 				d.RuntimeConfig.Recorder.Eventf(pv, v1.EventTypeWarning, common.EventVolumeFailedDelete, cleaningLocalPVErr.Error())
 				glog.Error(err)
@@ -79,6 +87,17 @@ func (d *Deleter) DeletePVs() {
 			}
 		}
 	}
+}
+
+func (d *Deleter) getVolModeAndRunJob(pv *v1.PersistentVolume) (v1.PersistentVolumeMode, bool) {
+	// Default is filesystem mode, so even if volume mode is not specified, mode should be filesystem.
+	volMode := v1.PersistentVolumeFilesystem
+	runjob := false
+	if pv.Spec.VolumeMode != nil && *pv.Spec.VolumeMode == v1.PersistentVolumeBlock {
+		volMode = v1.PersistentVolumeBlock
+		runjob = d.RuntimeConfig.UseJobForCleaning
+	}
+	return volMode, runjob
 }
 
 func (d *Deleter) deletePV(pv *v1.PersistentVolume) error {
@@ -96,13 +115,7 @@ func (d *Deleter) deletePV(pv *v1.PersistentVolume) error {
 		return err
 	}
 
-	// Default is filesystem mode, so even if volume mode is not specified, mode should be filesystem.
-	volMode := v1.PersistentVolumeFilesystem
-	runjob := false
-	if pv.Spec.VolumeMode != nil && *pv.Spec.VolumeMode == v1.PersistentVolumeBlock {
-		volMode = v1.PersistentVolumeBlock
-		runjob = d.RuntimeConfig.UseJobForCleaning
-	}
+	volMode, runjob := d.getVolModeAndRunJob(pv)
 
 	// Exit if cleaning is still in progress.
 	if d.CleanupStatus.InProgress(pv.Name, runjob) {
@@ -110,7 +123,7 @@ func (d *Deleter) deletePV(pv *v1.PersistentVolume) error {
 	}
 
 	// Check if cleaning was just completed.
-	state, err := d.CleanupStatus.RemoveStatus(pv.Name, runjob)
+	state, startTime, err := d.CleanupStatus.RemoveStatus(pv.Name, runjob)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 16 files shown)


cofyc:
cc @msau42 @ianchakeres

ianchakeres:
@dhirajh - this PR touches stuff that is in your open PR - https://github.com/kubernetes-incubator/external-storage/pull/711. Please feel free to chime in.

ianchakeres:
FYI: @jpisaac

msau42:
Sorry this fell off my radar. Will try to take a look today

local-volume/provisioner/pkg/metrics/metrics.go
```diff
@@ -0,0 +1,58 @@
...
+	ProvisionCreate = "create"
+	// ProvisionDelete represents delete action.
+	ProvisionDelete = "delete"
+)
+
+var (
+	// PersistentVolumeProvisionTotal is used to collect accumulating count of persistent volumes provisioned.
+	PersistentVolumeProvisionTotal = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Subsystem: LocalVolumeProvisionerSubsystem,
```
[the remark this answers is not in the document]
cofyc: @dhirajh 
That would be much more useful! Done in https://github.com/kubernetes-incubator/external-storage/pull/721/commits/b815a11445ba67165fb3df568e0424604fb49c0e.

local-volume/provisioner/pkg/metrics/collectors/helper_test.go
```diff
@@ -0,0 +1,182 @@
...
+
+	for _, l := range lines {
+		trimmedLine = strings.TrimSpace(l)
+
+		if len(trimmedLine) > 0 {
+			trimmedLines = append(trimmedLines, trimmedLine)
+		}
+	}
+
+	// The Prometheus metrics representation parser expects an empty line at the
```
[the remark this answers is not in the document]
cofyc: @dhirajh 

`gatherAndCompare` and related code is from `kube-state-metrics`. I submitted a PR in kube-state-metrics to expose `gatherAndCompare`: https://github.com/kubernetes/kube-state-metrics/pull/438. After the PR is merged, I will vendor it. Code will be more concise.

local-volume/provisioner/pkg/metrics/metrics.go
```diff
@@ -27,6 +27,10 @@ const (
 	ProvisionCreate = "create"
 	// ProvisionDelete represents delete action of provision.
 	ProvisionDelete = "delete"
+	// APIServerRequestCreate represents create method of apiserver request.
```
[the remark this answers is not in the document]
cofyc: @dhirajh OK. I will update it.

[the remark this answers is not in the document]
cofyc: @dhirajh Done.

local-volume/provisioner/pkg/metrics/collectors/helper_test.go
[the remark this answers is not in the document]
cofyc: @dhirajh Done, see https://github.com/kubernetes-incubator/external-storage/pull/721/commits/9eaa822860da78cd462e930900b73a703c39e5c6.

local-volume/provisioner/cmd/main.go
```diff
@@ -50,12 +64,26 @@ func main() {
...
+		metrics.PersistentVolumeProvisionTotal,
+		metrics.PersistentVolumeProvisionLatency,
+		metrics.APIServerRequestsTotal,
+		metrics.APIServerRequestsFailedTotal,
+		metrics.APIServerRequestsLantency,
+		metrics.PersistentVolumeCleanupFailedTotal,
+		collectors.NewProcTableCollector(procTable),
+	}...)
+	http.Handle(optMetricsPath, promhttp.Handler())
+	log.Fatal(http.ListenAndServe(optListenAddress, nil))
```
dhirajh: Have you had any issues shutting down the container with this new code? Does the daemonset stop responsively? or does the kubelet sleep for 30 seconds before doing a hard kill of the container? Just want to make sure this listen still responds to SIGINT.  If you do find that its not exiting quickly, you might need to add code along these lines: https://medium.com/@kpbird/golang-gracefully-stop-application-23c2390bb212

dhirajh (commented):
Apart from these two final comments, it looks good to me.

local-volume/provisioner/cmd/main.go
cofyc: Doesn't have any issues shutdown down the container with this new code. Container exits quickly after pod is deleted.

local-volume/provisioner/pkg/deleter/proctable.go
```diff
@@ -51,36 +59,47 @@ func NewProcTable() *ProcTableImpl {
...
 
 // MarkRunning Indicate that process is running.
-func (v *ProcTableImpl) MarkRunning(pvName string) error {
+func (v *ProcTableImpl) MarkRunning(pvName string) (time.Time, error) {
 	v.mutex.Lock()
 	defer v.mutex.Unlock()
 	_, ok := v.procTable[pvName]
 	if ok {
-		return fmt.Errorf("Failed to mark running of %q as it is already running, should never happen", pvName)
+		return time.Time{}, fmt.Errorf("Failed to mark running of %q as it is already running, should never happen", pvName)
```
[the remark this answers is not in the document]
cofyc: Cannot return nil here because I use `time.Time` struct as return value, not a interface/pointer. I will add `GetProcEntry(pvName)` as you suggested.

local-volume/provisioner/pkg/deleter/proctable.go
```diff
@@ -51,36 +59,47 @@ func NewProcTable() *ProcTableImpl {
...
-	v.mutex.Lock()
-	defer v.mutex.Unlock()
+	v.mutex.RLock()
+	defer v.mutex.RUnlock()
 	return len(v.procTable) == 0
 }
 
 // MarkRunning Indicate that process is running.
-func (v *ProcTableImpl) MarkRunning(pvName string) error {
+func (v *ProcTableImpl) MarkRunning(pvName string) (time.Time, error) {
```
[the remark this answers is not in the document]
cofyc: 2. Done in https://github.com/kubernetes-incubator/external-storage/pull/721/commits/73124679a218279d3e683ac7ce0c26da313e868d, only a few lines are injected in [deleter.go](https://github.com/kubernetes-incubator/external-storage/pull/721/files#diff-c66875cfe02a496df242b6d171010a33) now.

dhirajh (approved):
LGTM cc @msau42

local-volume/provisioner/cmd/main.go
```diff
@@ -18,18 +18,32 @@ package main
...
 	"k8s.io/client-go/kubernetes"
 )
 
+var (
+	optListenAddress string
+	optMetricsPath   string
+)
+
 func main() {
+	flag.StringVar(&optListenAddress, "listen-address", ":8080", "address on which to expose metrics")
```
msau42: We've been putting all our configuration parameters into a configmap, instead of command line arguments. @dhirajh do you have any preference?

local-volume/provisioner/pkg/deleter/deleter.go
```diff
@@ -140,6 +145,12 @@ func (d *Deleter) asyncDeletePV(pv *v1.PersistentVolume, volMode v1.PersistentVo
 
 	glog.Infof("Deleted PV %q", pv.Name)
 
+	action := metrics.ProvisionDelete
+	metrics.PersistentVolumeProvisionTotal.WithLabelValues(action).Inc()
+	procEntry := d.ProcTable.GetProcEntry(pv.Name)
+	if procEntry != nil {
+		metrics.PersistentVolumeProvisionLatency.WithLabelValues(action).Observe(time.Since(procEntry.StartTime).Seconds())
```
msau42: For delete, would it be useful to also break it out by PV capacity and delete method?

msau42 (commented):
Can you also add a section into the README documenting these endpoints?

msau42:
Would it be useful to create an example Service spec, so that users can easily hook up the Endpoints into a monitoring pipeline?

local-volume/provisioner/pkg/deleter/deleter.go
cofyc: It should be useful.

VolumeMode determines delete method, we can use VolumeMode here like `PersistentVolumeCleanupFailedTotal`.

Values of label in metrics can not have high cardinality, but we can classify capacity into:

- 1M
- 10M
- 100M
- 1000M
- 10000M
- 100000M
- ...

what do you think?

cofyc:
I will add examples and update README.

local-volume/provisioner/pkg/deleter/deleter.go
msau42: Maybe breaking it down into every 500gb could work. We mostly care about larger capacities where cleanup can take a very long time.

For delete action, we actually have customizable cleanup routines, some of which can be very fast (wipe fs), and some can be very slow (shred).  So that may be good to distinguish in the metrics.  Or May it's good enough to categorize by storageclass? A single storageclass can only have one cleanup method

cofyc: Understood. I will use cleanup command here. Using storageclass is not good, because storageclass can be deleted from apiserver or recreated with different configuration.

cofyc:
@msau42 @dhirajh 
I've rebased on latest master and made a few changes, please take a look at again.

List of changes:

- Remove `PersistentVolumeProvision(Total|Latency)`.
- Add `PersistentVolumeDiscoveryTotal` and `PersistentVolumeDiscoveryLatency` for discovery.
- Add `PersistentVolumeDeleteTotal` and `PersistentVolumeDeleteLatency` for delete.
- Rename `PersistentVolumeCleanupFailedTotal` to `PersistentVolumeDeleteFailedTotal`.
- Add `type` for `PersistentVolumeDelete(Total|FailedTotal|Latency)` metrics.
  - `type` can be `process` or `job` (https://github.com/kubernetes-incubator/external-storage/pull/711).
- Add `capacity`, `cleanup_command` for `PersistentVolumeDeleteLatency` metric as @msau42 suggested.
  - `capacity` is broke down by 500G.
- Update `CleanupStatus.RemoveStatus()` to return `startTime` for metrics.
- Update `provisioner/REDME.md` and add a example service yaml file.

See docs for all metrics: https://github.com/cofyc/external-storage/blob/6c86c3f1d2fa959bba41372ac8a283afc19751f1/local-volume/provisioner/README.md#prometheus-metrics

local-volume/provisioner/pkg/deleter/deleter.go
```diff
@@ -126,6 +139,24 @@ func (d *Deleter) deletePV(pv *v1.PersistentVolume) error {
 				return fmt.Errorf("Error deleting PV %q: %v", pv.Name, err.Error())
 			}
 		}
+		mode := string(volMode)
+		deleteType := metrics.DeleteTypeProcess
+		if runjob {
+			deleteType = metrics.DeleteTypeJob
+		}
+		metrics.PersistentVolumeDeleteTotal.WithLabelValues(mode, deleteType).Inc()
+		if startTime != nil {
```
msau42: When is startTime nil?

local-volume/provisioner/pkg/deleter/jobcontroller.go
```diff
@@ -289,7 +305,8 @@ func NewCleanupJob(pv *apiv1.PersistentVolume, imageName string, nodeName string
 
 	// Annotate job with useful information that cannot be set as labels due to label name restrictions.
 	annotations := map[string]string{
-		DeviceAnnotation: blkdevPath,
+		DeviceAnnotation:    blkdevPath,
+		StartTimeAnnotation: time.Now().Format(time.RFC3339Nano),
```
msau42: I think the Job object already has its own start time. Can that be used?

cofyc: @msau42 job `status.startTime` is the time when job was acknowledged by the job controller. It does not include network latency, apiserver/controller-manager process latency. If apiserver/controller-manager failed to process the job in time, we couldn't know the latency through these metrics.

local-volume/provisioner/pkg/metrics/metrics.go
```diff
@@ -27,35 +27,63 @@ import (
...
-	// PersistentVolumeProvisionLatency is used to collect latency in seconds to provision persistent volumes.
-	PersistentVolumeProvisionLatency = prometheus.NewHistogramVec(
+	// PersistentVolumeDeleteLatency is used to collect latency in seconds to delete persistent volumes.
+	PersistentVolumeDeleteLatency = prometheus.NewHistogramVec(
 		prometheus.HistogramOpts{
 			Subsystem: LocalVolumeProvisionerSubsystem,
-			Name:      "persistentvolume_provision_latency",
-			Help:      "Latency in seconds to provision persistent volumes. Broken down by action, mode, capacity and cleanup_command.",
+			Name:      "persistentvolume_delete_latency",
+			Help:      "Latency in seconds to delete persistent volumes. Broken down by mode, type, capacity and cleanup_command.",
```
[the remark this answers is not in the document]
cofyc: @msau42 Done in https://github.com/kubernetes-incubator/external-storage/pull/721/commits/d2b746d55a3f7c0d63b8126390f50aad1bd7048b.

 - `mode` -> `persistent volume mode`
- `type` -> `delete type (process or job)`

local-volume/provisioner/pkg/metrics/metrics.go
```diff
@@ -27,35 +27,63 @@ import (
...
+			Help:      "Total number of persistent volumes discoveried. Broken down by mode.",
 		},
-		[]string{"action"},
+		[]string{"mode"},
+	)
+	// PersistentVolumeDiscoveryLatency is used to collect latency in seconds to discovery persistent volumes.
+	PersistentVolumeDiscoveryLatency = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Subsystem: LocalVolumeProvisionerSubsystem,
+			Name:      "persistentvolume_discovery_latency",
```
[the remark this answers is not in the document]
cofyc: Done.

cofyc:
@msau42 Build failure fixed (forgot to update metric variables in `cmd`) and squashed.

local-volume/provisioner/cmd/main.go
dhirajh: @msau42 Looks like I missed this comment, sorry about the delayed response. ConfigMap would have been ideal as one does not need to update docker image or mess around with entry point overrides in the manifest to change the CLI option. cc @cofyc

cofyc: I prefer CLI option. If it's configured via ConfigMap, a few issues we have to consider:

- on port change, should we restart listen goroutine to listen on new port?
- if we should re-listen on new port, should it fail if there is port conflict?

IMO, if we need to configure listen address (in rare cases i think), it's simpler to update deployment or damonset (and delete pods) to restart pods again. If there is port conflict, pods will fail and we will know it. This is normal operation in kubernetes.

PR #723, opened 2018-04-09
"Delete PV" should be belongs to provisioner

wenlxie:
"Delete PV" should be belongs to provisioner but not k8s


wongma7:
/lgtm
good catch

PR #728, opened 2018-04-11
Set slave mount propgation in local provisioner

msau42:
Set mount propgation mode anyway if the default changes.

gce/gke doesn't need mount propgation as the disks are mounted during node install before provisioner runs.


jsafrane:
@childsb, please merge

Issue #738, opened 2018-04-24
Labels added to snapshot object are lost


disrani-px:
The labels for the volumesnapshot object are replaced with tags when updating the metadata in updateVolumeSnapshotMetadata(). User-defined labels should be retained when adding these tags instead of replacing them

/area snapshot

disrani-px:
Fixed with PR #740

PR #740, opened 2018-04-25
Don't overwrite labels provided by user in snapshot object

disrani-px:
Issue #738 

/area snapshot
/assign @tsmetana


cehoffman:
There is still an window in which user labels can be overwritten I think. The request to the api server needs to be a merge patch.

1. volumesnapshot created
2. snapshotter gets resource
3. user submits label merge patch
4. snapshotter commits updated volumesnapshot resource

In this setup the user labeling occurred in between the snapshotter getting and applying its work to the resource.

disrani-px:
@cehoffman The update on the outdated object would fail since the resourceVersion in the stored object would have been newer than the object being updated. So the labels will not get removed. We will just try updating again with the newer version of the object.

More info on how this is handled in Kubernetes here: https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#concurrency-control-and-consistency

disrani-px:
@wongma7 @childsb Can this be merged?

PR #743, opened 2018-04-26
Local volume dynamic provisioning

lichuqiang:
Design: https://github.com/kubernetes/community/pull/1914

The PR includes:
- Refactor config to include dynamic provisioning config
- Update discovery to be aware of volumemode
- Update populator to watch PVCs for dynamic provisioning
- Implement storage backend
- Implement dynamic provisioning manager
- Update deleter to also handle dynamically provisioned volumes
- Implement capacity reconciler
- Deployment update

TODO:
- Wait for https://github.com/kubernetes/kubernetes/pull/63193 merged, and update API dependency
- Rely on https://github.com/kubernetes/kubernetes/pull/63011, and pass in Fstype for provisioned volumes

### DIFF
--- a/local-volume/helm/provisioner/templates/provisioner.yaml
+++ b/local-volume/helm/provisioner/templates/provisioner.yaml
@@ -3,12 +3,21 @@
 {{- range $tmp, $val := $classPath }}
    {{- range $storageClass, $classConfig := $val }}
 {{ $storageClass }}:
+   {{- if $classConfig.hostDir }}
    hostDir: {{ $classConfig.hostDir }}
    mountDir: {{ if $classConfig.mountDir }} {{$classConfig.mountDir }} {{ else }} {{ $classConfig.hostDir }} {{ end }}
+   {{- end }}
    {{- if $classConfig.blockCleanerCommand }}
    blockCleanerCommand:
    {{- range $tmp, $val := $classConfig.blockCleanerCommand }}
      - "{{ $val -}}"{{- end}}
+   {{- end }}
+   {{- if $classConfig.volumeMode }}
+   volumeMode: {{ $classConfig.volumeMode }}
+   {{- end }}
+   {{- if $classConfig.lvmVolumeGroup }}
+   lvm:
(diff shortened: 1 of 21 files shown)


KlavsKlavsen:
https://github.com/kubernetes/kubernetes/pull/63193 As I understand it - is IN Kubernetes 1.11.. so this feature should be ready to use ?

KlavsKlavsen:
I have an 1.11 cluster I am dying to test this out on.. whats left, before it'll work? It would be great to just setup a config-map pointing to a VG the allocator can use - and then just have it use extents from that - if there is sufficient available (or select another node which HAS the necessary extents :)

lichuqiang:
@KlavsKlavsen Sorry, I'm afraid the feature will be delayed to next release due to some dependency issue.
The lvm part is relatively clear, what we've not decided yet is about how to integrate with k8s (e.g, how to get capacity reported).
And we are planning to implement an CSI driver for the dynamically provisioning things, the plugin here in external-storage repo will only be responsible for the static discovery.

KlavsKlavsen:
@lichuqiang aww.. Any links to current state of discussion on the missing part (CSI driver) and its implementation ? (and what is static discovery.. only the "node-local" part - where it finds WHICH folder under /mnt/fast-disks - to use ? pretty much meaning the current local-volume will be sufficient, as the CSI driver will setup the /mnt/fast-disks/$podmnt part ?)

lichuqiang:
>  Any links to current state of discussion on the missing part (CSI driver) and its implementation ?

Nope， it's still in an internal discussion stage. I'll have an update on the proposal once we overcome the capacity report issue.

> (and what is static discovery.. only the "node-local" part - where it finds WHICH folder under /mnt/fast-disks - to use ?

the static and dynamic provision are two independent pattern to make use of local volumes.
in static way, users need to pre-create volumes and expose to the plugin, and the plugin is only responsible to create PV objects in k8s with the provided volume.

In dynamic way, the plugin is responsible to maintain the lifecycle of both PVs in k8s and underlying volumes (and maybe the volume groups too)

PR #753, opened 2018-05-05
updated flexplugin to look for 200 volumes

ZachGlassman:
Updates the `getVolumeByName` function to request 200 volumes per name.  

Also fixes a type error in the `provision.go` code.

### DIFF
--- a/digitalocean/cmd/digitalocean-flexplugin/main.go
+++ b/digitalocean/cmd/digitalocean-flexplugin/main.go
@@ -110,7 +110,7 @@ func (c *cloud) findNode(nodeName string) (int, error) {
 }
 
 func (c *cloud) getVolumeByName(volumeName string) (string, error) {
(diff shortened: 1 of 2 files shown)


ZachGlassman:
@klausenbusk can you have a look here?

klausenbusk:
Could you explain why this is needed?

klausenbusk:
> We have found this is a bit more resiliant to API failures in the paging system. In addition, the Digital Ocean API has a rate limiter and so this can help reduce the number of requests needed when working with large numbers of volumes.

> As for the change in provision.go, I'm guessing this was due to an update, but this (or some other change) seems necessary to compile (at least on my machine).

Could you update the commit message to include this? Besides that it LGTM :)

klausenbusk:
LGTM :)

> As for the change in provision.go, I'm guessing this was due to an update, but this (or some other change) seems necessary to compile (at least on my machine).

It was changed in 1.10: https://github.com/kubernetes/kubernetes/pull/56460

Issue #757, opened 2018-05-09
[OpenStack Manila] Add support for other kinds of shares


gman0:
Currently, the [Manila external-provisioner](https://github.com/kubernetes-incubator/external-storage/tree/master/openstack-sharedfilesystems) only supports NFS shares and this support is hard-coded. I'd propose doing a refactoring, making it easier to accommodate other kinds of shares.

I already have working code that creates a Ceph share, and the resulting PV is then passed to [CSI CephFS driver](https://github.com/ceph/ceph-csi) which then mounts the volume.

I can have a PR ready shortly, but it requires to bump the vendored version of `gophercloud` library - I'll open a new issue for this.

PR #760, opened 2018-05-11
manila: add support for different kinds of shares

gman0:
For https://github.com/kubernetes-incubator/external-storage/issues/757

Things that were considered:
- a share may require its own specific configuration from `StorageClass`
- different types of shares may have different requirements on the access rules they're granted
- parameters of `PersistentVolumeSource` vary between share kinds
- with the goal of making it easier to add and maintain those different kinds

Starting with package `sharebackends`, a share backend is an interface that handles the share specific stuff when setting up a share: granting access and creating a `PersistentVolumeSource` object (used in the final `PersistentVolume` in `Provision()` call). Those actions may have allocated some resources (e.g. a new k8s Secret), and each share backend is responsible for freeing those resources in its `Release()` method, which is then called in `Delete()`.

As for configuration, package `shareoptions` contains validation and parsing of parameters passed from `StorageClass`. Those can be either common for all share kinds (in `.../pkg/shareoptions/common.go`), or specific for either different protocols (e.g. NFS, CEPHFS in `.../pkg/shareoptions/protocol.go`) or backends (e.g. nfs, csi-cephfs in `.../pkg/shareoptions/backend.go`)

Adding a share backend would now consist of:
1. adding struct fields to `.../pkg/shareoptions/backend.go` with appropriate field tags
2. creating a separate file in the `sharebackends` package with a struct implementing the `ShareBackend` interface
3. registering the backend in `.../pkg/sharebackend.go` in `init()`

Tested with a Ceph cluster and [csi-cephfs](https://github.com/ceph/ceph-csi)

I'll update the unit tests in the upcoming commits, I'd just like to have some feedback if the design is ok.


tsmetana:
It won't compile in my environment:
```
CGO_ENABLED=0 GOOS=linux go build -a -ldflags '-extldflags "-static"' -o _output/bin/manila-provisioner cmd/manila-provisioner/main.go
# github.com/kubernetes-incubator/external-storage/openstack-sharedfilesystems/pkg/sharedfilesystems/sharebackends
pkg/sharedfilesystems/sharebackends/csicephfs.go:94:24: undefined: shares.ListAccessRights
Makefile:46: recipe for target 'manila-provisioner' failed
make: *** [manila-provisioner] Error 2
```
Looks like it depends on a newer gophercloud than what we currently vendor in external-storage.

gman0:
Closing in favor of https://github.com/kubernetes/cloud-provider-openstack/pull/182. If that gets merged please continue the development over there.

PR #763, opened 2018-05-14
Manila provisioner: Fix export location parsing

tsmetana:
The Manila provisioner fails to correctly extract a share location if it's in the server:port:/path format or if using IPv6. This patch uses the net.SplitHostPort to extract the port and if the extraction is successful the found value is used as the mountport mount option (this looks to be the case for CephNFS at least).

@rootfs, @tombarron PTAL

### DIFF
--- a/openstack-sharedfilesystems/pkg/sharedfilesystems/sharedfilesystems.go
+++ b/openstack-sharedfilesystems/pkg/sharedfilesystems/sharedfilesystems.go
@@ -18,6 +18,7 @@ package sharedfilesystems
 
 import (
 	"fmt"
+	"net"
 	"strconv"
 	"strings"
 
@@ -171,7 +172,7 @@ func FillInPV(options controller.VolumeOptions, share shares.Share, exportLocati
 
 	storageSize := resource.MustParse(fmt.Sprintf("%dG", share.Size))
 	PVAccessMode := getPVAccessMode(options.PVC.Spec.AccessModes)
-	server, path, err := getServerAndPath(exportLocation.Path)
+	server, port, path, err := getServerAndPath(exportLocation.Path)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


tombarron:
This seems right.  CephNFS will use port 2049 and not specify it normally and IPv6 support for this back end is not there yet but may make it for the Rocky (upcoming) release.  FWIW here is what I see returned from OpenStack master with CephNFS back end:

(overcloud) [stack@undercloud-0 ~]$ curl -i -X GET http://10.0.0.101:8786/v2/ce4a143ce2e4447398249f35744aa6eb/shares/6cae32ef-ed70-4f17-aa31-31b004d0be51/export_locations -H "X-Openstack-Manila-Api-Version: 2.42" -H "X-Auth-Token: gAAAAABa-fBYWjDL4-kCfZYZ0Lz5c92M6cQwBp0DkN8yV68F-ClkYeuvJp0WB_3d0n8dTi_JPe6_GBOAxSB_Du6LgEEr4Shh92UVVcNOvF_jYB_ZUgzrirDqqH4rNJZTCmUhwAPqndY2jsXhJiyDmc4JPlGNhIYp4YBgGRaBf8HaymLN-tjrJ80" -H "Accept: application/json"
HTTP/1.1 200 OK
X-Compute-Request-Id: req-92ab30f6-7d67-41a0-b716-0bccdfe69efc
Content-Type: application/json
Content-Length: 253
X-Openstack-Manila-Api-Version: 2.42
Vary: X-OpenStack-Manila-API-Version
Date: Mon, 14 May 2018 20:25:16 GMT

{"export_locations": [{"path": "172.17.5.12:/volumes/_nogroup/9a13ff6a-6bb5-4e31-87c7-6040b5a67d51", "share_instance_id": "9a13ff6a-6bb5-4e31-87c7-6040b5a67d51", "is_admin_only": false, "id": "afb60a15-7091-440f-bffa-bd9e964b112a", "preferred": false}]}(overcloud) [stack@undercloud-0 ~]$ 

Of course we'd like this provider to work more generally than just CephNFS back end so handling the port seems a good idea.

tsmetana:
Thanks for the review.

If you seek more generic functionality from the Manila provisioner then you are interested in the PR #760. That one is larger and will take more time to review I suppose. This one is meant to fix an existing bug only and be backportable...

PR #770, opened 2018-05-18
cephfs support capacity

silenceshell:
There are two ways to mount cephfs in kubelet: ceph-fuse and kernel mount. While the kernel mount does not support quota, the ceph-fuse support quota.

However, project external-storage just ignore the capacity, so even I use ceph-fuse to mount volume, I still cannot set size of the volume of cephfs.

So I add some codes to support this feature and have checked in my env(there are also some updates in k8s and I will try to merge them to k8s).

Look forward to the community's opion!

### DIFF
--- a/ceph/cephfs/cephfs-provisioner.go
+++ b/ceph/cephfs/cephfs-provisioner.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"strconv"
 	"strings"
 
 	"github.com/golang/glog"
@@ -60,13 +61,16 @@ type cephFSProvisioner struct {
 	identity string
 	// Namespace secrets will be created in. If empty, secrets will be created in each PVC's namespace.
 	secretNamespace string
+	// enable PVC quota
+	enableQuota bool
 }
 
-func newCephFSProvisioner(client kubernetes.Interface, id string, secretNamespace string) controller.Provisioner {
+func newCephFSProvisioner(client kubernetes.Interface, id string, secretNamespace string, enableQuota bool) controller.Provisioner {
 	return &cephFSProvisioner{
 		client:          client,
 		identity:        id,
 		secretNamespace: secretNamespace,
+		enableQuota:     enableQuota,
 	}
 }
 
@@ -122,7 +126,13 @@ func (p *cephFSProvisioner) Provision(options controller.VolumeOptions) (*v1.Per
 	user := fmt.Sprintf("kubernetes-dynamic-user-%s", uuid.NewUUID())
 	// provision share
 	// create cmd
-	cmd := exec.Command(provisionCmd, "-n", share, "-u", user)
+	args := []string{"-n", share, "-u", user}
+	if p.enableQuota {
+		capacity := options.PVC.Spec.Resources.Requests[v1.ResourceName(v1.ResourceStorage)]
+		requestBytes := strconv.FormatInt(capacity.Value(), 10)
+		args = append(args, "-s", requestBytes)
+	}
+	cmd := exec.Command(provisionCmd, args...)
 	// set env
 	cmd.Env = []string{
 		"CEPH_CLUSTER_NAME=" + cluster,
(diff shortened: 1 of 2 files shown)


cofyc:
cc @rootfs

cofyc:
hi, for backward compatibility, could you add a CLI bool option to control this feature (default: false)? Users can turn it on when they expect quota limit on cephfs volume.

silenceshell:
Yeah this is nessary. Instead of add a bool option, what about make the `-s` parameter optional? If it is not specified(the old way), `size` will remain value 0, and `cephfs.create_share` won't take the size param. 
If this is OK I will update the code today. Thank you for your comment!

silenceshell:
Is the bool option necessary? I think the **optional** `-s` is enough for backward compatibility. 
BTW, if it is necessary, how about `--set-quota`?

cofyc:
What I mean is backward compatibility of `cephfs-provisioner` behavior not `cephfs_provisioner.py` script.

Currently, users may use cephfs-provisioner to create cephfs PV, no disk quota on volume path. If they upgrade `cephfs-provisioner` to a version which adds disk quota on volume path. Pods which expect to write more data than PVC capacity, may have problem.

silenceshell:
hi @cofyc 
I add a `--set-quota` option to the CLI tool so default the size is ignored. Could you help reviewing the codes?

cofyc:
Looks good to me. BTW, please squash into one commit.

ceph/cephfs/cephfs-provisioner.go
```diff
@@ -60,13 +61,16 @@ type cephFSProvisioner struct {
 	identity string
 	// Namespace secrets will be created in. If empty, secrets will be created in each PVC's namespace.
 	secretNamespace string
+	// set PVC quota
```
rootfs: prefer `enable quota`. `set PVC quota` could be confusing

ceph/cephfs/cephfs-provisioner.go
```diff
@@ -318,6 +328,7 @@ var (
 	kubeconfig      = flag.String("kubeconfig", "", "Absolute path to the kubeconfig")
 	id              = flag.String("id", "", "Unique provisioner identity")
 	secretNamespace = flag.String("secret-namespace", "", "Namespace secrets will be created in (default: '', created in each PVC's namespace)")
+	setQuota        = flag.Bool("set-quota", false, "Set PVC quota")
```
rootfs: ditto

ceph/cephfs/cephfs-provisioner.go
```diff
@@ -122,7 +126,13 @@ func (p *cephFSProvisioner) Provision(options controller.VolumeOptions) (*v1.Per
 	user := fmt.Sprintf("kubernetes-dynamic-user-%s", uuid.NewUUID())
 	// provision share
 	// create cmd
-	cmd := exec.Command(provisionCmd, "-n", share, "-u", user)
+	args := []string{"-n", share, "-u", user}
+	if p.setQuota {
+		capacity := options.PVC.Spec.Resources.Requests[v1.ResourceName(v1.ResourceStorage)]
+		requestBytes := strconv.FormatInt(capacity.Value(), 10)
+		args = append(args, "-s", requestBytes)
```
rootfs: also fix the comment here
https://github.com/silenceshell/external-storage/blob/ae906166db5d758bf2de9abefe8488b2951ee7cd/ceph/cephfs/cephfs-provisioner.go#L188

ceph/cephfs/cephfs-provisioner.go
```diff
@@ -175,7 +185,9 @@ func (p *cephFSProvisioner) Provision(options controller.VolumeOptions) (*v1.Per
 		Spec: v1.PersistentVolumeSpec{
 			PersistentVolumeReclaimPolicy: options.PersistentVolumeReclaimPolicy,
 			AccessModes:                   options.PVC.Spec.AccessModes,
-			Capacity: v1.ResourceList{ //FIXME: kernel cephfs doesn't enforce quota, capacity is not meaningless here.
+			Capacity: v1.ResourceList{
+				// Quotas are supported by the userspace client(ceph-fuse, libcephfs), or kernel client >= 4.17 but only on mimic clusters.
+				// In other cases capacity is meaningless here.
```
cofyc: Perhaps we need add something like "If quota is enabled, provisioner will set `ceph.quota.max_bytes` on volume path." because quota is optional now.

ceph/cephfs/cephfs_provisioner/cephfs_provisioner.py
```diff
@@ -19,6 +19,7 @@
 import getopt
 import sys
 import json
+import cephfs
```
ltxdzh: module cephfs seems not be used and should be removed.

silenceshell: yeah. I will remove it.

ceph/cephfs/cephfs-provisioner.go
cofyc: Please update comments here, then squash into one commit. Thanks for your patience! I tested it and it works very well.

cc @rootfs Any opinion?

silenceshell: Everything is done now  ~~

cofyc:
@wongma7 Could you help merge this?

PR #773, opened 2018-05-23
CephFS Provisioner: Add option to disable namespace isolation.

cofyc:

Linux kernel cephfs only support Rados Pool Namespace in 4.8+. See
https://github.com/torvalds/linux/commit/779fe0fb8e1883d5c479ac6bd85fbd237deed1f7.

If security is not critical, we can configure cephfs provisioner not to enforce pool namespace, then we can use it with linux kernels before 4.8.

fixes https://github.com/kubernetes-incubator/external-storage/issues/345

This patches `ceph_volume_client.py` with patch from https://github.com/ceph/ceph/pull/21808.

### DIFF
--- a/ceph/cephfs/Dockerfile
+++ b/ceph/cephfs/Dockerfile
@@ -15,9 +15,10 @@
 FROM centos:7
 
 ENV CEPH_VERSION "mimic"
-RUN rpm -Uvh https://download.ceph.com/rpm-$CEPH_VERSION/el7/noarch/ceph-release-1-1.el7.noarch.rpm && \
+RUN rpm --import 'https://download.ceph.com/keys/release.asc' && \
+  rpm -Uvh https://download.ceph.com/rpm-$CEPH_VERSION/el7/noarch/ceph-release-1-1.el7.noarch.rpm && \
   yum install -y epel-release && \
-  yum install -y --nogpgcheck ceph-common python-cephfs && \
+  yum install -y ceph-common python-cephfs && \
   yum clean all
 
 COPY cephfs-provisioner /usr/local/bin/cephfs-provisioner
--- a/ceph/cephfs/cephfs-provisioner.go
+++ b/ceph/cephfs/cephfs-provisioner.go
@@ -149,6 +149,9 @@ func (p *cephFSProvisioner) Provision(options controller.VolumeOptions) (*v1.Per
 	if deterministicNames {
 		cmd.Env = append(cmd.Env, "CEPH_VOLUME_GROUP="+options.PVC.Namespace)
 	}
+	if *disableCephNamespaceIsolation {
+		cmd.Env = append(cmd.Env, "CEPH_NAMESPACE_ISOLATION_DISABLED=true")
+	}
 
 	output, cmdErr := cmd.CombinedOutput()
 	if cmdErr != nil {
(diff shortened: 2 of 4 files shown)


rootfs:
not sure if patch cephfs volume is a good idea here, cc @jcsp

ceph/cephfs/Dockerfile
```diff
@@ -14,12 +14,21 @@
...
 FROM centos:7
 
-ENV CEPH_VERSION "luminous"
-RUN rpm -Uvh https://download.ceph.com/rpm-$CEPH_VERSION/el7/noarch/ceph-release-1-1.el7.noarch.rpm && \
+ENV CEPH_RELEASE "luminous"
+ENV CEPH_VERSION "12.2.5"
+RUN rpm -Uvh https://download.ceph.com/rpm-$CEPH_RELEASE/el7/noarch/ceph-release-1-1.el7.noarch.rpm && \
   yum install -y epel-release && \
-  yum install -y --nogpgcheck ceph-common python-cephfs && \
+  yum install -y --nogpgcheck librados2-$CEPH_VERSION libradosstriper1-$CEPH_VERSION libcephfs2-$CEPH_VERSION librbd1-$CEPH_VERSION librgw2-$CEPH_VERSION && \
```
jcsp: It is not necessary to use --nogpgcheck (I know the existing Dockerfile was using it, that doesn't make it okay).  It's better (although still not ideal) to install the GPG key, like Rook does:
https://github.com/rook/rook/blob/master/images/ceph/Dockerfile.base#L18

jcsp:
This isn't an approach I'd recommend -- it is confusing to have a package installed that claims to be 12.2.5, but is in reality running different code.  The correct way to patch an RPM is to rebuild it with the patch included, that way you can properly adjust the version to indicate that what you're running is not a normal Ceph package.

The least effort thing here is to wait until the backport merges upstream (https://github.com/ceph/ceph/pull/21897) and the next 12.2.x release comes out.

cofyc:
OK, I'll wait until the next 12.2.x release comes out.

rpardini:
I was hit by this, using default Xenial 4.4.0 kernel.
It was simple to upgrade to 4.15.0 (HWE kernel), which gives me better CRUSH tunables and RADOS namespace support.
It might be helpful to document the 4.8+ requirement somewhere...?

cofyc:
@jcsp @rootfs 
Ceph release has been [updated to "mimic"](https://github.com/kubernetes-incubator/external-storage/pull/841), its latest version supports namespace isolation now: https://github.com/ceph/ceph/blob/v13.2.0/src/pybind/ceph_volume_client.py#L603. Now we don't need to patch libcephfs python module. Please review it again.

ceph/cephfs/cephfs-provisioner.go
```diff
@@ -345,12 +351,13 @@ func (p *cephFSProvisioner) parsePVSecret(namespace, secretName string) (string,
...
-	secretNamespace = flag.String("secret-namespace", "", "Namespace secrets will be created in (default: '', created in each PVC's namespace)")
-	enableQuota     = flag.Bool("enable-quota", false, "Enable PVC quota")
-	metricsPort     = flag.Int("metrics-port", 0, "The port of the metrics server (set to non-zero to enable)")
+	master                    = flag.String("master", "", "Master URL")
+	kubeconfig                = flag.String("kubeconfig", "", "Absolute path to the kubeconfig")
+	id                        = flag.String("id", "", "Unique provisioner identity")
+	secretNamespace           = flag.String("secret-namespace", "", "Namespace secrets will be created in (default: '', created in each PVC's namespace)")
+	enableQuota               = flag.Bool("enable-quota", false, "Enable PVC quota")
+	metricsPort               = flag.Int("metrics-port", 0, "The port of the metrics server (set to non-zero to enable)")
+	disableNamespaceIsolation = flag.Bool("disable-namespace-isolation", false, "Disable namespace isolation")
```
rootfs: nit: `disable cephfs namespace isolation` , since namespace itself has too many meanings

cofyc: Fixed. I used `ceph` instead of `cephfs` because other names in this program use `ceph` prefix too.

rootfs:
/lgtm
@wongma7 can you merge it? thanks

PR #776, opened 2018-05-24
Local Volume Provisioner: Support Reclaim Policy

### LINKED ISSUE #771: Is there a way to set RECLAIM POLICY to "RETAIN" in the helm chart implementation?
I was running the following commands to deploy the provisioner:

$ helm template ./helm/provisioner --set engine=gke > ./provisioner/deployment/kubernetes/provisioner_generated.yaml
$ kubectl apply --username=admin --password=xxxxxx -f provisioner/deployment/kubernetes/provisioner_generated.yaml

This set my pv(s)'s Reclaim Policy to "Delete".

cofyc:

Support Reclaim Policy. 

fixes https://github.com/kubernetes-incubator/external-storage/issues/771

`Recycle` policy is deprecated (https://kubernetes.io/docs/concepts/storage/persistent-volumes/#recycle), I suggest not supporting it.

Although we can treat it as `Delete` policy, but for `Recycle` if we want to support we should not delete it.

### DIFF
--- a/local-volume/provisioner/pkg/populator/populator.go
+++ b/local-volume/provisioner/pkg/populator/populator.go
@@ -17,18 +17,11 @@ limitations under the License.
 package populator
 
 import (
-	"os"
-	"time"
-
 	"github.com/golang/glog"
 	"github.com/kubernetes-incubator/external-storage/local-volume/provisioner/pkg/common"
 
 	"k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/util/wait"
-	"k8s.io/apimachinery/pkg/watch"
-	kcache "k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/cache"
 )
 
 // The Populator uses an Informer to populate the VolumeCache.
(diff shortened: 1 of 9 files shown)


local-volume/provisioner/pkg/util/api_util.go
```diff
@@ -97,6 +101,10 @@ func (u *apiUtil) DeleteJob(jobName string, namespace string) error {
 	return nil
 }
 
+func (u *apiUtil) GetStorageClass(scName string) (*storage_v1.StorageClass, error) {
+	return u.client.StorageV1().StorageClasses().Get(scName, metav1.GetOptions{})
```
msau42: The RBAC permissions need to add StorageClass Get permissions

cofyc: The `system:persistent-volume-provisioner` clusterrole [we are using](https://github.com/kubernetes-incubator/external-storage/blob/master/local-volume/helm/provisioner/templates/provisioner-cluster-role-binding.yaml#L13) contains `get` permissions on `storageclasses`, no need to add.

msau42: Can we use informer to get StorageClass objects? To reduce chance of exhausting API quota.

cofyc: @msau42 Updated in https://github.com/kubernetes-incubator/external-storage/pull/776/commits/346c2e28611b81616f4ad21f36913d7228e00712.

- start StorageClass informer controller in Discoverer, then it can get StorageClass from informer cache
- change `common.RuntimeConfig.Client` to be `kubernetes.Interface`, then we can use `fake.Clientset` for testing
- update `Populator.Start` to be same as `Discoverer.Start` which are almost the same
- prefer using `wait.NeverStop` which is more idiomatic in k8s

local-volume/provisioner/pkg/populator/populator.go
```diff
@@ -41,8 +39,8 @@ func NewPopulator(config *common.RuntimeConfig) *Populator {
 	return &Populator{RuntimeConfig: config}
 }
 
-// Start launches the PV informer
-func (p *Populator) Start() {
+// Start launches the PV informer and waits for cache to sync, returns error if it fails.
+func (p *Populator) Start(stopCh <-chan struct{}) error {
 	_, controller := kcache.NewInformer(
```
msau42: Would using SharedInformers be better here for both PV and StorageClass objects?

cofyc: No, IMO [`SharedInformer`](https://github.com/kubernetes/client-go/blob/v7.0.0/tools/cache/shared_informer.go#L34) is preferred if there are multiple listeners for same cache. For examples, two controllers who both will listen on events of PersistentVolume can use `SharedInformer ` to share data cache. In local volume provisioner, PersistentVolums cache is only consumed Populator,  StorageClasse Cache only by Discoverer.

msau42: The interface seems simpler as you don't have to start separate controllers

cofyc: Agree. I'll updated to use `SharedInformer`.

cofyc:
BTW, `Populator` use `0` as resync period, which I think in theory is right choice, because we don't need to reconcile periodically. But in practice, `list` & `watch` probably lost events (someone is tracking [event lost issues](https://github.com/kubernetes/kubernetes/pull/61067)), should we use a non-zero value as last insurance?

cofyc:
However, `scheduler` disabled resync period too, i'd found a old discussion why scheduler does not use resync period here: https://github.com/kubernetes/kubernetes/pull/16840#issuecomment-154051747.

I actually don't know what's the right resync period value should to use in this case. I prefer `0` but am  afraid cache may become stale during long running time. But if using non-zero value, don't know how much time is enough.

local-volume/provisioner/pkg/populator/populator.go
cofyc: Updated in https://github.com/kubernetes-incubator/external-storage/pull/776/commits/af8c713a0e7c3bf30eeada9aef449f0baf7cda6a.

I've done some more refactoring which I found necessary:

- Use `informers.SharedInformerFactory` to create SharedInformers for PersistentVolumes, StorageClasses, and Jobs.
  - Note that, for Jobs we are using a customized constructor
- With `informers.SharedInformerFactory`, it's easier to start informers and wait for all informers' cache were synced before running controller logic
  - Previously, `JobController` waits for cache in a separate go routine which is not right, because it is not a standalone controller from my perspective.

msau42:
I think it is not necessary for this to be very performant. Maybe a resync every 5 minutes is fine?

msau42:
This lgtm, can you squash into 2 commits? one for informers, and one for the new functionality?

cofyc:
Done.

- https://github.com/kubernetes-incubator/external-storage/pull/776/commits/68957947197dc19f4519938541078a62cbcd84da: use SharedInformers and some refactoring
- https://github.com/kubernetes-incubator/external-storage/pull/776/commits/5133a8996ee855b3b359eb01af33fc4a09a96b3a: support StorageClass ReclaimPolicy

About using non-zero resync period, I'll create another PR for it (probably it's better to add a user config allowing users to override like `kube-controller-mananger`'s `--min-resync-period`).

msau42:
/lgtm
Thanks!

PR #789, opened 2018-06-04
local volume provisioner: improve helm chart

cofyc:
- Get rid of `engine` thing, it's better to use differnt values files for different environments.
- Rename `jobsrole` and `local-storage-provisioner-job-binding` to `local-storage-provisioner`, and only create them when necessary.
- Able to create StorageClass object and configure it.
- Add docs for all chart configurations.
  - https://github.com/cofyc/external-storage/blob/refactor_heml_chart/local-volume/helm/README.md#configurations
- Add some examples.
  - https://github.com/cofyc/external-storage/blob/refactor_heml_chart/local-volume/helm/README.md#examples

PR #792, opened 2018-06-06
Added support for tolerations in local-volume provisioner daemonset helm template

dubuc:
Added support for tolerations in the DaemonsSet Helm template for the static local volume provisioner.

### DIFF
--- a/local-volume/helm/provisioner/values.yaml
+++ b/local-volume/helm/provisioner/values.yaml
@@ -85,3 +85,9 @@ daemonset:
   #  nodeLabels:
   #    - failure-domain.beta.kubernetes.io/zone
   #    - failure-domain.beta.kubernetes.io/region
+  #
+  # If configured, tolerations will add a toleration field to the DaemonSet PodSpec.
+  #
+  # Node tolerations for local-volume-provisioner scheduling to nodes with taints.
+  # Ref: https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/
+  tolerations: []
--- a/local-volume/helm/README.md
+++ b/local-volume/helm/README.md
@@ -69,6 +69,7 @@ provisioner chart and their default values.
 | daemonset.serviceAccount               | Provisioner DaemonSet service account.                                                                | str      | `local-storage-admin`                                      |
 | daemonset.kubeConfigEnv                | Specify the location of kubernetes config file.                                                       | str      | `-`                                                        |
 | daemonset.nodeLabels                   | List of node labels to be copied to the PVs created by the provisioner.                               | list     | `-`                                                        |
+| daemonset.tolerations                  | List of tolerations to be applied to the Provisioner DaemonSet.                                       | list     | `-`                                                        |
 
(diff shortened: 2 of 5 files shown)


msau42:
cc @cofyc

dubuc:
CLA verification

cofyc:
This makes sense to me. Could you wait a while? There are some refactorings in  https://github.com/kubernetes-incubator/external-storage/pull/789, you can make changes based on new helm chart values/template files after that pr is merged.

local-volume/helm/provisioner/values.yaml
```diff
@@ -103,3 +103,12 @@ daemonset:
...
   #    - failure-domain.beta.kubernetes.io/region
+  #
+  # If configured, tolerations will add a toleration field to the DaemonSet PodSpec.
+  #
+  #  tolerations:
+  #   - key: "{Key is the taint key that the toleration applies to.}"
+  #     operator: Equal|Exist
+  #     value: "{Value is the taint value the toleration matches to. If the operator is Exists, the value should be empty, otherwise just a regular string.}"
+  #     effect: NoSchedule|PreferNoSchedule|NoExecution
+  #     tolerationSeconds: <optional, by default not set>
```
cofyc: Because toleration is a spec of kubernetes api, add a reference to its docs will be better.

```
  # Node tolerations for local-volume-provisioner scheduling to nodes with taints.
  # Ref: https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/
  tolerations: []
    # - key: "key"
    #   operator: "Equal|Exists"
    #   value: "{Value is the taint value the toleration matches to. If the operator is Exists, the value should be empty, otherwise just a regular string.}"
    #   effect: "NoSchedule|PreferNoSchedule|NoExecute(1.6 only)"
    #   tolerationSeconds: "{Optional field that dictates how long the pod will stay bound to the node after the taint is added.}"
```

dubuc: Thank you for that input, it does make more sense :) Fixed.

dubuc:
Yes no worries, I manually packaged a chart in the meantime. :) I can refactor a little after your updates. Thanks for the feedback.

cofyc:
@dubuc https://github.com/kubernetes-incubator/external-storage/pull/789 has been merged, please rebase on latest master.
btw please also
- add a example in `helm/examples` (and also added a expected yaml file in `helm/test/expected`, test it with `helm/test/run.sh` script) 
- update helm/README.md configurations section

dubuc:
@cofyc Hi! Sorry for the delay, but was quite busy the past few weeks. I've amended the PR to reflect the changes required. Thanks for the work you did with the tests and documentation! :) 👍 Please let me know if something else is required.

msau42:
/lgtm
Thanks!

msau42:
@dubuc I had a follow up question on this.  In your example, you are tolerating a master node.  Does that mean you want to expose the local disks on your master node and also have workload pods scheduled to the master node?

PR #796, opened 2018-06-13
Add metrics server support for provision controller

cofyc:

This pr add metrics server support for provision controller. 

Like kubelet's `storage_operation_xxx` metrics, it's better we can have some metrics to monitor latency of provision/delete operations and failures in provisioner.

Current metrics:

- controller_persistentvolumeclaim_provision_total
- controller_persistentvolumeclaim_provision_failed_total
- controller_persistentvolumeclaim_provision_duration_seconds
- controller_persistentvolume_delete_total
- controller_persistentvolume_delete_failed_total
- controller_persistentvolume_delete_duration_seconds

Metrics server is disabled by default. Each provisioner can enable based on needs. 
rbd/cephfs example: https://github.com/kubernetes-incubator/external-storage/pull/797/files

Examples:

```
...
controller_persistentvolume_delete_duration_seconds_bucket{class="rbd",le="10"} 4
controller_persistentvolume_delete_duration_seconds_bucket{class="rbd",le="+Inf"} 4
controller_persistentvolume_delete_duration_seconds_sum{class="rbd"} 1.0242772740000001
controller_persistentvolume_delete_duration_seconds_count{class="rbd"} 4
controller_persistentvolume_delete_total{class="rbd"} 4
...
controller_persistentvolumeclaim_discovery_duration_seconds_bucket{class="rbd",le="10"} 4
controller_persistentvolumeclaim_discovery_duration_seconds_bucket{class="rbd",le="+Inf"} 4
controller_persistentvolumeclaim_discovery_duration_seconds_sum{class="rbd"} 1.467650504
controller_persistentvolumeclaim_discovery_duration_seconds_count{class="rbd"} 4
controller_persistentvolumeclaim_provision_failed_total{class="rbd"} 9
controller_persistentvolumeclaim_provision_total{class="rbd"} 4
```

### DIFF
--- a/lib/controller/controller.go
+++ b/lib/controller/controller.go
@@ -18,15 +18,21 @@ package controller
 
 import (
 	"fmt"
+	"net"
+	"net/http"
 	"os/exec"
 	"reflect"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
 
 	"github.com/golang/glog"
+	"github.com/kubernetes-incubator/external-storage/lib/controller/metrics"
 	"github.com/kubernetes-incubator/external-storage/lib/leaderelection"
 	rl "github.com/kubernetes-incubator/external-storage/lib/leaderelection/resourcelock"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"k8s.io/api/core/v1"
 	storage "k8s.io/api/storage/v1"
 	storagebeta "k8s.io/api/storage/v1beta1"
@@ -34,6 +40,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/uuid"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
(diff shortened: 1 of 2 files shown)


cofyc:
cc @childsb What do you think?

cofyc:
cc @wongma7

lib/controller/controller.go
```diff
@@ -156,6 +170,12 @@ const (
 	DefaultRetryPeriod = 2 * time.Second
 	// DefaultTermLimit is used when option function TermLimit is omitted
 	DefaultTermLimit = 30 * time.Second
+	// DefaultMetricsPort is used when option function MetricsPort is omitted
+	DefaultMetricsPort = 0
```
wongma7: Should there be a port other than random we should default to? Like some convention for all provisioners. Will the port need to be exposed via the pod yaml? Because then a convention would make it easy to distribute yamls with the port already filled in

wongma7 (approved):
lgtm, there is just one question I have about the port

lib/controller/controller.go
cofyc: Though port `0` means picking a random port in network programming, I used it to [disabling metrics server](https://github.com/kubernetes-incubator/external-storage/pull/796/files/58b4e13e3d6da50dfde002534d411e2efd3c3963#diff-16daa5c3f02eee06add7f1af0a3fa901R552) here, like kubelet `--cadvisor-port` flag. Sorry, comments here is a bit confusing.

> Will the port need to be exposed via the pod yaml? Because then a convention would make it easy to distribute yamls with the port already filled in

It depends on provisioner implementation. Provisioner can have a non-zero metrics port and pass it to `NewProvisionController()` to enable metrics server by default.

My intent is to make this feature optional for backward compatibility, each provisioner can customize based on needs, e.g. disable/enable by default, choose default port. We can have some convention for default port (e.g. 8000), but we may need to add another option to toggle feature. 

What do you prefer? Do you want to enable metrics server by default?

wongma7: hmm okay I agree with you actually let's disable it by default

sandaymin123:
@cofyc and @wongma7 
I want to clarify if these metrics for out-of-tree flex drivers are also supported:

```
kubelet_volume_stats_available_bytes
kubelet_volume_stats_capacity_bytes
kubelet_volume_stats_inodes
kubelet_volume_stats_inodes_free
kubelet_volume_stats_inodes_used
kubelet_volume_stats_used_bytes```

wongma7:
@sandaymin123 no, those stats are calculated by kubernetes for in-tree volumes that support the GetMetrics call. We will have to figure out a different solution for out-of-tree.

PR #798, opened 2018-06-13
Add links to user guide and proposal

discordianfish:


wongma7:
/lgtm
/area snapshot
thank you

Issue #795, opened 2018-06-13
Use non-zero resync period as last insurance in local volume provisioner


cofyc:
See discussion: https://github.com/kubernetes-incubator/external-storage/pull/776#issuecomment-393748758

PR #800, opened 2018-06-14
Use non-zero resync period as last insurance in local volume provisioner

cofyc:
Fixes https://github.com/kubernetes-incubator/external-storage/issues/795.

- Add `MinResyncPeriod` configuration, defaults: 5m0s.
- Update JobController to use default resync period.
- Test `TestLoadProvisionerConfigs ` with real cases.
- Add example for customizing resync period.
- Update test/run.sh to fail on errors.

### DIFF
--- a/local-volume/provisioner/cmd/main.go
+++ b/local-volume/provisioner/cmd/main.go
@@ -19,8 +19,10 @@ package main
 import (
 	"flag"
 	"log"
+	"math/rand"
 	"net/http"
 	"os"
+	"time"
 
 	"github.com/golang/glog"
 	"github.com/kubernetes-incubator/external-storage/local-volume/provisioner/pkg/common"
@@ -42,6 +44,7 @@ var (
 )
 
 func main() {
+	rand.Seed(time.Now().UTC().UnixNano())
 	flag.StringVar(&optListenAddress, "listen-address", ":8080", "address on which to expose metrics")
 	flag.StringVar(&optMetricsPath, "metrics-path", "/metrics", "path under which to expose metrics")
 	flag.Set("logtostderr", "true")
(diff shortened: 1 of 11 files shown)


local-volume/provisioner/pkg/controller/controller.go
```diff
@@ -49,6 +50,11 @@ func StartLocalController(client *kubernetes.Clientset, ptable deleter.ProcTable
 	broadcaster.StartRecordingToSink(&v1core.EventSinkImpl{Interface: v1core.New(client.CoreV1().RESTClient()).Events("")})
 	recorder := broadcaster.NewRecorder(scheme.Scheme, v1.EventSource{Component: provisionerName})
 
+	// We choose a random resync period between MinResyncPeriod and 2 *
+	// MinResyncPeriod, so that local provisioners deployed on multiple nodes
+	// at same time don't list the apiserver simultaneously.
+	resyncPeriod := time.Duration(float64(config.MinResyncPeriod.Nanoseconds()) * (1 + rand.Float64()))
```
msau42: The final value should be multiplied by time.Nanosecond.  Also, do we really need Nanosecond granularlity? I think probably Seconds should be sufficient.

cofyc: Fixed in https://github.com/kubernetes-incubator/external-storage/pull/800/commits/fec4ef952a9006121e3b5ac61b3149cb2eb46b38. Also initialized default source in main.go.

local-volume/provisioner/pkg/controller/controller.go
```diff
@@ -53,7 +53,7 @@ func StartLocalController(client *kubernetes.Clientset, ptable deleter.ProcTable
 	// We choose a random resync period between MinResyncPeriod and 2 *
 	// MinResyncPeriod, so that local provisioners deployed on multiple nodes
 	// at same time don't list the apiserver simultaneously.
-	resyncPeriod := time.Duration(float64(config.MinResyncPeriod.Nanoseconds()) * (1 + rand.Float64()))
+	resyncPeriod := time.Duration(config.MinResyncPeriod.Seconds()*(1+rand.Float64())) * time.Second
```
msau42: I think if you don't cast it to float64 like before, then you don't need to multiple by time.Second at the end.

local-volume/provisioner/cmd/main.go
```diff
@@ -43,6 +44,7 @@ var (
 )
 
 func main() {
+	rand.Seed(time.Now().UTC().UnixNano())
```
msau42: This may be causing mac osx build to fail?

cofyc: helm install script, fixed in https://github.com/kubernetes-incubator/external-storage/pull/806.

local-volume/provisioner/pkg/controller/controller.go
cofyc: `Seconds()` method divides it by time.Second to get number of seconds, so we need to multiple by time.Second at the end. If you prefer not to multiple by time.Second, we need to calculate it using nanoseconds. Actually time.Duration is int64 in nanoseconds.

Demo: https://play.golang.org/p/taL1ywwQ1wt

cofyc:
Rebased due to conflicts with https://github.com/kubernetes-incubator/external-storage/pull/799.

Issue #705, opened 2018-04-03
Restore from snapshot fails if PVC is deleted


disrani-px:
After deleting the PVC, restoring from its snapshot fails as it can't find the PVC anymore.

Steps to reproduce:
1. Create PVC
2. Take a snapshot of PVC
3. Delete PVC created in (1)
4. Try to create a clone PVC from snapshot in (2)
5. Create of the PVC fails since the PVC in (1) has been deleted

Logs:
```
[stork-6bd64f8969-9pk7k] I0403 00:11:49.602703       1 leaderelection.go:178] successfully acquired lease to provision for pvc default/mysql-snap-clone 
[stork-6bd64f8969-9pk7k] I0403 00:11:49.602809       1 controller.go:1080] scheduleOperation[provision-default/mysql-snap-clone[9aa99d07-36d3-11e8-98cc-0214683e8447]] 
[stork-6bd64f8969-9pk7k] time="2018-04-03T00:11:49Z" level=info msg="restore from VolumeSnapshotData k8s-volume-snapshot-75178d27-36d3-11e8-b1a0-e24f4dc55876" 
[stork-6bd64f8969-9pk7k] E0403 00:11:49.617504       1 controller.go:808] Failed to provision volume for claim "default/mysql-snap-clone" with StorageClass "stork-snapshot-sc": failed to create a PV from snapshot mysql-snapshot: Failed to retrieve PV pvc-638d7bf1-36d3-11e8-98cc-0214683e8447 from the API server: "persistentvolumes "pvc-638d7bf1-36d3-11e8-98cc-0214683e8447" not found" 
[stork-6bd64f8969-9pk7k] E0403 00:11:49.617805       1 goroutinemap.go:165] Operation for "provision-default/mysql-snap-clone[9aa99d07-36d3-11e8-98cc-0214683e8447]" failed. No retries permitted until 2018-04-03 00:11:50.117768412 +0000 UTC m=+358332.689745478 (durationBeforeRetry 500ms). Error: failed to create a PV from snapshot mysql-snapshot: Failed to retrieve PV pvc-638d7bf1-36d3-11e8-98cc-0214683e8447 from the API server: "persistentvolumes "pvc-638d7bf1-36d3-11e8-98cc-0214683e8447" not found" 
[stork-6bd64f8969-9pk7k] I0403 00:11:51.614633       1 leaderelection.go:198] stopped trying to renew lease to provision for pvc default/mysql-snap-clone, task failed 
```

[...]

PR #803, opened 2018-06-14
Don't check source PV when restoring from snapshot

disrani-px:
The original PV could have been deleted. Use SnapshotData to check
if Restore is supported instead

Issue #705


disrani-px:
OSX build is failing, not related to this change.

tsmetana:
Thanks -- I did a quick test in AWS and it fixes the broken provisioning...

PR #806, opened 2018-06-15
Add osx support for helm installation script to fix travis OSX test

cofyc:
Fixes https://travis-ci.org/kubernetes-incubator/external-storage/jobs/392538641.

I also updated script to install binary in `/usr/local/bin` which is more appropriate to install programs not managed by distribution.


wongma7:
/lgtm
excellent, thank you

PR #807, opened 2018-06-15
EFS should be mounted with nfs options vers=4.1

sp-joseluis-ledesma:
EFS performance will improve by mounting with the vers=4.1 nfs option (https://docs.aws.amazon.com/efs/latest/ug/mounting-fs.html).

### DIFF
--- a/aws/efs/cmd/efs-provisioner/efs-provisioner.go
+++ b/aws/efs/cmd/efs-provisioner/efs-provisioner.go
@@ -171,6 +171,7 @@ func (p *efsProvisioner) Provision(options controller.VolumeOptions) (*v1.Persis
 					ReadOnly: false,
 				},
 			},
+			MountOptions: []string{"vers=4.1"},
 		},
 	}
 	if gidAllocate {


wongma7:
lgtm, but should we also add 
```
rsize=1048576
wsize=1048576
hard
timeo=600
retrans=2
```
while we're here?

sp-joseluis-ledesma:
NFS volumes on kubernetes already have these parameters, this is why I just changed the "vers" one.

wongma7:
ok, sure. please sign the CLA and I will merge

sp-joseluis-ledesma:
Already signed

whereisaaron:
Hi, is there a release with this mounting option change in it that can be used with k8s 1.8 and 1.9?

I note that the current build is suffixed `k8s.1.10` does that mean it is incompatible with earlier versions of k8s? https://quay.io/repository/external_storage/efs-provisioner?tag=latest&tab=tags

Alternatively could we backport it to the previous release?

image

wongma7:
@whereisaaron no, it does not mean it is incompatible. It is only there to indicate what version of the kube API the image has been built against, and the kube API will always be backwards compatible. I put it there to make it easier for users to identify if an image is outdated. Like if a storageclass field is added in 1.11 then it should be obvious that images with the 1.10 suffix won't be able to recognize that. 

Of course if it causes more confusion than it reduces then we should not have the suffix at all. Realistically people may just pull `:latest` and come here when they are confused why it doesn't actually have the latest features. I basically unilaterally decided to add the suffix. (I will also note that the specific case of storageclass fields being added & old versions not being able to discover them is my fault in the first place)

whereisaaron:
Great, thank you for explaining that @wongma7. As a sampling of one, I was immediately confused by it 😄. What was not confusing is that it definitely works with k8s 1.10, so it is clear in that regard certainly.

As you explained, knowing the API it is built against doesn't immediately tell you which other API versions with which it is still compatible. Whether you keep that suffix or not, maybe you could have a simple compatibility table on the [EFS home page](https://github.com/kubernetes-incubator/external-storage/tree/master/aws/efs) similar to that on the [cluster-autoscaler home page](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler#releases)?

The suffix probably makes sense to say, _'this is built for k8s X.Y, it 100% definitely works with k8s X.Y'_, and then the table could identify likely backwards/forward compatible k8s versions, based on your better knowlegde of API changes.

https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler#releases
> # Releases
> 
> We strongly recommend using Cluster Autoscaler with version for which it was meant. We don't
> do ANY cross version testing so if you put the newest Cluster Autoscaler on an old cluster
> there is a big chance that it won't work as expected.
> 
> | Kubernetes Version  | CA Version   |
> |--------|--------|
> | 1.10.X | 1.2.X  |
> | 1.9.X  | 1.1.X  |
> | 1.8.X  | 1.0.X  |
> | 1.7.X  | 0.6.X  |
> | 1.6.X  | 0.5.X, 0.6.X<sup>*</sup>  |
> | 1.5.X  | 0.4.X  |
> | 1.4.X  | 0.3.X  |
> 
> <sup>*</sup>Cluster Autoscaler 0.5.X is the official version shipped with k8s 1.6. We've done some basic tests using k8s 1.6 / CA 0.6 and we're not aware of any problems with this setup. However, CA internally simulates k8s scheduler and using different versions of scheduler code can lead to subtle issues.

wongma7:
I agree with a table, my only real concern is keeping it up-to-date :). I've created a new issue to track this here https://github.com/kubernetes-incubator/external-storage/issues/819 . Thank you for raising this with me, feel free to always...

PR #814, opened 2018-06-16
[standalone-cinder-provisioner] Support for HTTPS endpoints and auth url

zonca:
`alpine` doesn't have certificates needed to connect to a HTTPS auth url or HTTPS endpoints, so I added certificates to the `DockerFile`


zonca:
cla signed

PR #825, opened 2018-06-20
Use rate limited work queues and up resync from 15s to 15m

wongma7:
* Up resync period to 15 minutes
* Ditch goroutinemap and instead use the exponential backoff that comes with rate limited work queues
* Ditch failedProvisionStats and failedDeleteStats, and instead use NumRequeues that comes with rate limited work queues

### DIFF
--- a/lib/controller/controller.go
+++ b/lib/controller/controller.go
@@ -21,7 +21,6 @@ import (
 	"net"
 	"net/http"
 	"os/exec"
-	"reflect"
 	"strconv"
 	"strings"
 	"sync"
@@ -33,12 +32,14 @@ import (
 	rl "github.com/kubernetes-incubator/external-storage/lib/leaderelection/resourcelock"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"golang.org/x/time/rate"
 	"k8s.io/api/core/v1"
 	storage "k8s.io/api/storage/v1"
 	storagebeta "k8s.io/api/storage/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/uuid"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/apimachinery/pkg/watch"
@@ -48,8 +49,8 @@ import (
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/record"
 	ref "k8s.io/client-go/tools/reference"
+	"k8s.io/client-go/util/workqueue"
 	"k8s.io/kubernetes/pkg/apis/core/v1/helper"
-	"k8s.io/kubernetes/pkg/util/goroutinemap"
 	utilversion "k8s.io/kubernetes/pkg/util/version"
 )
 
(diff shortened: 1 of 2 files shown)


wongma7:
@cofyc if you would like to review. @childsb if you would like to also/find somebody else willing to :)

The PR looks big but the code for claim queue and volume queue is pretty much identical. I considered refactoring it so there is less duplicate code but I will let y'all decide if it is okay as-is.

lib/controller/controller.go
```diff
@@ -473,8 +511,8 @@ func NewProvisionController(
 
 	volumeHandler := cache.ResourceEventHandlerFuncs{
 		AddFunc:    nil,
```
cofyc: Why omitting `AddFunc` here? If omitted, volumes will not be queued until first resync happens (15 minutes later).

lib/controller/controller.go
```diff
@@ -568,130 +640,181 @@ func (ctrl *ProvisionController) Run(stopCh <-chan struct{}) {
...
+		var ok bool
+		if key, ok = obj.(string); !ok {
+			ctrl.claimQueue.Forget(obj)
+			utilruntime.HandleError(fmt.Errorf("expected string in workqueue but got %#v", obj))
+			return nil
 		}
+
+		if err := ctrl.syncClaimHandler(key); err != nil {
+			if ctrl.claimQueue.NumRequeues(obj) < ctrl.failedProvisionThreshold {
+				glog.Errorf("retrying syncing claim '%s' because failures %v < threshold %v", key, ctrl.claimQueue.NumRequeues(obj), ctrl.failedProvisionThreshold)
```
cofyc: how about using warning here to distinguish with giving up?

lib/controller/controller.go
```diff
@@ -568,130 +640,181 @@ func (ctrl *ProvisionController) Run(stopCh <-chan struct{}) {
...
+				ctrl.claimQueue.AddRateLimited(obj)
+			} else {
+				glog.Errorf("giving up syncing claim '%s' because failures %v >= threshold %v", key, ctrl.claimQueue.NumRequeues(obj), ctrl.failedProvisionThreshold)
+				// Done but do not Forget: it will not be in the queue but NumRequeues
+				// will be saved until the obj is deleted from kubernetes
+			}
+			return fmt.Errorf("error syncing claim '%s': %s", key, err.Error())
```
cofyc: nit: `%q` is more idiomatic than `'%s'`

lib/controller/controller.go
```diff
@@ -568,130 +640,181 @@ func (ctrl *ProvisionController) Run(stopCh <-chan struct{}) {
...
-				return nil
-			})
+	err := func(obj interface{}) error {
+		defer ctrl.claimQueue.Done(obj)
+		var key string
+		var ok bool
+		if key, ok = obj.(string); !ok {
+			ctrl.claimQueue.Forget(obj)
+			utilruntime.HandleError(fmt.Errorf("expected string in workqueue but got %#v", obj))
+			return nil
```
cofyc: probably we can simply return this error here, because outside function will call `utilruntime.HandleError` if err != nil.

lib/controller/controller.go
wongma7: we only care about volume Updates and not Adds because we are checking if a volume has entered 'Released' state and needs to be deleted

lib/controller/controller.go
wongma7: glog does not give us a Warnf :(

lib/controller/controller.go
wongma7: yeah, replaced everywhere.

lib/controller/controller.go
```diff
@@ -400,15 +429,11 @@ func NewProvisionController(
...
-		runningOperations:             goroutinemap.NewGoRoutineMap(DefaultExponentialBackOffOnError),
+		threadiness:                   DefaultThreadiness,
 		createProvisionedPVRetryCount: DefaultCreateProvisionedPVRetryCount,
 		createProvisionedPVInterval:   DefaultCreateProvisionedPVInterval,
 		failedProvisionThreshold:      DefaultFailedProvisionThreshold,
 		failedDeleteThreshold:         DefaultFailedDeleteThreshold,
-		failedProvisionStats:          make(map[types.UID]int),
-		failedDeleteStats:             make(map[types.UID]int),
-		failedProvisionStatsMutex:     &sync.Mutex{},
-		failedDeleteStatsMutex:        &sync.Mutex{},
```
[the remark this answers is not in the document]
wongma7: done

lib/controller/controller.go
wongma7: done

lib/controller/controller.go
cofyc: it has `Warningf`, see https://godoc.org/github.com/golang/glog#Warningf

lib/controller/controller.go
cofyc: but volumes entered 'Released' state during provisioner is not running (e.g. restarting or migrating to another node) will be delayed 15 minutes to delete.  I'm not sure if this is acceptable. Old controller implementation omitted `Add` handler also, but its resync period is only 15 seconds.

`shouldDelete` checks volume itself, queuing all volumes should not affect performance much IMO.

lib/controller/controller.go
wongma7: autocomplete failed me... done!

also sidenote, logging verbosity is kinda all over the place in general, I will fix it later.

lib/controller/controller.go
wongma7: Nice, you're correct, I've added AddFunc.

cofyc:
LGTM

wongma7:
thank you very much for your review, I'll merge this shortly and continue working on controller improvements.

PR #830, opened 2018-06-26
[v2] Add block volume support check to external provisioners

mkimuram:
This PR add a block volume support check to external provisioners. This PR introduces BlockProvisioner interface and SupportsBlock() to check whether the provisioner supports block volume. They are used by the external provisioner library and if a provisioner doesn't supports block volume, the library output an error and doesn't try provisioning the block volume.
    
With this change, external provisioner that supports block volume is required to implement SupportsBlock() and return true. (Also, it must set volumeMode to pv spec that it returns.)

Specification and implementation have been discussed in below PR.
https://github.com/kubernetes-incubator/external-storage/pull/787

### DIFF
--- a/iscsi/targetd/provisioner/iscsi-provisioner.go
+++ b/iscsi/targetd/provisioner/iscsi-provisioner.go
@@ -475,3 +475,7 @@ func (p *iscsiProvisioner) getConnection() (*jsonrpc2.Client, error) {
 	log.Debugln("targetd client created")
 	return client, nil
 }
+
+func (p *iscsiProvisioner) SupportsBlock() bool {
+	return true
+}
--- a/lib/controller/controller.go
+++ b/lib/controller/controller.go
@@ -30,6 +30,7 @@ import (
 	"github.com/kubernetes-incubator/external-storage/lib/controller/metrics"
 	"github.com/kubernetes-incubator/external-storage/lib/leaderelection"
 	rl "github.com/kubernetes-incubator/external-storage/lib/leaderelection/resourcelock"
+	"github.com/kubernetes-incubator/external-storage/lib/util"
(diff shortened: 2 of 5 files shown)


mkimuram:
@wongma7  

I read through the codes again and come to think that current my code might output unnecessary error messages in multiple external provisioner case.

Correct me if I'm wrong. 
shouldProvision() might be called for all external provisioners before provisioners are identified to be the ones that are specified by storageClass or annotation. And once below check has completed, it is decided that the provisioner is the one that should be used for provision. 

https://github.com/kubernetes-incubator/external-storage/blob/master/lib/controller/controller.go#L844

If above is yes, block-unsupported external provisioners that are not used for provisioning a certain claim will output messages that they don't support block volume, because I put a block volume check code before the check. Therefore, block volume check should be moved to after the check.  

Also, I found below check inside provisionClaimOperation(). If similar issue might happen here, block volume check code should be moved to after this check.
(I guess that this is just for checking misconfiguration, so provisionClaimOperation() won't be called for multiple external provisioner.)

https://github.com/kubernetes-incubator/external-storage/blob/master/lib/controller/controller.go#L1016

I would appreciate it if you could give me an advice on it. I will fix it if needed.

wongma7:
You're right, we need to move the check to inside provisionClaimOperation, after that `provisionerName` check is a good place. I was not thinking very hard when i first reviewed the code so thank you for catching this! `shouldProvision` has a strict meaning, we should not modify it here

mkimuram:
Thank you for your comment. I will re-implement my code to reflect your comment and will ask you for review.

mkimuram:
I've fixed the codes as we discussed. PTAL @wongma7

wongma7:
/lgtm

thank you for the tests as well

PR #832, opened 2018-06-27
Fix error imports using glide up

max8899:
* update-helm-value -> update-helm-values-pre-v2.2.0
* github.com/digitalocean/godo/context -> context


wongma7:
/lgtm
I am confused as to how it passed in the first place but okay!

PR #834, opened 2018-06-27
Propagate MountOptions from StorageClasses to nfs-client-provisioner PVs

pgagnon:
This PR will allow StorageClasses to propagate MountOptions to PVs provisioned by nfs-client-provisioner. This is useful, for example, to specify the NFS protocol version to use to communicate with a server.


pgagnon:
Will resubmit with linked email...

PR #837, opened 2018-07-02
optimize leader election and abuse api server

orainxiong:
1. Optimize leader election from per PV to per instance;
2. Use informer cache rather than talking against api server.

### DIFF
--- a/lib/controller/controller.go
+++ b/lib/controller/controller.go
@@ -25,7 +25,6 @@ import (
 	"time"
 
 	"github.com/golang/glog"
-	"github.com/kubernetes-incubator/external-storage/lib/leaderelection"
 	rl "github.com/kubernetes-incubator/external-storage/lib/leaderelection/resourcelock"
 	"k8s.io/api/core/v1"
 	storage "k8s.io/api/storage/v1"
@@ -126,10 +125,6 @@ type ProvisionController struct {
 	// when multiple controllers are running: they race to lock (lead) every PVC
 	// so that only one calls Provision for it (saving API calls, CPU cycles...)
 	leaseDuration, renewDeadline, retryPeriod, termLimit time.Duration
-	// Map of claim UID to LeaderElector: for checking if this controller
-	// is the leader of a given claim
-	leaderElectors      map[types.UID]*leaderelection.LeaderElector
-	leaderElectorsMutex *sync.Mutex
 
 	hasRun     bool
 	hasRunLock *sync.Mutex
@@ -360,8 +355,6 @@ func NewProvisionController(
 		renewDeadline:                 DefaultRenewDeadline,
 		retryPeriod:                   DefaultRetryPeriod,
 		termLimit:                     DefaultTermLimit,
-		leaderElectors:                make(map[types.UID]*leaderelection.LeaderElector),
-		leaderElectorsMutex:           &sync.Mutex{},
 		hasRun:                        false,
 		hasRunLock:                    &sync.Mutex{},
 	}
@@ -530,23 +523,12 @@ func (ctrl *ProvisionController) addClaim(obj interface{}) {
 	}
 
 	if ctrl.shouldProvision(claim) {
-		ctrl.leaderElectorsMutex.Lock()
-		le, ok := ctrl.leaderElectors[claim.UID]
-		ctrl.leaderElectorsMutex.Unlock()
-		if ok && le.IsLeader() {
-			opName := fmt.Sprintf("provision-%s[%s]", claimToClaimKey(claim), string(claim.UID))
-			ctrl.scheduleOperation(opName, func() error {
-				err := ctrl.provisionClaimOperation(claim)
-				ctrl.updateProvisionStats(claim, err)
-				return err
-			})
-		} else {
-			opName := fmt.Sprintf("lock-provision-%s[%s]", claimToClaimKey(claim), string(claim.UID))
-			ctrl.scheduleOperation(opName, func() error {
-				ctrl.lockProvisionClaimOperation(claim)
-				return nil
-			})
-		}
+		opName := fmt.Sprintf("provision-%s[%s]", claimToClaimKey(claim), string(claim.UID))
+		ctrl.scheduleOperation(opName, func() error {
+			err := ctrl.provisionClaimOperation(claim)
+			ctrl.updateProvisionStats(claim, err)
+			return err
+		})
 	}
 }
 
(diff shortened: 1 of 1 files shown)


orainxiong:
> When 100 PVCs are created at same time, the CSI external-provisioner hammers the kube apiserver with requests and gets throttled, causing all sorts of issues.

more detals : https://github.com/kubernetes-csi/external-provisioner/issues/68

lib/controller/controller.go
```diff
@@ -838,9 +751,13 @@ func (ctrl *ProvisionController) provisionClaimOperation(claim *v1.PersistentVol
 	//  A previous doProvisionClaim may just have finished while we were waiting for
 	//  the locks. Check that PV (with deterministic name) hasn't been provisioned
 	//  yet.
+	nameSpace := claim.GetNamespace()
 	pvName := ctrl.getProvisionedVolumeNameForClaim(claim)
-	volume, err := ctrl.client.CoreV1().PersistentVolumes().Get(pvName, metav1.GetOptions{})
```
wongma7: IMO we should keep this. It is copied from upstream https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/volume/persistentvolume/pv_controller.go#L1390. The idea is to deliberately bypass the cache in case a PV was already created. In this case we are trading a kube API `Get` to avoid an unnecessary storage backend `Provision` which is usually worse.

lib/controller/controller.go
```diff
@@ -1118,10 +1035,25 @@ func (ctrl *ProvisionController) deleteVolumeOperation(volume *v1.PersistentVolu
 	// Our check does not have to be as sophisticated as PV controller's, we can
 	// trust that the PV controller has set the PV to Released/Failed and it's
 	// ours to delete
-	newVolume, err := ctrl.client.CoreV1().PersistentVolumes().Get(volume.Name, metav1.GetOptions{})
```
wongma7: Unlike above, this `Get` I am okay with removing.

lib/controller/controller.go
```diff
@@ -716,75 +698,6 @@ func (ctrl *ProvisionController) shouldDelete(volume *v1.PersistentVolume) bool
 	return true
 }
 
-// lockProvisionClaimOperation wraps provisionClaimOperation. In case other
```
wongma7: The issue if we remove it without a replacement is we will be trading kube API abuse for storage backend API abuse, in case somebody is running more than one provisioner instance (which is way too easy given they can just e.g. change deployment # from 1 to 2). I.e. instead of racing to lock a PVC and spamming the kube API server they will race to talk to the storage backend then race to create a PV, so we need a replacement per-storageclass leader election if we want to remove this.

Also if we get rid of this leader election there is a lot more code that can be removed, which I will be happy to be rid of, but I will do it myself in a subsequent and I would appreciate a review!

lib/controller/controller.go
wongma7: Actually I need to think more about this. Since this `Get` is premised on the cache being stale, and I am not sure if we have the same problem of a stale cache as upstream has.

wongma7: nvm, sorry for noise, I am okay with this change. I cannot think of a scenario where we need this, we are no longer using goroutinemap like upstream is either.

lib/controller/controller.go
wongma7: 1. will be much more difficult since there are many other provisioner besides csi external-provisioner depending on external-storage and they will each be burdened with copying the same complementation, even if the implementation is simple. 2 is better I think. The more opaque all of this is to library consumers the better. I don't like the idea of our little controller depending on some configmap for maintaining leader state but I also don't want to overload storage classes for that purpose like we are overloading pvcs at the moment. I will have more time to think/work on this in the coming days

orainxiong: Agree with you. From the implementation perspective, `external-storage` could copy kind of leader election as existing [`operator`](https://github.com/coreos/etcd-operator/blob/c8f63d508266990a4d20718d94363c30e75e6282/cmd/operator/main.go#L121) does. 

Any new ideas, please let me know. Many thanks

cofyc: hi, in which scenario current `per-PVC` lock is bad? I like `per-PVC` lock idea, because I can simply deploy more provisioners to scale. Each provisioner can work independently.

orainxiong: Each provisioner tries to write a PVC's annotation in order to be a leader. If failed, each one will retry after some time and lead to a race condition. As a matter of fact, the implementation of `per-PVC` lock hinders `external-provisioner` scalable.

wongma7: Thank you @cofyc for chiming in, i was truly ready to dump per-pvc lock assuming nobody uses the feature. :) So we don’t want per-pvc lock because it scales poorly wrt apiserver but we don’t want to switch to per-cluster or per-class because that prevents simple horizontal scaling of provisioners. 

I am still leaning toward increasing the lock granularity. Horizontal scaling is nice to have but in practice I have to wonder how much it helps. Controller-manager is just one instance but it doesn’t choke with many pvcs, why should our little controller be any different.

In the beginning sharding was seen as the ideal solution but I could not find a feasible way to do it. If we are going to be writing to an external endpoint/configmap anyway, perhaps there is something fancier than plain leader election that could enable us to do sharding but I think that is a dangerous idea :)

wongma7: It would also help me to understand how apiserver throttling happens. Is it the burst of x ‘Updates’ at once that kills it? Would jitter or back off help alleviate the problem? I’ll research this tomorrow....

orainxiong: From my point of view, `active-passive` could be more proper in practice than `sharding`.

wongma7: I am not convinced client-side throttling is even a problem anymore after we introduced PVC work queues, since `Provision` typically takes a lot of time. But I am never opposed to adding more construct functions. 

I've opened https://github.com/kubernetes-incubator/external-storage/issues/847 to discuss operator principles in general.

I think we need to bring this up with more people, e.g. in the next sig-storage meeting if there's time, otherwise I will ramble on indecisively forever. We should have this resolved AT LATEST before 1.12 release IMO

wongma7: I'll also write an email to sig-storage google group next week

vladimirvivien: @wongma7 any chance you can join us to discuss this issue ?

wongma7: @vladimirvivien yes, where?

vladimirvivien: @wongma7 
if you can, join the next CSI meeting (Wednesday 10am PST) for a quick report on the stuff you are working (LE/informers).  Here is the link below:

VC on Zoom: https://zoom.us/j/614261834

Notes and agenda doc: https://docs.google.com/document/d/1-WmRYvqw1FREcD1jmZAOjC0jX6Gop8FMOzsdqXHFoT4/edit?usp=sharing

PR #841, opened 2018-07-04
Update CEPH_VERSION to mimic

hswong3i:


wongma7:
/lgtm
/area ceph/cephfs
/area ceph/rbd

This means we should push new images to quay right?

hswong3i:
> This means we should push new images to quay right?

Good idea ;-)

wongma7:
pushed `v1.1.0-k8s1.10`

PR #859, opened 2018-07-16
NFS: Skip deletion if the corresponding directory is not found

w1ndy:
If the volume directory is removed, the scheduled deletion will hang forever because os.Rename always fails. This commit adds a check on the volume directory before deletion, so if the directory is not available, the controller will proceed removing PV entries.


wongma7:
/area nfs-client
/lgtm
thanks!

PR #863, opened 2018-07-19
Able to clean filesystem volume with job

### LINKED ISSUE #737: [local-volume] Provide async fs clean option
Similar to how block cleanup is async, fs cleanup can also be asynchronous for very large filesystems.

/area local-volume
cc @dhirajh

cofyc:
Fixes #737.

- ~~Add `-delete-contents` flag to delete contents in directory~~ (use `/scripts/fsclean.sh` script now)
  - Reuse in-process delete code
- ~~Rename DeviceAnnotation ("device") to VolumePathAnnotation ("volume-path")~~
- Create job to clean filesystem volume too when `useJobForCleaning` is true

Discussion:

- Should we allow user to specify minimum capacity of volumes to clean by jobs? For examples, provisioner only starts cleaner jobs for volumes larger than 1T. For small volumes, in-process will be much quicker.

### DIFF
--- a/local-volume/provisioner/pkg/deleter/jobcontroller.go
+++ b/local-volume/provisioner/pkg/deleter/jobcontroller.go
@@ -253,20 +253,27 @@ func (c *jobController) RemoveJob(pvName string) (CleanupState, *time.Time, erro
 }
 
 // NewCleanupJob creates manifest for a cleaning job.
-func NewCleanupJob(pv *apiv1.PersistentVolume, imageName string, nodeName string, namespace string, blkdevPath string,
-	config common.MountConfig) *batch_v1.Job {
-	// Exec scripts expect devicepath as environment variables.
-	env := []apiv1.EnvVar{{Name: common.LocalPVEnv, Value: blkdevPath}}
+func NewCleanupJob(pv *apiv1.PersistentVolume, volMode apiv1.PersistentVolumeMode, imageName string, nodeName string, namespace string, mountPath string,
+	config common.MountConfig) (*batch_v1.Job, error) {
 	priv := true
 	// Container definition
 	jobContainer := apiv1.Container{
-		Name:    JobContainerName,
-		Image:   imageName,
-		Command: config.BlockCleanerCommand,
-		Env:     env,
+		Name:  JobContainerName,
+		Image: imageName,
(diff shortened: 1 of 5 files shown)


local-volume/provisioner/cmd/main.go
```diff
@@ -39,17 +40,29 @@ import (
...
+	optListenAddress  string
+	optMetricsPath    string
+	optDeleteContents string
 )
 
 func main() {
 	rand.Seed(time.Now().UTC().UnixNano())
 	flag.StringVar(&optListenAddress, "listen-address", ":8080", "address on which to expose metrics")
 	flag.StringVar(&optMetricsPath, "metrics-path", "/metrics", "path under which to expose metrics")
+	flag.StringVar(&optDeleteContents, "delete-contents", "", "delete contents in directory")
```
msau42: It is kind of confusing to reuse the same process to do a different function.  Can we create a new util binary for this?  Or can the job just use a shell command to do the deletion?

local-volume/provisioner/pkg/deleter/jobcontroller.go
```diff
@@ -46,8 +46,8 @@ const (
 	PVLabel = "pv"
 	// PVUuidLabel  is the label name whose value is the pv uuid.
 	PVUuidLabel = "pvuuid"
-	// DeviceAnnotation is the annotation that specifies the device path.
-	DeviceAnnotation = "device"
```
msau42: Changing this could break any monitoring/logging built on top of this

cofyc: Reverted.
I updated it because the name is a bit confusing now.

local-volume/provisioner/cmd/main.go
cofyc: Updated to use shell script, see https://github.com/kubernetes-incubator/external-storage/pull/863/files#diff-ba78747a8f5a1366cfa9a5c8bb9b77a2.

local-volume/provisioner/pkg/deleter/jobcontroller.go
msau42: Yeah it is a little confusing now. Would it make sense to have a new annotation for the filesystem case?

PR #864, opened 2018-07-19
Rename MY_NAMESPACE to JOB_NAMESPACE

cofyc:
I think `JOB_NAMESPACE` is a better name, because it's used to specify namespace to create jobs in, although we use same namespace as provisioner in helm chart by default (hardcoded now but it's possible to make it configurable).

### DIFF
--- a/local-volume/helm/provisioner/templates/provisioner.yaml
+++ b/local-volume/helm/provisioner/templates/provisioner.yaml
@@ -69,7 +69,7 @@ spec:
             valueFrom:
               fieldRef:
                 fieldPath: spec.nodeName
-          - name: MY_NAMESPACE
+          - name: JOB_NAMESPACE
             valueFrom:
               fieldRef:
                 fieldPath: metadata.namespace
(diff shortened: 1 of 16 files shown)


local-volume/helm/provisioner/templates/provisioner.yaml
```diff
@@ -69,7 +69,7 @@ spec:
             valueFrom:
               fieldRef:
                 fieldPath: spec.nodeName
-          - name: MY_NAMESPACE
+          - name: JOB_NAMESPACE
```
msau42: How about PROVISIONER_NAMESPACE?  Both the provisioner and its jobs should run in the same namespace.

cofyc: From [this discussion](https://github.com/kubernetes-incubator/external-storage/pull/711/files#r181179874), I thought we allow admin to specify arbitrary namespace to run jobs in. 

If we require jobs must be run in provisioner namespace, I think `MY_NAMESPACE` is a good name, because `MY` means `PROVISIONER` here and we already use `MY_NODE_NAME` to specify node provisioner run on.

PR #866, opened 2018-07-20
Enabling k8s Deployment ability to control Replica Set's life cycle with selectors.

yevhenvolchenko:
Before:
  Every time you update `specs` of the Deployment, new Replica Set is created by Deployment without deleting an old one.
After: 
  Deployment is able to delete old Replica Set.


wongma7:
/lgtm
we just need a CLA, ty

wongma7:
kubernetes is supposed to default spec.selector for extensions/v1beta1 deployments. I think in the near future we will want to update all yamls to apps/v1 but until then..

Issue #869, opened 2018-07-23
deployment and pod YAML file format error under gluster/file/deploy/ directory


dove-young:
These two YAML file cannot be applied into K8S. The expected container and env variables would be array but they were mistakenly given as map data type

```
pwd
~/external-storage/gluster/file/deploy
kubectl create -f deployment.yaml
error: error validating "deployment.yaml": error validating data: ValidationError(Deployment.spec.template.spec.containers): invalid type for io.k8s.api.core.v1.PodSpec.containers: got "map", expected "array"; if you choose to ignore these errors, turn validation off with --validate=false
```

```
pwd
~/external-storage/gluster/file/deploy
kubectl create -f glusterfile-provisioner-pod.yaml
error: error validating "glusterfile-provisioner-pod.yaml": error validating data: ValidationError(Pod.spec.containers): invalid type for io.k8s.api.core.v1.PodSpec.containers: got "map", expected "array"; if you choose to ignore these errors, turn validation off with --validate=false
```

ahmadkamilmasyhur:
based on `https://kubernetes.io/docs/concepts/workloads/pods/pod-overview/`

as far as I know, the problem is, you expected to make an array using '-' before the word, but you use the name without '-' so that why you got that error
shouldn't it be like this?
```
apiVersion: v1
kind: Pod
metadata: 
  name: glusterfile-provisioner
spec: 
  containers: 
  - name: glusterfile-provisioner
    image: "quay.io/external_storage/glusterfile-provisioner:latest"
    env:
      name: PROVISIONER_NAME
      value: gluster.org/glusterfile
```

PR #870, opened 2018-07-23
deployment and pod YAML file format error under gluster/file/deploy/ directory

dove-young:
This is the PR to address the issue #869  deployment and pod YAML file format error under gluster/file/deploy/ directory. 

These two YAML file cannot be applied into K8S. The expected container and env variables would be array but they were mistakenly given as map data type

```
pwd
~/external-storage/gluster/file/deploy
kubectl create -f deployment.yaml
error: error validating "deployment.yaml": error validating data: ValidationError(Deployment.spec.template.spec.containers): invalid type for io.k8s.api.core.v1.PodSpec.containers: got "map", expected "array"; if you choose to ignore these errors, turn validation off with --validate=false
```

```
pwd
~/external-storage/gluster/file/deploy
kubectl create -f glusterfile-provisioner-pod.yaml
error: error validating "glusterfile-provisioner-pod.yaml": error validating data: ValidationError(Pod.spec.containers): invalid type for io.k8s.api.core.v1.PodSpec.containers: got "map", expected "array"; if you choose to ignore these errors, turn validation off with --validate=false
```

### DIFF
--- a/gluster/file/deploy/deployment.yaml
+++ b/gluster/file/deploy/deployment.yaml
@@ -12,8 +12,8 @@ spec:
         app: glusterfile-provisioner
     spec:
       containers:
-          env:
-              name: PROVISIONER_NAME
+        - env:
+            - name: PROVISIONER_NAME
               value: gluster.org/glusterfile
           image: "quay.io/external_storage/glusterfile-provisioner:latest"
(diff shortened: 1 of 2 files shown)


wongma7:
/lgtm
just need a CLA

Issue #772, opened 2018-05-23
RBD provisioner monitor lookup is wonky


andor44:
The way the RBD provisioner looks up monitors from the storage class has multiple issues. Relevant code is [here](https://github.com/kubernetes-incubator/external-storage/blob/fd1946598363284474dd664cc75eaa905f9c0679/ceph/rbd/pkg/provision/provision.go#L262-L287).

1. First of all, the code has a hardcoded dependency on a `kube-dns` service. This is wrong, because while it is the convention to have such a service it is AFAIK not a requirement. Especially now that CoreDNS is becoming the default (which is also how we discovered these issues).
2. It tries to look up each monitor's address. This in itself has multiple issues:
3. It is not well documented that the list of monitors has to be comma separated, *but also* there must be no spaces between the commas and the elements, otherwise everything but the first element will have a space as its first character, which will make DNS lookup fail. The resulting array elements after the split should be stripped of any whitespace at minimum. E.g. `"mon1,mon2,mon3"` is a-OK, but `"mon1, mon2, mon3"` will cause the elements after the split to be `"mon1"`, `" mon2"` and `" mon3"` (note the whitespace). DNS lookup on those records will fail, i.e. all connections will go to `mon1`.
4. The lookup is forced to be an `A` record lookup. This is incorrect, because it should be perfectly OK to use bare IPs for monitors.
5. It seems that the DNS library used to [perform the lookup](https://github.com/kubernetes-incubator/external-storage/blob/fd1946598363284474dd664cc75eaa905f9c0679/ceph/rbd/pkg/provision/provision.go#L212-L229) will treat an answer with _no answer section_ and _no authority section_ as an error, therefore [this](https://github.com/kubernetes-incubator/external-storage/blob/fd1946598363284474dd664cc75eaa905f9c0679/ceph/rbd/pkg/provision/provision.go#L272-L277) code will be skipped. However, if there *is* an authority section then the answer is considered *not* an error. Therefore, if you try to look up an IP address you'll end up with an empty result list from the lookup, and no monitors will be added.

To summarize: the format of the monitor list is not well-defined and the parsing is fragile. Additionally, the DNS lookup should only be performed if the address given is a hostname and skip on IPs altogether.

mrnicoco:
We could check if the monitor is already an IP address otherwise resolving through DNS. Now we need to create (and maintain) NS A entries ...

A check with something like https://golang.org/pkg/net/#ParseIP

[...]

PR #872, opened 2018-07-23
#772 Add net.ParseIP ip call for lookuphost allowing use of monitors with IP

mrnicoco:


wongma7:
/lgtm
thank you.
we just need a CLA
/area ceph/rbd

PR #877, opened 2018-07-24
Update README in rbd and cephfs

nak3:
Currently there are three issues when walk through the README.

1. cephfs example creates secret in `cephfs` even though [`class.yaml` specified kube-system](https://github.com/kubernetes-incubator/external-storage/blob/83f870173c188fccd8262df05bf64d82ecaa744b/ceph/cephfs/example/class.yaml#L10)
2. `docker run cephfs-provisioner` fails as the image is stored in `quay.io/external_storage`.
3. (Not an issue but) `ceph auth get-key client.xxx` could be used instead of `ceph auth get client.admin` with grep/awk/xargs.

Hence, this patch chagnes to:
- create secret for cephfs in kube-system namespace as `cephfs/example/class.yaml` specified so.
- put `quay.io/external_storage/` in front of image name (cephfs-provisioner and rbd-provisioner)
- use `ceph auth get-key client.xxx` command to get secret

### DIFF
--- a/ceph/cephfs/README.md
+++ b/ceph/cephfs/README.md
@@ -25,9 +25,9 @@ See https://kubernetes.io/.
 * Create a Ceph admin secret
 
 ```bash
-ceph auth get client.admin 2>&1 |grep "key = " |awk '{print  $3'} |xargs echo -n > /tmp/secret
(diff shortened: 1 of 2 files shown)


wongma7:
/lgtm
thank you

Issue #868, opened 2018-07-23
[flex-provisioner]  How can interface from flex-provision.sh?


moonek:
If I print `$*` from the `doprovision()` function, there is only `kubernetes.io/pvOrVolumeName` and `parameters` of `StorageClass`.

Can't get pvc `capacity` information from `doprovision()`?

And Can I set the value of the parameter in the `doprovision()` function and return it?

wongma7:
https://github.com/kubernetes-incubator/external-storage/blob/master/flex/pkg/volume/provision.go#L110 only passes `PVName`, we will need to make it pass capacity as well, manually

For return value, I think the only thing you can do is return some json stdout and change the code to parse it https://github.com/kubernetes-incubator/external-storage/blob/master/flex/pkg/volume/provision.go#L114 .

I have not used flex myself

moonek:
OK, After all, I knew I had to modify the `provision.go`.
I was able to import it in any way I wanted by referring to other sources.

Like this..
```
capacity := volumeOptions.PVC.Spec.Resources.Requests[v1.ResourceName(v1.ResourceStorage)]
requestBytes := capacity.Value()
requestGiB := int(util.RoundUpSize(requestBytes, 1024*1024*1024))
extraOptions["capacity"] = strconv.Itoa(requestGiB) + "G"
```
I felt that I needed to study golang.
Thank you.

PR #880, opened 2018-07-25
flex: Add capacity value to createVolume

moonek:
I think volume `capacity` is the most necessary value for `doProvision()` right now.

`Bytes`, `MiB`, and `GiB` units have been added so that the capacity value can be obtained from the `doProvision()` function. (not a breaking change.)

Related issues. https://github.com/kubernetes-incubator/external-storage/issues/868

This is the tested output.
```
2018-07-25 04:29:47 flex[181]: provision() called: {"kubernetes.io/pvOrVolumeName":"pvc-55160e9c-8fc3-11e8-8cea-525400d87180","requestBytes":"1073741824","requestGiB":"1","requestMiB":"1024"}
```

### DIFF
--- a/flex/pkg/volume/provision.go
+++ b/flex/pkg/volume/provision.go
@@ -24,6 +24,8 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/utils/exec"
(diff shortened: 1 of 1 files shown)


moonek:
I signed a Contributor License Agreement (CLA) now..

moonek:
need to change `import`..

PR #886, opened 2018-07-27
Fix logger calls

alpe:
fixes `Info call has possible formatting directive %s`


wongma7:
/lgtm
/area snapshot
ty

Issue #890, opened 2018-07-30
Failures not saved in queue


cofyc:
```
# grep 'retrying syncing volume "pvc-1de2178f-f67c-11e7-ab1d-6c92bf12a534"' rbd-provisioner.log | tail -n 10
W0730 08:48:06.414673       1 controller.go:726] retrying syncing volume "pvc-1de2178f-f67c-11e7-ab1d-6c92bf12a534" because failures 0 < threshold 15
W0730 08:48:24.964085       1 controller.go:726] retrying syncing volume "pvc-1de2178f-f67c-11e7-ab1d-6c92bf12a534" because failures 0 < threshold 15
W0730 08:48:44.072163       1 controller.go:726] retrying syncing volume "pvc-1de2178f-f67c-11e7-ab1d-6c92bf12a534" because failures 0 < threshold 15
W0730 08:49:03.357006       1 controller.go:726] retrying syncing volume "pvc-1de2178f-f67c-11e7-ab1d-6c92bf12a534" because failures 0 < threshold 15
W0730 08:49:22.107176       1 controller.go:726] retrying syncing volume "pvc-1de2178f-f67c-11e7-ab1d-6c92bf12a534" because failures 0 < threshold 15
W0730 08:49:42.211770       1 controller.go:726] retrying syncing volume "pvc-1de2178f-f67c-11e7-ab1d-6c92bf12a534" because failures 0 < threshold 15
W0730 08:50:00.997637       1 controller.go:726] retrying syncing volume "pvc-1de2178f-f67c-11e7-ab1d-6c92bf12a534" because failures 0 < threshold 15
W0730 08:50:20.125474       1 controller.go:726] retrying syncing volume "pvc-1de2178f-f67c-11e7-ab1d-6c92bf12a534" because failures 0 < threshold 15
W0730 08:50:38.808873       1 controller.go:726] retrying syncing volume "pvc-1de2178f-f67c-11e7-ab1d-6c92bf12a534" because failures 0 < threshold 15
W0730 08:50:57.976285       1 controller.go:726] retrying syncing volume "pvc-1de2178f-f67c-11e7-ab1d-6c92bf12a534" because failures 0 < threshold 15
```

wongma7:
Even when exponentialBackOffOnError is false I guess there should be some other kind of `FailureRateLimiter`

wongma7:
The fix is to make a NewItemExponentialFailureRateLimiter with maxDelay == baseDelay

PR #894, opened 2018-07-30
Fix non-exponentialbackoff having 0 delay, make it 15s

wongma7:
Fixes https://github.com/kubernetes-incubator/external-storage/issues/890 so that even if exponentialBackOff is false we retry every ~15s and respect the threshold

W0730 21:07:41.215547       1 controller.go:688] retrying syncing claim "default/nfs-with-bad-class" because failures 5 < threshold 15
W0730 21:07:58.235704       1 controller.go:688] retrying syncing claim "default/nfs-with-bad-class" because failures 6 < threshold 15
W0730 21:08:15.259908       1 controller.go:688] retrying syncing claim "default/nfs-with-bad-class" because failures 7 < threshold 15

### DIFF
--- a/lib/controller/controller.go
+++ b/lib/controller/controller.go
@@ -457,6 +457,7 @@ func NewProvisionController(
 	)
 	if !controller.exponentialBackOffOnError {
 		ratelimiter = workqueue.NewMaxOfRateLimiter(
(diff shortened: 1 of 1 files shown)


wongma7:
@cofyc fyi

wongma7:
i'll probably release 4.0.1 and rebuild every provisioner i built for 4.0.0 with this change

Issue #879, opened 2018-07-25
StorageClass does not support fstype parameter in snapshot-promoter Provisioner


joshisumit:
Kubernetes v1.9.5

```
$ cat <<EOF | kubectl apply -f -
kind: StorageClass
apiVersion: storage.k8s.io/v1
metadata:
  name: snapshot-promoter
provisioner: volumesnapshot.external-storage.k8s.io/snapshot-promoter
parameters:
  type: gp2
  fsType: xfs
EOF
```

If I add fsType: xfs parameter, I will not be able to run PVC properly.

```
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: busybox-snapshot
  annotations:
    snapshot.alpha.kubernetes.io/snapshot: snapshot-demo
spec:
  accessModes:
  - ReadWriteOnce
  resources:
    requests:
      storage: 10Gi
  storageClassName: snapshot-promoter

```

PVC creation fails with an error:
`Warning  ProvisioningFailed    9s (x4 over 51s)  volumesnapshot.external-storage.k8s.io/snapshot-promoter Failed to provision volume with StorageClass "snapshot-promoter": invalid option "fsType" for volumesnapshot.external-storage.k8s.io/snapshot-promoter`

Default is `ext4` filesystem and it is hardcoded [here](https://github.com/kubernetes-incubator/external-storage/blob/master/snapshot/pkg/volume/awsebs/processor.go#L187)

tsmetana:
I'm not sure it's possible to restore the volume from snapshot and change the filesystem along with it -- so adding fsType parameter support to the snapsht provisioner class might not make sense. However there is definitely a bug here: AWS uses the original volume's filesystem when restoring a volume and since we hard-code "ext4", the new PV fails to mount. I will try to fix this first.

Thanks for reporting.

PR #900, opened 2018-08-02
Snapshot: Store fsType in VolumeSnapshotData for AWS

tsmetana:
When AWS restores snapshot it creates a volume that is formatted exactly as the original. The original volume filesystem has to be stored somewhere so the provisioner can create a PV with proper fsType.

The patch is simple but adds new API attribute to the AWSElasticBlockStoreVolumeSnapshotSource.

Related: issue #879


wongma7:
/lgtm

i guess gce, etc. may be the same story but we can deal with it as it comes up

tsmetana:
@wongma7 thanks. You're right that this is most probably an issue in other environments too, I would like to try them out first to be sure. Also it's an API change so I wanted somebody else's opinion on that. :)

PR #902, opened 2018-08-03
Plumbing topology information through Provision call in external PVController library.

verult:
This is a dependency for adding topology functionality to CSI.

In the future, we'd like to cache Node objects since almost all plugins will go through CSI. For now I hope to get the provisioner interface change in first to unblock CSI work, and open an issue to track the work once this PR is merged.

/cc @wongma7 @msau42 @lichuqiang

### DIFF
--- a/lib/controller/controller.go
+++ b/lib/controller/controller.go
@@ -73,6 +73,10 @@ const annDynamicallyProvisioned = "pv.kubernetes.io/provisioned-by"
 
 const annStorageProvisioner = "volume.beta.kubernetes.io/storage-provisioner"
 
+// This annotation is added to a PVC that has been triggered by scheduler to
+// be dynamically provisioned. Its value is the name of the selected node.
+const annSelectedNode = "volume.alpha.kubernetes.io/selected-node"
+
 // ProvisionController is a controller that provisions PersistentVolumes for
 // PersistentVolumeClaims.
 type ProvisionController struct {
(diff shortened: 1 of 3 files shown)


lichuqiang:
Seems you need go-fmt.
Also, can you add some test cases to validate the change?

lib/controller/controller.go
```diff
@@ -1055,12 +1059,35 @@ func (ctrl *ProvisionController) provisionClaimOperation(claim *v1.PersistentVol
...
+	var allowedTopologies []v1.TopologySelectorTerm
+	if ctrl.kubeVersion.AtLeast(utilversion.MustParseSemantic("v1.11.0")) {
+		// Get SelectedNode
+		if nodeName, ok := claim.Annotations[annSelectedNode]; ok {
+			selectedNode, err = ctrl.client.CoreV1().Nodes().Get(nodeName, metav1.GetOptions{}) // TODO (verult) cache Nodes
+			if err != nil {
+				strerr := fmt.Sprintf("Failed to get target node: %v", err)
+				glog.Infof("unexpected error getting target node %q for claim %q: %v", nodeName, claimToClaimKey(claim), err)
+				ctrl.eventRecorder.Event(claim, v1.EventTypeWarning, "ProvisioningFailed", strerr)
+				return err
```
msau42: can we use the same error message for the Event + return?

lib/controller/controller.go
```diff
@@ -1055,12 +1059,35 @@ func (ctrl *ProvisionController) provisionClaimOperation(claim *v1.PersistentVol
...
 
+	var selectedNode *v1.Node
+	var allowedTopologies []v1.TopologySelectorTerm
+	if ctrl.kubeVersion.AtLeast(utilversion.MustParseSemantic("v1.11.0")) {
+		// Get SelectedNode
+		if nodeName, ok := claim.Annotations[annSelectedNode]; ok {
+			selectedNode, err = ctrl.client.CoreV1().Nodes().Get(nodeName, metav1.GetOptions{}) // TODO (verult) cache Nodes
+			if err != nil {
+				strerr := fmt.Sprintf("Failed to get target node: %v", err)
+				glog.Infof("unexpected error getting target node %q for claim %q: %v", nodeName, claimToClaimKey(claim), err)
```
msau42: Do we need to log the error if it's being returned?

lib/controller/controller.go
verult: The behavior in this block matches the behavior in the internal PVController

lib/controller/controller.go
verult: Same as the other comment

msau42: @wongma7 what do you think of potentially cleaning this up (as a separate PR)?  I think the style of logging an error that you also return is unusual

lib/controller/controller.go
wongma7: agreed let's make error message same for event + return + glog, we do the same weird thing here https://github.com/verult/external-storage/blob/5dee038e0ee3133a8e40fbe39d915cef0c722af2/lib/controller/controller.go#L1102

lib/controller/controller.go
wongma7: yes I will clean it up on next week, the caller of `provisionClaimOperation` should log the error.

upstream is weirder because the error is bubbled up through goroutinemap and such so I imagine there is a logical explanation for the weirdness but at least here we should make it simpler if we can

lib/controller/controller.go
verult: ACK

lib/controller/controller.go
verult: ACK

verult:
Cleaned up errors/logs/events as mentioned above, and added tests

lib/controller/controller.go
```diff
@@ -1055,12 +1059,34 @@ func (ctrl *ProvisionController) provisionClaimOperation(claim *v1.PersistentVol
...
+				err = fmt.Errorf("failed to get target node: %v", err)
+				ctrl.eventRecorder.Event(claim, v1.EventTypeWarning, "ProvisioningFailed", err.Error())
+				return err
+			}
+		}
+
+		// Get AllowedTopologies
+		allowedTopologies, err = ctrl.fetchAllowedTopologies(claimClass)
+		if err != nil {
+			return err
```
msau42: should an event be logged here too?

lib/controller/controller.go
```diff
@@ -1408,6 +1433,25 @@ func (ctrl *ProvisionController) fetchMountOptions(storageClassName string) ([]s
...
+	}
+
+	switch class := classObj.(type) {
+	case *storage.StorageClass:
+		return class.AllowedTopologies, nil
+	case *storagebeta.StorageClass:
+		return class.AllowedTopologies, nil
+	}
+
+	return nil, fmt.Errorf("Cannot convert object to StorageClass: %+v", classObj)
```
msau42: nit: we've been trying to follow the convention of error messages starting with lowercase

lib/controller/controller_test.go
```diff
@@ -238,6 +238,98 @@ func TestController(t *testing.T) {
...
+		},
+	}
+
+	tests := []struct {
+		name           string
+		objs           []runtime.Object
+		expectedParams *provisionParams
+	}{
+		{
+			name: "provision with AllowedTopologies",
```
msau42: also a test case to provision with neither

PR #905, opened 2018-08-08
Adds archiveOnDelete parameter to nfs-client provisioner.

### LINKED ISSUE #596: NFS-Client "reclaimPolicy: Delete" doesn't delete.
I was expecting `reclaimPolicy: Delete` to actually delete the created directory on my nfs server. 

Unless I'm missing something the current behavior for Delete is to actually move the directory to an "archive" directory leaving old data laying around.

Is there a process that comes and cleans up these archive directories?

Would you accept a PR to configure this behavior with a ENV var or something?

gmorse81:
Fixes #596 
Adds archiveOnDelete parameter to nfs-client storageClass. Allows user to disable the archiving functionality. Sets the examples and readme to use this behavior.

If this parameter is not included, the old functionality will still work.

### DIFF
--- a/nfs-client/cmd/nfs-client-provisioner/provisioner.go
+++ b/nfs-client/cmd/nfs-client-provisioner/provisioner.go
@@ -22,11 +22,15 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
+
+	"k8s.io/kubernetes/pkg/apis/core/v1/helper"
 
 	"github.com/golang/glog"
 	"github.com/kubernetes-incubator/external-storage/lib/controller"
 	"k8s.io/api/core/v1"
+	storage "k8s.io/api/storage/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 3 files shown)


nfs-client/cmd/nfs-client-provisioner/provisioner.go
```diff
@@ -100,9 +104,39 @@ func (p *nfsProvisioner) Delete(volume *v1.PersistentVolume) error {
...
+	// Get the storage class for this volume.
+	storageClass, err := p.getClassForVolume(volume)
+	if err != nil {
+		return err
+	}
+	// Determine if the "archiveOnDelete" parameter exists.
+	// If it exists and has a value of false, delete the directory.
+	// Otherwise, archive it.
+	archiveOnDelete, exists := storageClass.Parameters["archiveOnDelete"]
+	if exists && archiveOnDelete == "false" {
```
wongma7: slightly prefer to use strconv.ParseBool here

gmorse81:
@wongma7 i squashed, but also removed an extra blank line that didn't need to be there. the lgtm was removed.

Issue #908, opened 2018-08-09
[nfs] Add support for specifying NFS clients in exports file


krzwalko:
By default nfs-provisioner uses * as NFS clients allowed to mount NFS shares.
For some outside-cluster installation it would be nice to specify NFS clients allowed to mount shares instead of world.

This could be done by adding new option -export-subnet to nfs-provisioner binary (with default value *).

Issue #854, opened 2018-07-11
controller tests are flaky/racy


wongma7:
just in case anybody creates a pr in the next few weeks and tests fail at `+go test ./controller` please ignore it. I introduced the problem with work queues and don't care to fix it because I think there is a high chance it will be obsolete soon anyway

PR #909, opened 2018-08-09
Add export-subnet for limit NFS clients

krzwalko:
This is PR for issue #908.
It adds support for export-subnet parameter to limit exports to particular NFS clients.


nfs/cmd/nfs-provisioner/main.go
```diff
@@ -42,6 +42,7 @@ var (
 	gracePeriod    = flag.Uint("grace-period", 90, "NFS Ganesha grace period to use in seconds, from 0-180. If the server is not expected to survive restarts, i.e. it is running as a pod & its export directory is not persisted, this can be set to 0. Can only be set if both run-server and use-ganesha are true. Default 90.")
 	enableXfsQuota = flag.Bool("enable-xfs-quota", false, "If the provisioner will set xfs quotas for each volume it provisions. Requires that the directory it creates volumes in ('/export') is xfs mounted with option prjquota/pquota, and that it has the privilege to run xfs_quota. Default false.")
 	serverHostname = flag.String("server-hostname", "", "The hostname for the NFS server to export from. Only applicable when running out-of-cluster i.e. it can only be set if either master or kubeconfig are set. If unset, the first IP output by `hostname -i` is used.")
+	exportSubnet   = flag.String("export-subnet", "*", "Subnet for NFS export to allow mount only from")
```
wongma7: we should add a note that it has no effect if *useGanesha is true. Although we could enable it for ganesha as well someday.

Some validation code to check for the conflict "useGanesha is true but exportSubnet is non-default" could also be nice but I don't mind not having it.

krzwalko: Thanks! I added warning when using useGanesha together with exportSubnet, I adjusted also some test cases, however one of test job is failing and I don't know why.

wongma7:
it's probably just a flake https://github.com/kubernetes-incubator/external-storage/issues/854

krzwalko:
Thank you @wongma7!
That's not a big thing, but that was my first contribution to Kubernetes project.

PR #911, opened 2018-08-09
Revamp nfs e2e tests to fork k8s instead of vendoring it and use yamls 

wongma7:
Depends on https://github.com/kubernetes-incubator/external-storage/pull/892

I hope this makes the nfs e2e travis tests more maintainable. They are very useful. The bash script to patch kubernetes is hideous but at least it is short.


wongma7:
travis problems https://www.traviscistatus.com/incidents/pxj3py3xzn14

Issue #921, opened 2018-08-14
Plugin interface contract not fulfilled for SnapshotDelete


princerachit:
The `Plugin` interface in `snapshot/pkg/volume/interface.go` must be implemented by a snapshot plugin.
The interface contains method `SnapshotDelete(*crdv1.VolumeSnapshotDataSource, *v1.PersistentVolume) error` which should be called when snapshot has to be deleted. However, the [snapshotter code](https://github.com/kubernetes-incubator/external-storage/blob/63c3df76fa4beca1f07cb12170e390e8d58211b9/snapshot/pkg/controller/snapshotter/snapshotter.go#L289)  does not honor the contract. ` err := plugin.SnapshotDelete(&source, nil ) `The `*v1.PersistentVolume` object is not being passed to the `SnapshotDelete` method, instead nil is being passed.

PR #922, opened 2018-08-14
Fixes argument passed to `SnapshotDelete` method.

princerachit:
https://github.com/kubernetes-incubator/external-storage/issues/921

Changes:
1. Carved out method `getPVFromName` from `getPVFromVolumeSnapshot`
2. Updated `deleteSnapshot` so that it passes the PV object.

Signed-off-by: Prince Rachit Sinha <prince.rachit@mayadata.io>

### DIFF
--- a/snapshot/pkg/controller/snapshotter/snapshotter.go
+++ b/snapshot/pkg/controller/snapshotter/snapshotter.go
@@ -25,7 +25,7 @@ import (
 	crdv1 "github.com/kubernetes-incubator/external-storage/snapshot/pkg/apis/crd/v1"
 	"github.com/kubernetes-incubator/external-storage/snapshot/pkg/controller/cache"
 	"github.com/kubernetes-incubator/external-storage/snapshot/pkg/volume"
-	v1 "k8s.io/api/core/v1"
+	"k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/uuid"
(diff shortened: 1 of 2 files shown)


princerachit:
Can somebody review this please? @wongma7 @tsmetana

snapshot/pkg/controller/snapshotter/snapshotter.go
```diff
@@ -286,7 +291,11 @@ func (vs *volumeSnapshotter) deleteSnapshot(spec *crdv1.VolumeSnapshotDataSpec)
 		return fmt.Errorf("%s is not supported volume for %#v", volumeType, spec)
 	}
 	source := spec.VolumeSnapshotDataSource
-	err := plugin.SnapshotDelete(&source, nil /* *v1.PersistentVolume */)
+	pv, err := vs.getPVFromName(spec.PersistentVolumeRef.Name)
+	if err != nil {
+		glog.Warningf("failed to retrieve PV %s from the API server: %q", spec.PersistentVolumeRef.Name, err)
```
tsmetana: I'm not quite sure about the warning here: none of the plugins actually require a PV to delete the snapshot. I think it's a good idea to try to get the PV since we have it in the interface (just in the case some plugin needs it in the future). I would leave it up to the plugin to complain about the missing PV eventually.

princerachit: Logic of snapshot should be independent of plugins. That is why I think controller should log its own warnings. Further, it is possible that plugin does not make proper check for nil values. In this case this log would be helpful.

princerachit: If you are not convinced then I can remove this.

tsmetana: Hm... It's not that important and you're right. The API was written to be future-proof but with the CSI on the way we're unlikely to see new plugins added here. So your change is OK for the current state of things. Maybe we should simplify the API and don't require the PV for snapshot deletion -- it would save us some API calls too.

PR #923, opened 2018-08-14
Update clusterrole.yaml events verbs

FerronN:
The current events verbs are not sufficient (kubernetes 1.11.1). I've added watch, list and get to the resource events to get it working again.

### DIFF
--- a/nfs-client/deploy/auth/clusterrole.yaml
+++ b/nfs-client/deploy/auth/clusterrole.yaml
@@ -14,7 +14,7 @@ rules:
     verbs: ["get", "list", "watch"]
   - apiGroups: [""]
     resources: ["events"]
-    verbs: ["create", "update", "patch"]
+    verbs: ["get", "list", "watch", "create", "update", "patch"]
   - apiGroups: [""]
     resources: ["endpoints"]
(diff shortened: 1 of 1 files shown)


wongma7:
Can you try again with `nfs-client-provisioner-v3.0.0-k8s1.11 `? I pushed it today. Apologies, I changed all the yamls on Friday without also pushing images to `latest`, it should be corrected now.

wongma7:
Arguably we should keep the events permissions around anyway, in case people want to downgrade...

msau42:
What I've done in the past is have a Kubernetes compatibility table to say that version A of the provisioner works with >= Kubernetes X. While technically, a newer provisioner version may work against an older Kubernetes version, I just document it as >= to reduce the testing and support explosion.

wongma7:
Yes I need to do that. This case is more like the provisioner becoming incompatible with itself though. i.e. if you are on the `latest` tag, it will need new RBAC rules. And if you use the new RBAC rules but then roll-back to a previous tag, you will need the old RBAC rules. Unsure of best practices here. I'm glad a number of people figured it out themselves to edit the RBAC depending on their situation but they shouldn't have to.

Issue #925, opened 2018-08-15
[nfs-client] Old Volumes Not Being Deleted


Queuecumber:
Since latest pull, my volumes are not being deleted when the claims are deleted with the following error message:

```
W0815 00:22:42.949984       1 controller.go:747] retrying syncing volume "pvc-e3b3ee7f-a01e-11e8-8553-f44d306586f3" because failures 3 &lt; threshold 15
E0815 00:22:42.950044       1 controller.go:762] error syncing volume "pvc-e3b3ee7f-a01e-11e8-8553-f44d306586f3": strconv.ParseBool: parsing "": invalid syntax
I0815 00:22:42.950948       1 event.go:221] Event(v1.ObjectReference{Kind:"PersistentVolume", Namespace:"", Name:"pvc-e3b3ee7f-a01e-11e8-8553-f44d306586f3", UID:"bfc044dc-a020-11e8-8553-f44d306586f3", APIVersion:"v1", ResourceVersion:"419488", FieldPath:""}): type: 'Warning' reason: 'VolumeFailedDelete' strconv.ParseBool: parsing "": invalid syntax
```

There is definitely an issue with the use of ParsePool, I see it quite often in my logs now

wongma7:
ParseBool does not like empty string, I’ll fix tomorrow.

PR #929, opened 2018-08-15
Fix archiveOnDelete parsing

wongma7:
fixes https://github.com/kubernetes-incubator/external-storage/issues/925 (hopefully)

I can't verify this fixes it. e2e tests may be needed in the future; nfs, nfs-client, and flex can definitely be e2e tested. Depends how much we want to abuse travis.


Queuecumber:
Did this change make it to the official image yet? If so then I'm still seeing the problem, if not then I'll standby to test

wongma7:
@Queuecumber it should be out now, try v3.0.1-k8s1.11. Thanks for your patience this past week!

Queuecumber:
Confirmed that this fixes it, thanks for the fast turnaround

Issue #432, opened 2017-11-02
nfs-provisioner: FSAL validation error with cri-containerd


w9n:
I followed the quickstart and get this from ```kubectl describe pvc nfs```

```		
ProvisioningFailed	Failed to provision volume with StorageClass "example-nfs": error creating export for volume: error exporting export block 
EXPORT
{
	Export_Id = 1;
	Path = /export/pvc-e9de1f6b-bf99-11e7-bb4e-0ae1ff7ff0a2;
	Pseudo = /export/pvc-e9de1f6b-bf99-11e7-bb4e-0ae1ff7ff0a2;
	Access_Type = RW;
	Squash = no_root_squash;
	SecType = sys;
	Filesystem_id = 1.1;
	FSAL {
		Name = VFS;
	}
}
: error calling org.ganesha.nfsd.exportmgr.AddExport: 0 export entries in /export/vfs.conf added because (export create, block validation) errors. Details:
Config File (/export/vfs.conf:53): 1 validation errors in block FSAL
Config File (/export/vfs.conf:53): Errors processing block (FSAL)
Config File (/export/vfs.conf:44): 1 validation errors in block EXPORT
Config File (/export/vfs.conf:44): Errors found in configuration block EXPORT

  43m	3m	167	persistentvolume-controller		Normal	ExternalProvisioning	waiting for a volume to be created, either by external provisioner "cluster.local/nfs" or manually created by system administrator
```
and this from /export/ganesha.log
```
02/11/2017 06:49:51 : epoch 59fab538 : nfs-provisioner-7cddd454ff-2mp8s : nfs-ganesha-18[dbus_heartbeat] vfs_get_root_handle :FSAL :MAJ :Could not open VFS mount point /etc/resolv.conf: rc = Not a directory (20)
02/11/2017 06:49:51 : epoch 59fab538 : nfs-provisioner-7cddd454ff-2mp8s : nfs-ganesha-18[dbus_heartbeat] process_claim :FSAL :CRIT :FSAL VFS failed to claim file system /etc/resolv.conf error Not a directory
02/11/2017 06:49:51 : epoch 59fab538 : nfs-provisioner-7cddd454ff-2mp8s : nfs-ganesha-18[dbus_heartbeat] vfs_create_export :FSAL :CRIT :resolve_posix_filesystem(/export/pvc-e9de1f6b-bf99-11e7-bb4e-0ae1ff7ff0a2) returned Not a directory (20)
[...]

PR #930, opened 2018-08-15
Patch ganesha to skip over files when parsing mount table

wongma7:
fixes https://github.com/kubernetes-incubator/external-storage/issues/432

i'll try to get the patch in upstream ganesha as well, it is very small.


wongma7:
verified locally; merging...

PR #940, opened 2018-08-19
make sure all provision() being done sequential and duplicate pv name generated logic in external-provisioner

orainxiong:
As the title said

### DIFF
--- a/lib/controller/controller.go
+++ b/lib/controller/controller.go
@@ -76,6 +76,9 @@ const annStorageProvisioner = "volume.beta.kubernetes.io/storage-provisioner"
 // be dynamically provisioned. Its value is the name of the selected node.
 const annSelectedNode = "volume.alpha.kubernetes.io/selected-node"
 
+// Length in characters for the generated uuid of a created volume
+const volumeNameUUIDLength = 16
+
 // ProvisionController is a controller that provisions PersistentVolumes for
 // PersistentVolumeClaims.
 type ProvisionController struct {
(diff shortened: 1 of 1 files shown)


lib/controller/controller.go
```diff
@@ -1162,7 +1161,7 @@ func (ctrl *ProvisionController) deleteVolumeOperation(volume *v1.PersistentVolu
 // getProvisionedVolumeNameForClaim returns PV.Name for the provisioned volume.
 // The name must be unique.
 func (ctrl *ProvisionController) getProvisionedVolumeNameForClaim(claim *v1.PersistentVolumeClaim) string {
-	return "pvc-" + string(claim.UID)
+	return fmt.Sprintf("%s-%s", "pvc", strings.Replace(string(claim.UID), "-", "", -1)[0:volumeNameUUIDLength])
```
wongma7: we can't change this. Instead, need to expose a parameter

lib/controller/controller.go
```diff
@@ -149,7 +152,7 @@ const (
 	// DefaultResyncPeriod is used when option function ResyncPeriod is omitted
 	DefaultResyncPeriod = 15 * time.Minute
 	// DefaultThreadiness is used when option function Threadiness is omitted
-	DefaultThreadiness = 4
+	DefaultThreadiness = 1
```
wongma7: why?

orainxiong: To ensure there is the only one operation in-flight for a given PVC.

lib/controller/controller.go
orainxiong: OK, got it

lib/controller/controller.go
wongma7: the work queue should take care of it. If you try to add the same PVC while it is in the queue nothing happens.

THe problem is not that more than one operation can be in-flight, it is that an operation can happen immediately after another under certain conditions. We can try to make controller smarter to avoid this (e.g. this is just a dumb example but wait x seconds after successful provision before removing PVC from queue) but technically it will be impossible to avoid entirely since update to a PVC can come from anywhere in the system and if it happens in the window between successful `Provision` and PV controller binding the PVC and PV, we will try to Provision again. Upstream pv controller has the same issue so whatever we want to do we should do it there first. https://github.com/kubernetes-incubator/external-storage/pull/937 should mitigate it.

orainxiong: I've got it. I will close this pr.

PR #947, opened 2018-08-20
Add option to generate local provisioner name only with node name and not uid

### LINKED ISSUE #636: [local-volume] Add option to use node name instead of node uid for provisioner name
Until we get can move away from using non-unique disk naming in GKE, we need an option to make the provisioner name based off of node name instead of node uid, to avoid the scenario outined in kubernetes/kubernetes#51431 where a 2nd PV for the same path on the "same" node name is created.

/assign

msau42:
Helm template and deployment spec updates will be done after this is merged and a new container image is tagged.

Fixes #636


wongma7:
/lgtm
tests pass merging

Issue #955, opened 2018-08-21
nfs e2e statefulset update race


wongma7:
Expected error:
        <*errors.StatusError | 0xc420a693b0>: {
            ErrStatus: {
                TypeMeta: {Kind: "", APIVersion: ""},
                ListMeta: {SelfLink: "", ResourceVersion: "", Continue: ""},
                Status: "Failure",
                Message: "Operation cannot be fulfilled on statefulsets.apps \"nfs-provisioner\": the object has been modified; please apply your changes to the latest version and try again",
                Reason: "Conflict",
                Details: {
                    Name: "nfs-provisioner",
                    Group: "apps",
                    Kind: "statefulsets",
                    UID: "",
                    Causes: nil,
                    RetryAfterSeconds: 0,
                },
                Code: 409,
            },
        }
        Operation cannot be fulfilled on statefulsets.apps "nfs-provisioner": the object has been modified; please apply your changes to the latest version and try again
    not to have occurred
https://travis-ci.org/kubernetes-incubator/external-storage/jobs/418708810

PR #956, opened 2018-08-21
Try to prevent ss Update failure

wongma7:
fix https://github.com/kubernetes-incubator/external-storage/issues/955


wongma7:
/lgtm
we'll see if this fixes it

PR #957, opened 2018-08-21
Make leader-election configurable: default endpoints object namespace to controller's instead of kube-system

wongma7:
Technically this will not break anything. The worst that will happen is users may have useless endpoints objects left over in the namespace `kube-system` when they update to this version of the controller. If they've already given cluster permissions over endpoints, they will not see any disruption. However, the purpose of this change is to make it so that they don't have to give cluster permissions over endpoints in the first place, and ideally they will at some point reduce those permissions to a single provisioner-isolated namespace.


wongma7:
/cc @verult 
leader election enabled by default. Instead of clusterwide endpoint editing privileges, reduce it to a single namespace. The problem is we cannot know what namespace the provisioner serviceaccount will be created in so we need to instruct users to do something like https://github.com/kubernetes-incubator/external-storage/pull/957/commits/9dc43cab82241c9434d7a4f8c267bdbfddd58e4a#diff-b6908537b8e41b48f4e049f946c70aeeR145

wongma7:
@verult yes. same as external-attacher

PR #969, opened 2018-09-05
Fixing documentation to be correct for Kubernetes

flyhard:
Converted from openshift command


wongma7:
/lgtm
/area nfs-client
thanks

PR #962, opened 2018-08-31
fix rbd monitors missing

silenceshell:
close #942 #778

If I set monitors in IP address format like this:

```
 monitors: 192.168.200.15:6789,192.168.200.17:6789,192.168.200.19:6789
```

rbd provisioner will complain `Failed to provision volume for claim "nextcloud/ceph-claim" with StorageClass "dynamic": missing Ceph monitors`.

This is because rbd provisioner treat the monitors always as DNS hostname and try to call `lookuphost` to resolve `mhost`. Because the monitor is an IP so `lookuphost` return an empty iplist array and error is nil. So, no monitor is added to `opt.monitors` and rbd provisioner complains.

```go
	if dnsip != "" {
		var lookup []string
		if lookup, err = lookuphost(mhost, dnsip); err == nil {
			for _, a := range lookup {
				glog.V(1).Infof("adding %+v from mon lookup\n", a)
				opts.monitors = append(opts.monitors, joinHostPort(a, mport))
			}
		} else {
			opts.monitors = append(opts.monitors, joinHostPort(mhost, mport))
		}
	}
```

I add a check `&& net.ParseIP(mhost) == nil` to avoid `lookuphost` for IP address, and it works well.

Signed-off-by: silenceshell <me@ieevee.com>

Issue #942, opened 2018-08-20
missing Ceph monitors


ygqygq2:
I deploy with the latest documentation, and use the example file. The error log like this:

`kubectl logs -f rbd-provisioner-857866b5b7-hwx79`

```
E0820 09:41:25.984983       1 controller.go:722] error syncing claim "default/claim2": failed to provision volume with StorageClass "rbd": failed to get admin secret from ["kube-system"/"ceph-admin-secret"]: secrets "ceph-admin-secret" is forbidden: User "system:serviceaccount:default:rbd-provisioner" cannot get secrets in the namespace "kube-system"
I0820 09:41:25.984999       1 event.go:221] Event(v1.ObjectReference{Kind:"PersistentVolumeClaim", Namespace:"default", Name:"claim2", UID:"4a136790-a45a-11e8-ba76-000c29ea3e30", APIVersion:"v1", ResourceVersion:"4557080", FieldPath:""}): type: 'Warning' reason: 'ProvisioningFailed' failed to provision volume with StorageClass "rbd": failed to get admin secret from ["kube-system"/"ceph-admin-secret"]: secrets "ceph-admin-secret" is forbidden: User "system:serviceaccount:default:rbd-provisioner" cannot get secrets in the namespace "kube-system"
```

wongma7:
Please try these steps https://github.com/kubernetes-incubator/external-storage/blob/master/ceph/rbd/deploy/README.md#install-with-rbac-roles

ygqygq2:
@wongma7 
I used the RBAC, also the error.

ygqygq2:
My env:
CEPH: "luminous"
Kubeadmin: 1.11
System: CentOS7

ygqygq2:
I rebuild the docker image, but have a new problem:

```
I0822 03:53:23.799799       1 controller.go:948] provision "default/rbd-pv-claim2" class "rbd": started
I0822 03:53:23.806644       1 event.go:221] Event(v1.ObjectReference{Kind:"PersistentVolumeClaim", Namespace:"default", Name:"rbd-pv-claim2", UID:"75e6ed6b-a5be-11e8-8a0d-000c29ea3e30", APIVersion:"v1", ResourceVersion:"5044205", FieldPath:""}): type: 'Normal' reason: 'Provisioning' External provisioner is provisioning volume for claim "default/rbd-pv-claim2"
[...]

Issue #778, opened 2018-05-25
rbd-provisioner: missing Ceph monitors


knfoo:
Kubernetes version:
Client Version: v1.10.2
Server Version: v1.10.2

OS version
PRETTY_NAME="Debian GNU/Linux 9 (stretch)"
NAME="Debian GNU/Linux"

Ceph version
Luminous 12.2.5

rbd-provisioner version
quay.io/external_storage/rbd-provisioner     latest              dcc584102b93

storage class:
```apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  annotations:
    storageclass.beta.kubernetes.io/is-default-class: "true"
  creationTimestamp: 2018-05-23T06:39:27Z
  name: dynamic
  resourceVersion: "7297999"
  selfLink: /apis/storage.k8s.io/v1/storageclasses/dynamic
  uid: 0a494817-5e54-11e8-ab89-94c69119d0e9
parameters:
  adminId: admin
  adminSecretName: ceph-secret
  adminSecretNamespace: kube-system
  fsType: ext4
  imageFeatures: layering
  imageFormat: "2"
  monitors: 192.168.200.15:6789,192.168.200.17:6789,192.168.200.19:6789
  pool: ssd
  userId: kube
  userSecretName: ceph-secret-user
provisioner: kubernetes.io/rbd
reclaimPolicy: Delete
volumeBindingMode: Immediate
```
pvc:
```apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  annotations:
    control-plane.alpha.kubernetes.io/leader: '{"holderIdentity":"9d2b5b91-5e54-11e8-901d-7a1ea547e356","leaseDurationSeconds":15,"acquireTime":"2018-05-25T06:48:52Z","renewTime":"2018-05-25T06:51:26Z","leaderTransitions":0}'
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"v1","kind":"PersistentVolumeClaim","metadata":{"annotations":{},"name":"ceph-claim","namespace":"nextcloud"},"spec":{"accessModes":["ReadWriteOnce"],"resources":{"requests":{"storage":"2Gi"}}}}
    volume.beta.kubernetes.io/storage-provisioner: kubernetes.io/rbd
  creationTimestamp: 2018-05-25T06:48:52Z
  finalizers:
  - kubernetes.io/pvc-protection
  name: ceph-claim
  namespace: nextcloud
[...]

PR #977, opened 2018-09-07
Add IP handling to ceph monitor lookup. Closes #772

### LINKED ISSUE #772: RBD provisioner monitor lookup is wonky

joasepul:
Return IP address and avoid hostname lookup when ceph monitors are given as IP addresses.
Closes #772

### DIFF
--- a/ceph/rbd/pkg/provision/provision.go
+++ b/ceph/rbd/pkg/provision/provision.go
@@ -210,6 +210,10 @@ func findDNSIP(p *rbdProvisioner) (dnsip string) {
 
 // Look up hostname in dns server serverip.
 func lookuphost(hostname string, serverip string) (iplist []string, err error) {
+	if net.ParseIP(hostname) != nil {
+		glog.V(4).Infof("detected IP address %q\n", hostname)
+		return append(iplist, hostname), nil
+	}
 	glog.V(4).Infof("lookuphost %q on %q\n", hostname, serverip)
 	m := new(dns.Msg)
 	m.SetQuestion(dns.Fqdn(hostname), dns.TypeA)


wongma7:
I think https://github.com/kubernetes-incubator/external-storage/pull/962 fixed it? Can you check?

joasepul:
Yes that fixes it. Looks like issues #942 #778 #772 are all duplicates!

Issue #975, opened 2018-09-07
How to create a hostPath with specific uid:gid with local-volume-provisioner?


kuznero:
If it is not currently possible through configuration, can you please guide me in the local-volume code base such that I properly determine where I can change it to achieve `chown -R uid:gid $host_path` effect.

msau42:
Hi can you clarify the issue you are seeing?  Is it that you're running the local-volume-provisioner as a non-root user and so it doesn't have permissions to read the discovery directories?

kuznero:
@msau42 no, the scenario is typical in the sense that local-volume-provisioner is actually running as root (which is typical). But as a result it create folders with `root:root` ownership attributes. Now, imagine you are trying to run Elasticsearch cluster that is not running as root, but as `1000:1000`. It will simply be unable to right to the folder with `root:root` attributes.

msau42:
The local volume provisioner does not create folders for use by local PVs.  It only discovers mount points that have been placed underneath the discovery directories.  So it shouldn't matter to the workload what permissions the discovery directory has, only the permissions of the mounts inside them will matter.

kuznero:
It actually does create folders. Tested.

msau42:
Can you give an example your local PV spec and the config of your local provisioner?

kuznero:
PV is not created at the time when `local-volume-provisioner` is deployed. The point was to auto-create required directories before deploying app definitions, including PVs, etc.

kuznero:
Sounds interesting! That was a long path to figure this out :) Is it possible for you to give an example for `hostPath` volume that will create the folder if it does not exist? Last time I tested PV that points to non-existing path it complained.

msau42:
Here's an example of a pod using hostpath volumes:
https://kubernetes.io/docs/concepts/storage/volumes/#example-pod-2

kuznero:
```
In addition to the required path property, user can optionally specify a type for a hostPath volume.
The supported values for field type are: DirectoryOrCreate, ...
```

Excellent! Thanks a lot! Will stop (mis-)using `local-volume-provisioner` for that.

kuznero:
But perhaps I was a bit too fast. Will `hostPath` allow to specify `uid:gid` for the path to be created?

msau42:
No, hostpath volumes do not perform any permissions management.  You can use an init container to set the appropriate permissions as needed.

[...]

PR #979, opened 2018-09-08
Chown mount dirs

kuznero:
Related to #975


kuznero:
Closed as it was clearly a mis-use of local-volume-provisioner. More details can be found in the #975 discussion.

Issue #973, opened 2018-09-07
[local-volume] support fs volumes on block


lichuqiang:
External local provisioner should be aware of the in-tree change of https://github.com/kubernetes/kubernetes/pull/63011.
Desire volume mode does not have to be same to that of volume source.
Users should be able to specify desire volume mode in config (per storageClass).

/assign

/cc @msau42

lichuqiang:
Closing due to #980

PR #980, opened 2018-09-10
[local-volume] support fs volumes on block

lichuqiang:
Address: #973 

TODO: specify fstype when creating PV object. We need to update api (k8s) dependency for the fstype stuff. But I'm not sure how broad the effect would be to other storages.
Maybe we should wait for an official release (1.12?) for update?
Will post a follow-up for the fstype issue when api is ready. And now let's leave it default.

/area local-volume
/kind feature

/assign @msau42


msau42:
cc @wongma7 can we pull k8s v1.12.0-beta.1 tag into external-storage? or do you want to wait for an official release?

local-volume/provisioner/pkg/common/common.go
```diff
@@ -113,6 +116,9 @@ type MountConfig struct {
 	MountDir string `json:"mountDir" yaml:"mountDir"`
 	// The type of block cleaner to use
 	BlockCleanerCommand []string `json:"blockCleanerCommand" yaml:"blockCleanerCommand"`
+	// Desire volume mode of created PersistentVolume object,
+	// default to Filesystem if not specified.
+	DesireVolumeMode string `json:"desireVolumeMode" yaml:"desireVolumeMode"`
```
msau42: I would just call it `VolumeMode`

local-volume/provisioner/pkg/discovery/discovery.go
```diff
@@ -213,10 +213,8 @@ func (d *Discoverer) discoverVolumesAtPath(class string, config common.MountConf
 		pvName := generatePVName(file, d.Node.Name, class)
 		pv, exists := d.Cache.GetPV(pvName)
 		if exists {
-			if volMode == v1.PersistentVolumeBlock && (pv.Spec.VolumeMode == nil ||
-				*pv.Spec.VolumeMode != v1.PersistentVolumeBlock) {
-				errStr := fmt.Sprintf("Incorrect Volume Mode: PV %q (path %q) was not created in block mode. "+
-					"Please check if BlockVolume features gate has been enabled for the cluster.", pvName, filePath)
+			if volMode == v1.PersistentVolumeFilesystem && *pv.Spec.VolumeMode == v1.PersistentVolumeBlock {
```
msau42: Block volume feature is still alpha, so pv.spec.VolumeMode should be nil in non-alpha clusters.

msau42:
Can you also update the helm chart to add this as an option?

lichuqiang:
Comments addressed

local-volume/helm/README.md
```diff
@@ -62,6 +62,7 @@ provisioner chart and their default values.
 | classes.[n].hostDir                    | Path on the host where local volumes of this storage class are mounted under.                         | str      | `-`                                                        |
 | classes.[n].mountDir                   | Optionally specify mount path of local volumes. By default, we use same path as hostDir in container. | str      | `-`                                                        |
 | classes.[n].blockCleanerCommand        | List of command and arguments of block cleaner command.                                               | list     | `-`                                                        |
+| classes.[n].volumeMode                 | Optionally specify volume mode of created PersistentVolume object. By default, we use Filesystem.     | str      | `-`                                                        |
```
msau42: hm this is going to break backwards compatibility if someone using Block feature before upgrades to this newer version.  Block is still alpha though, maybe that's fine.

Can you also add this to CHANGELOG.md?  You can add a new "HEAD" entry that we'll change to a release version later.

local-volume/helm/provisioner/templates/provisioner.yaml
```diff
@@ -32,6 +32,9 @@ data:
        {{- range $val := $classConfig.blockCleanerCommand }}
          - "{{ $val -}}"{{- end}}
        {{- end }}
+       {{- if $classConfig.volumeMode }}
+       volumeMode: {{ $classConfig.volumeMode }}
```
msau42: we have some test case yamls too. Can you add this there? (actually I think the helm tests will fail because the test yamls have not been updated with the new field)

lichuqiang: Aha, I guess the helm tests won't fail since we don't have the field in either source or output files.
But you are right, we need to add the field there for validation. Maybe add it in one of the cases should be enough?

local-volume/helm/README.md
lichuqiang: done

local-volume/helm/provisioner/templates/provisioner.yaml
lichuqiang: done, and I left some of them as is, to validate the volumeMode unset scenario

local-volume/CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+# HEAD
+
+Deployment updates:
+* support fs volumes on block
+  ([#980](https://github.com/kubernetes-incubator/external-storage/pull/980)).
+  The change breaks backwards compatibility: Users must explicitly set volumeMode to "Block" in config if a StorageClass is expected to be used for block volumes.
```
msau42: compatibility for block volumes

lichuqiang:
updated to include the fsType support

gnufied:
@msau42 @wongma7 can we merge this? We sort of need this for better 1.12 support.

PR #982, opened 2018-09-12
pvcnamespace-pvcname-uuid nomenclature pv pvc rbd

xbernpa:
For a more human readable volume names and helps to avoid using commands to link pvc, pv and the rbd volume associated.

Instead of having a different pv, pvc and rbd name, everything is normalized and unique from the uuid generated.

This is based on k8s netapp trident provisioner nomenclature for pv, pvc and nfs volume name.


xbernpa:
Having issues with CLA, will redo PR once situation is resolved

geerlingguy:
@xbernpa - Were you ever able to resolve those issues? This would also be useful in other provisioners, e.g. `nfs-client`.

PR #985, opened 2018-09-13
lib/controller: refactoring using SharedInformers

### LINKED ISSUE #967: ProvisionController: refactoring using SharedInformers
Hi, I'd like to propose refactoring `ProvisionController` using `SharedInformers` totally.

As mentioned in the controller's Guidelines:

> This saves us connections against the API server, duplicate serialization costs server-side, duplicate deserialization costs controller-side, and duplicate caching costs controller-side.

If this sounds reasonable, I'd be happy to work on this.

/cc @wongma7

iawia002:
close #967 

/area lib

/cc @wongma7

### DIFF
--- a/lib/controller/controller.go
+++ b/lib/controller/controller.go
@@ -102,11 +102,11 @@ type ProvisionController struct {
 	// * 1.6: storage classes enter GA
 	kubeVersion *utilversion.Version
 
-	claimInformer  cache.SharedIndexInformer
+	claimInformer  cache.SharedInformer
 	claims         cache.Store
-	volumeInformer cache.SharedIndexInformer
+	volumeInformer cache.SharedInformer
 	volumes        cache.Store
-	classInformer  cache.SharedIndexInformer
+	classInformer  cache.SharedInformer
 	classes        cache.Store
 
 	// To determine if the informer is internal or external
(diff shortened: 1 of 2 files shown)


lib/controller/controller.go
```diff
@@ -332,44 +329,6 @@ func RetryPeriod(retryPeriod time.Duration) func(*ProvisionController) error {
 	}
 }
 
-// ClaimsInformer sets the informer to use for accessing PersistentVolumeClaims.
```
wongma7: we shouldn't remove these. someone out there is using them and anyway it's in the spirit of sharedinformers to allow users to pass them in: theoretically they would pass sharedinformers into this controller + whatever other controllers are also using the sharedinformer and then start the informers whenever they want.
hope that makes sense.
Don't get me wrong, I would like to clean up the code as much as possible but we have to maintain support for the two cases of: 1) user provides their own informers, 2) we create and run them ourselves

iawia002: > whatever other controllers are also using the sharedinformer and then start the informers whenever they want.

I think this is the main reason. Since we all using `SharedInformers`, no matter where we create the informer, we are using the same instance of the cache as everyone else.

lib/controller/controller.go
```diff
@@ -334,39 +333,42 @@ func RetryPeriod(retryPeriod time.Duration) func(*ProvisionController) error {
...
 	}
 }
 
 // ClassesInformer sets the informer to use for accessing StorageClasses.
 // The informer must use the versioned resource appropriate for the Kubernetes cluster version
 // (that is, v1.StorageClass for >= 1.6, and v1beta1.StorageClass for < 1.6).
-// Defaults to using a private (non-shared) informer.
-func ClassesInformer(informer cache.SharedInformer) func(*ProvisionController) error {
+// Defaults to using a internal informer.
+func ClassesInformer(informer cache.SharedIndexInformer) func(*ProvisionController) error {
```
wongma7: Can we keep it cache.SharedInformer since cache.SharedIndexInformer embeds it?

iawia002: done

wongma7:
/lgtm
thank you and ty for your patience with reviews

PR #991, opened 2018-09-18
fix several format issues

iawia002:
The following are pure strings, although the code works now, we don't need a formatted function.

/assign @wongma7


wongma7:
/lgtm
thank you.

PR #995, opened 2018-09-19
Point nfs-client users to Helm and split up yamls

wongma7:


nfs-client/README.md
```diff
@@ -8,6 +8,18 @@
...
 To note again, you must *already* have an NFS Server.
 
+## With Helm
+
+Follow the instructions for the stable helm chart maintained at https://github.com/helm/charts/tree/master/stable/nfs-client-provisioner
+
+The tl;dr is
+
+```bash
+$ helm install stable/nfs-client-provisioner
```
verwilst: This actually will never work since you need to specify a hostname and path by adding "--set nfs.server=x.x.x.x --set nfs.path=/exported/path". You've taken this from the helm chart readme itself, and i will update that one accordingly so it also makes sense. :-)

wongma7: fixed, thanks

verwilst:
lgtm

PR #996, opened 2018-09-20
cephfs-provisioner: Add cluster DNS resolution for Ceph monitors

elemental-lf:
The Ceph RBD provisioner contains code to resolve monitor names via the cluster DNS service so that a Kubernetes host, when trying to provide a volume to a container, doesn't have to participate in the cluster DNS even when the monitor name is cluster internal. This is normally the case for deployments where the monitors are containerized. 
This PR is a port of this feature to the cephfs-provisioner. It requires an additional rule in the ClusterRole to access the cluster DNS service when RBAC is in use (same as rbd-provisioner). If the service lookup fails or the DNS lookup itself is unsuccessful it behaves just like the original code.

### DIFF
--- a/ceph/cephfs/cephfs-provisioner.go
+++ b/ceph/cephfs/cephfs-provisioner.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"flag"
 	"fmt"
+	"net"
 	"os"
 	"os/exec"
 	"strconv"
@@ -28,6 +29,7 @@ import (
 
 	"github.com/golang/glog"
 	"github.com/kubernetes-incubator/external-storage/lib/controller"
+	"github.com/kubernetes-incubator/external-storage/lib/util"
 	"k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/uuid"
@@ -63,6 +65,8 @@ type cephFSProvisioner struct {
 	secretNamespace string
 	// enable PVC quota
 	enableQuota bool
+	// cached IP address of cluster DNS service
+	dnsip string
 }
 
 func newCephFSProvisioner(client kubernetes.Interface, id string, secretNamespace string, enableQuota bool) controller.Provisioner {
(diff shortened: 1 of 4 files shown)


wongma7:
@elemental-lf 
'gofmt -s -w' needs to be run on the following files: 
./ceph/cephfs/cephfs-provisioner.go

thanks for PR, will review shortly

elemental-lf:
Oops, sorry, thought I had already fixed the formatting issue.

ceph/cephfs/cephfs-provisioner.go
```diff
@@ -284,6 +288,73 @@ func (p *cephFSProvisioner) Delete(volume *v1.PersistentVolume) error {
 	return nil
 }
 
+// Look up the cluster dns service by label "coredns", falling back to "kube-dns" if not found
+func findDNSIP(p *cephFSProvisioner) (dnsip string) {
```
cofyc: how about moving these functions into a util package, then we can don't have duplicate code both in rbd/cephfs provisioners? https://github.com/kubernetes-incubator/external-storage/blob/master/ceph/rbd/pkg/provision/provision.go#L193

elemental-lf: @cofyc I could do that. I would suggest to only move ``lookuphost`` (should be renamed to ``lookupHost`` I think), ``splitHostPort`` and ``joinHostPort`` there as findDNSIP takes a provisioner specific struct and would require more changes. What would be the right file to put this? https://github.com/kubernetes-incubator/external-storage/blob/master/lib/util/util.go?

cofyc: From my point, `findDNSIP` can take one argument `client` to get DNS IP, e.g.

```
if p.dnsip == "" {
    p.dnsip = util.FindDNSIP(p.client)
}
```

lib/util/util.go is right file to put these functions.

elemental-lf: Thank you for the hints. I think I've addressed your suggestions.

elemental-lf:
I've updated the PR as per the suggestions and rebased to current master. I've tested both the ``rbd-provisioner`` and the ``cephfs-provisioner`` in my environment (Kubernetes 1.11.3, containerized Ceph originally based on https://github.com/openstack/openstack-helm-infra) and have found no problems so far. Images including these changes are available for testing on docker.io: ``docker.io/elementalnet/cephfs-provisioner:v2.1.0-k8s1.11`` and ``docker.io/elementalnet/rbd-provisioner:v2.1.0-k8s1.11`` (ignore the tags, images are automatically build from the branch referenced in this PR).
Please review again. Thank you!

wongma7:
review comments addressed & lgtm
/lgtm
/approve

PR #1005, opened 2018-09-28
mountOptions support for local storage

lichuqiang:
Address https://github.com/kubernetes/kubernetes/issues/68317
Follow-up https://github.com/kubernetes/kubernetes/pull/69211
let the plugin pass the mountOptions specified in storageclasses into the PVs to make it effective

/area local-volume
/kind feature

/assign @msau42

### DIFF
--- a/local-volume/provisioner/pkg/discovery/discovery.go
+++ b/local-volume/provisioner/pkg/discovery/discovery.go
@@ -168,6 +168,15 @@ func (d *Discoverer) getReclaimPolicyFromStorageClass(name string) (v1.Persisten
 	return v1.PersistentVolumeReclaimDelete, nil
 }
 
+func (d *Discoverer) getMountOptionsFromStorageClass(name string) ([]string, error) {
+	class, err := d.classLister.Get(name)
+	if err != nil {
+		return nil, err
+	}
+
+	return class.MountOptions, nil
+}
+
 func (d *Discoverer) discoverVolumesAtPath(class string, config common.MountConfig) {
 	glog.V(7).Infof("Discovering volumes at hostpath %q, mount path %q for storage class %q", config.HostDir, config.MountDir, class)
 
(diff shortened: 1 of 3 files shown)


local-volume/provisioner/pkg/discovery/discovery.go
```diff
@@ -255,12 +274,16 @@ func (d *Discoverer) discoverVolumesAtPath(class string, config common.MountConf
 				glog.Errorf("Path %q fs stats error: %v", filePath, err)
 				continue
 			}
+			if len(mountOptions) != 0 {
+				glog.Warningf("Path %q is already in filesystem mode, "+
```
msau42: When we do the bind mount for filesystem mode, could you actually specify different mount options?

lichuqiang: Oh, thanks for pointing this out. I was mainly focus on the change this time, and did not think so thoroughly.

Then I wonder why we don't support mount options before, since we can also have the requirement in the bind mount scenario.

lichuqiang: Hmm, I think it's because the mount options here is kind of different from that of others.
It is not used in `MountDevice` or `Provision` phase.

Then I think it worth considering if users do have practical requirement for this, though it's supported to specify mount options in bind mount theoretically.
I see other kinds of storages did not take mount options into account in `SetUp`

msau42: This PR updated all the volume plugins to take the same mount options for both MountDevice and SetUp: https://github.com/kubernetes/kubernetes/pull/68945

gnufied: The fact that all volume plugins do not pass mount options when calling bind was an oversight - https://github.com/kubernetes/kubernetes/pull/68945 which we fixed.

I know it is tricky for local volumes. We know that, file system mount options are ignored by bind mount and only userspace mount options do get applied. local volumes may not need too many userspace mount options so we might be safe.

PR #1020, opened 2018-10-09
Add standard Helm labels

johngmyers:


johngmyers:
Recheck CLA

local-volume/helm/test/expected/baremetal-cleanbyjobs.yaml
```diff
@@ -5,6 +5,10 @@ kind: ConfigMap
 metadata:
   name: local-provisioner-config
   namespace: default
+  labels:
+    heritage: "Tiller"
+    release: "RELEASE-NAME"
+    chart: provisioner-2.0.0
```
msau42: We haven't been keeping the chart version up to date.  Is there a standard practice on helm chart versioning? Should it mirror the container image version?

johngmyers: I'm not sure I'm the best person to ask about "standard" practice. The charts I maintain are for internal microservices so bump version on every build.

Since you bump the container image version in values.yaml, it would make sense to bump the chart version at the same time. The alternative would be to use Chart.AppVersion for the image version and let the chart remain unmodified across several image releases.

msau42: We definitely need to revamp this whole repo to automate the build process better to do things like update the version on every build.  For now, could you update it to 2.3.0? I plan to tag that image version soon.

PR #1023, opened 2018-10-11
tests: use "go lint" from golang.org/x/lint/golint

nixpanic:
Travis currently fails with this error:

    +go get -u github.com/golang/lint/golint
    package github.com/golang/lint/golint: code in directory /home/travis/gopath/src/github.com/golang/lint/golint expects import "golang.org/x/lint/golint"

It seems 'github.com/golang/lint/golint' is the old path, and now
'golang.org/x/lint/golint' should be used.

See-also: golang/lint#415

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -45,7 +45,7 @@ if [ "$TRAVIS_OS_NAME" != "osx" ]; then
 fi
 
 # Install golint, cfssl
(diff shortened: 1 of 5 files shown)


wongma7:
we don't need to edit the files in vendor/ (but it makes no difference since they'll get overwritten anyway)

/lgtm

PR #1033, opened 2018-10-23
Add support for in-use snapshotting of cinder volumes.

drekle:
- Adding failure case if a cloudprovider is misconfigured
- Adding support for openstack in-use snapshotting via annotation.
- Adding docker files which perform the build instead of make to be used with docker hub's automated build.

The goal of this PR is to add an annotation which will allow a force flag to be passed to openstack to create a snapshot of a volume which is in-use.

We are using the added dockerfiles for docker automated builds.  I'm okay removing these if there is issue with them, keeping my fork for the automated builds of the containers.
https://hub.docker.com/r/ciscosso/snapshot-controller/
https://hub.docker.com/r/ciscosso/snapshot-provisioner/

I've changed the behavior of a bad configuration to crash.  If mis-configured the service cannot perform its function, so I believe entering crashloop is a more appropriate way of notifying operators that something is wrong here.

Note: We are also using the kube-controller-manager method for external-storage.  This sub project is valuable enough on its own.

### DIFF
--- a/Dockerfile-controller
+++ b/Dockerfile-controller
@@ -1,0 +1,11 @@
+FROM golang:latest AS golang
+WORKDIR /go/src/github.com/kubernetes-incubator/external-storage
+RUN go get -u github.com/golang/dep/cmd/dep
+COPY . .
+RUN dep ensure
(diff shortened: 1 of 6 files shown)


drekle:
closing in favor of kubernetes-csi work

PR #1022, opened 2018-10-10
Add support for nodeSelector constraint for Helm DaemonSet template

johngmyers:

PR #1039, opened 2018-10-31
Enhancements to helm template

### LINKED ISSUE #946: [local-volume] Enhancements to helm template
* Change DaemonSet template to use apps/v1
* Add Daemonset option for a node selector
* Add script to regenerate test/expected files from examples
* Rename test/expected to generated_examples

/area local-volume

cofyc:
fix #946 
/assign @msau42 

- Add script to regenerate test/expected files from examples
- Rename test/expected to generated_examples
- Able to create namespace. (Create when `createNamespace` is true, I found most helm charts  does this way, create if it's not default/kube-system/kube-public, may override pre-created namespace. On deletion, helm will delete that namespace).
- Change DaemonSet template to use apps/v1
- ~~Add Daemonset option for a node selector~~ (already implemented in https://github.com/kubernetes-incubator/external-storage/pull/1022)


wongma7:
/approve
oops

PR #1040, opened 2018-11-01
Upgrade to use kubernetes common base image.

### LINKED ISSUE #1036: Update local provisioner base image
It's old

cofyc:
Current latest version is `0.4.0`, see https://console.cloud.google.com/gcr/images/google-containers/GLOBAL/debian-base-amd64.
/assign @msau42 
fix #1036

### DIFF
--- a/local-volume/provisioner/Makefile
+++ b/local-volume/provisioner/Makefile
@@ -44,8 +44,9 @@ push: container
 	docker push $(MUTABLE_IMAGE)
 .PHONY: push
 
-test:
+test: container
 	go test ./...
+	docker run --privileged -v $(PWD)/deployment/docker/test.sh:/test.sh --entrypoint bash quay.io/external_storage/local-volume-provisioner:latest /test.sh
 .PHONY: test
 
 clean:
--- a/local-volume/provisioner/deployment/docker/Dockerfile
+++ b/local-volume/provisioner/deployment/docker/Dockerfile
@@ -19,12 +19,11 @@ ADD . .
 RUN cd local-volume/provisioner; \
 	CGO_ENABLED=0 go build -a -ldflags '-extldflags "-static"' -o local-volume-provisioner ./cmd
 
-FROM centos:7
+FROM k8s.gcr.io/debian-base-amd64:0.4.0
 
-RUN yum update -y && \
-    yum install -y e2fsprogs && \
-    yum install -y hdparm && \
-    yum clean all
+RUN clean-install \
+    util-linux \
+    e2fsprogs 
 
 COPY --from=builder /go/src/github.com/kubernetes-incubator/external-storage/local-volume/provisioner/local-volume-provisioner /local-provisioner
 ADD local-volume/provisioner/deployment/docker/scripts /scripts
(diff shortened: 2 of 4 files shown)


cofyc:
Which script uses `hdparm` binary? 
/hold
I plan to add some tests to test these scripts.

msau42:
https://github.com/kubernetes-incubator/external-storage/tree/master/local-volume/provisioner/deployment/docker/scripts

msau42:
They're invoked by the block cleaner config

cofyc:
I've checked all binaries these scripts invoked, but didn't find `hdparm`.  Invoked indirectly by some binary?

cofyc:
@msau42 
I've add a test script to test these scripts. It seems like `hdparm` is not required. 
By the way, `bash` is added, used to run shell scripts.

local-volume/provisioner/deployment/docker/Dockerfile
```diff
@@ -12,12 +12,12 @@
...
+FROM k8s.gcr.io/debian-base-amd64:0.4.0
 
-RUN yum update -y && \
-    yum install -y e2fsprogs && \
-    yum install -y hdparm && \
-    yum clean all
+RUN clean-install \
+    util-linux \
+    e2fsprogs \
+    bash
```
msau42: hmm would prefer to avoid adding bash to the actual shipped image if possible.  Could we make a special image from the prod image just for testing and inject bash there?

local-volume/provisioner/deployment/docker/Dockerfile
```diff
@@ -12,12 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM centos:7
+FROM k8s.gcr.io/debian-base-amd64:0.4.0
 
-RUN yum update -y && \
-    yum install -y e2fsprogs && \
-    yum install -y hdparm && \
```
msau42: cc @dhirajh do you remember why hdparm was needed?

dhirajh: @msau42 hdparm was included in anticipation that someone might want to use ATA secure erase https://ata.wiki.kernel.org/index.php/ATA_Secure_Erase
as an option. However, I didn't have a script for it out of the box.

msau42: thanks, then I think it is fine to remove in this case. We can add it back if we need it.

local-volume/provisioner/deployment/docker/Dockerfile
msau42: Oh, I see, these scripts need to be executed in the container too when you're not using jobs for cleaning.  Then this makes sense. I thought it was only for testing.

local-volume/provisioner/Makefile
```diff
@@ -44,8 +44,9 @@ push: container
 	docker push $(MUTABLE_IMAGE)
 .PHONY: push
 
-test:
+test: container
 	go test ./...
+	docker run --privileged -v $(PWD)/deployment/docker/test.sh:/test.sh --entrypoint bash quay.io/external_storage/local-volume-provisioner:latest /test.sh
```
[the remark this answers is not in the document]
cofyc: `test` depends `container`, so it will build `quay.io/external_storage/local-volume-provisioner:latest` for testing first

cofyc:
cc @wongma7 @jsafrane 
I've [updated root Makefile](https://github.com/kubernetes-incubator/external-storage/pull/1040/files#diff-b67911656ef5d18c4ae36cb6741b7965), could you please approve it?

PR #1042, opened 2018-11-01
Add containerized build process.

### LINKED ISSUE #949: [local-volume] Add containerized build process
/area local-volume

To reduce dependency on local tools and libraries, we should support containerized builds.  Here's an [example](https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/blob/master/Dockerfile)

cofyc:
fix #949

Old `build` target is renamed to `build-local`.

### DIFF
--- a/local-volume/provisioner/Makefile
+++ b/local-volume/provisioner/Makefile
@@ -20,12 +20,26 @@ ifeq ($(VERSION),)
 	VERSION = latest
 endif
 
+ifeq ($(GOVERSION),)
+	GOVERSION = 1.11.1
+endif
+
 IMAGE = $(REGISTRY)local-volume-provisioner:$(VERSION)
 MUTABLE_IMAGE = $(REGISTRY)local-volume-provisioner:latest
 
-all build:
+all: build
+.PHONY: all
+
+build:
+	docker run --rm -w /go/src/github.com/kubernetes-incubator/external-storage/local-volume/provisioner \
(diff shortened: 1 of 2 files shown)


local-volume/provisioner/Makefile
```diff
@@ -20,12 +20,26 @@ ifeq ($(VERSION),)
...
 IMAGE = $(REGISTRY)local-volume-provisioner:$(VERSION)
 MUTABLE_IMAGE = $(REGISTRY)local-volume-provisioner:latest
 
-all build:
+all: build
+.PHONY: all
+
+build:
+	docker run --rm -w /go/src/github.com/kubernetes-incubator/external-storage/local-volume/provisioner \
+		-v $(PWD)/../..:/go/src/github.com/kubernetes-incubator/external-storage -e CGO_ENABLED=0 -e GOOS=linux \
```
msau42: Can we put these steps inside the Dockerfile?  Then all you have to do here is just run docker build.  This makes it easier to integrate into CI systems.

cofyc: Then it's not possible to support `make quick-container`. How about we only support these targets?

- `make all/build`: same as before build on the host locally
- `make container`: build in the container and make a image
- `make quick-container`: removed
- `make push/clean/test`: same as before

cofyc: Updated

PR #1060, opened 2018-11-14
Fill in rbac.yaml with ServiceAccount manifest.

t-sato:
rbac.yaml already has ServiceAccount name (nfs-client-provisioner).
This PR add manifest to create it.


t-sato:
I signed CLA.

wongma7:
I think because the serviceaccount is in deployment.yaml. I guess this is a better place for it. anyway having it both for now is not an issue, will clean it up at some point.
/lgtm
/approve

PR #1087, opened 2018-12-20
Fix namespace and bump to Deployment apps/v1

remche:
- Namespace declaration was not consistent across files.
- Bump to apiVersion: apps/v1 for Deployment


remche:
CLA signed !

PR #1088, opened 2018-12-20
Build efs-provisioner in a multi stage Dockerfile

### LINKED ISSUE #1043: Instructions to build Docker image for AWS EFS ?
The Dockerfile [here](https://github.com/kubernetes-incubator/external-storage/blob/09d7855e68dbfebb47f251845052e09f920b7a55/aws/efs/Dockerfile) is as follows

```
FROM alpine:3.6
RUN apk update --no-cache && apk add ca-certificates
COPY efs-provisioner /
ENTRYPOINT ["/efs-provisioner"]

```

Seems like it's unclear where to run this from. When does the [efs-provisioner.go](https://github.com/kubernetes-incubator/external-storage/blob/09d7855e68dbfebb47f251845052e09f920b7a55/aws/efs/cmd/efs-provisioner/efs-provisioner.go) file even run ? 

I am building a custom docker image to support custom domains for the EFS instead of hardcoded aws assigned domain names for EFS.

Tenzer:
Using a multi stage Dockerfile makes it slightly easier to build the Docker
image, as we then don't rely on any intermediate state during the build process.

This also fixes a minor bug in the `Dockerfile` as `apk update --no-cache` was
previously, which doesn't have any effect. Updating the package repository index
and then telling `apk` to not store the cache means the command doesn't serve
any purpose.

Instructions on how to build the Docker image has been added to the README file
as well.

Fixes #1043.


wongma7:
/lgtm
/approve

thanks

PR #1090, opened 2018-12-30
Make nfs-client ARM deployment consistent with regular deployment.

geerlingguy:
I happen to be using the nfs-client provisioner for a [project](http://www.pidramble.com) which runs Kubernetes on both ARM and AMD64 platforms, so I was trying to see what differences there were in the deployment (since there's one specific for ARM).

A `diff` showed a number of differences, but it seems that they are probably just the result of the two manifests getting out of sync (especially evidenced by the fact that the ARM manifest uses the deprecated `serviceAccount` instead of `serviceAccountName`).

I've made it so the ARM manifest matches the normal one; the only difference is the addition of the `-arm` to the end of the image name (which I've preserved).

Merging this PR will help people be able to see the exact, (currently) single difference between the templates to help them template the manifest or just to understand what's different.

I'm not sure what 'ifs' is, but I figure it's better to be consistent with the primary deployment manifest than have something that makes sense to me :)

### DIFF
--- a/nfs-client/deploy/deployment-arm.yaml
+++ b/nfs-client/deploy/deployment-arm.yaml
@@ -16,7 +16,7 @@ spec:
       labels:
         app: nfs-client-provisioner
     spec:
-      serviceAccount: nfs-client-provisioner
+      serviceAccountName: nfs-client-provisioner
       containers:
         - name: nfs-client-provisioner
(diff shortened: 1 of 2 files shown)


geerlingguy:
I have signed the contributor's agreement.

wongma7:
/lgtm
/approve
thank you. i think "ifs" is just an example directory, but agree let's be consistent to avoid confusion

Issue #1000, opened 2018-09-23
rbd-provisioner: `rbd` command hangs with insufficient configuration, no log output


briantopping:
When the provisioner calls `rbd`, if the configuration is incorrect, the command hangs and no logs are generated. Without logging into the container, it is difficult to determine what is going on. 

It may be helpful if the exec could timeout and report the failure rather than simply hang.

Similarly, at a higher verbosity level, it would be helpful to print the full `rbd` command so that it could easily be tested at the rbd container command line when there are problems.

Zyqsempai:
I can take this one.

PR #1108, opened 2019-01-22
Added timeout for rbd commands execution and improved logging

Zyqsempai:
Added context to control the command execution time.
Added addition log with full command for better debugging and fail reproduction.

Fixed: https://github.com/kubernetes-incubator/external-storage/issues/1000

### DIFF
--- a/ceph/rbd/pkg/provision/rbd_util.go
+++ b/ceph/rbd/pkg/provision/rbd_util.go
@@ -30,12 +30,14 @@ import (
 )
 
 const (
-	imageWatcherStr       = "watcher="
-	defaultCommandTimeOut = 5
+	imageWatcherStr = "watcher="
 )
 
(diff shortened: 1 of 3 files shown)


Zyqsempai:
cc @wongma7

ceph/rbd/pkg/provision/rbd_util.go
```diff
@@ -28,7 +30,8 @@ import (
 )
 
 const (
-	imageWatcherStr = "watcher="
+	imageWatcherStr       = "watcher="
+	defaultCommandTimeOut = 5
```
cofyc: How about to make it configurable? If someone want a shorter or longer timeout.

Zyqsempai: Sounds good. Let me add it.

cofyc:
/lgtm
/approve
Thanks!

Issue #1080, opened 2018-12-12
[rbd] Use PV name to create rbd image


cofyc:
Because pv name are unique too, no need to use another random name.

To keep backward compatibility, add an option for this.

Zyqsempai:
I can take care of this one.

PR #1109, opened 2019-01-24
Added new flag to control image name

Zyqsempai:
Added new bool flag `use-pv-name` to control which name use on creation image step, random or PV name.

Fixed: https://github.com/kubernetes-incubator/external-storage/issues/1080


cofyc:
/lgtm
/approve

This behavior aligns with [ceph-csi behaivor](https://github.com/ceph/ceph-csi/blob/master/pkg/rbd/controllerserver.go#L116-L121).

cofyc:
Thanks for contribution!

PR #1116, opened 2019-01-30
Clarify nfs provisioner quickstart steps regarding RBAC

hkolvenbach:
I followed the NFS provisioner quickstart to deploy it on my cluster, but noticed that applying the necessary roles and rolebindings is not mentioned. (It is described here though: https://github.com/kubernetes-incubator/external-storage/blob/master/nfs/docs/deployment.md#in-kubernetes---deployment-of-1-replica)

I added a few lines regarding RBAC and applying the roles and rolebindings to the quickstart.

### DIFF
--- a/nfs/README.md
+++ b/nfs/README.md
@@ -38,6 +38,15 @@ service "nfs-provisioner" created
 deployment "nfs-provisioner" created
 ```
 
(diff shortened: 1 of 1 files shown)


wongma7:
thank you very much.
/lgtm
/approve

PR #1119, opened 2019-02-05
Use GlusterfsPersistentVolumeSource

jayunit100:
I dep'd the dependencies, and had to do this to make everything compile.  

Without dep, I got other errors b/c of the generic vendor directory at the top of this tree.


jayunit100:
I'm pretty sure we will want to kill the top level vendor directory ?  I may be missing something, but it doesn't look like master builds properly

jayunit100:
fixes #1120

jayunit100:
note i don't really know anything about this repo so if this is a bad idea, LMK :)

jayunit100:
closing, i think i know the real  issue

PR #1130, opened 2019-02-08
fix(cephfs): RBAC: allow to manage secrets

mcluseau:
The ceph provisionner also need to be able to create secrets:

```
cephfs-provisioner.go:186] Cephfs Provisioner: create volume failed, err: secrets is forbidden: User "system:serviceaccount:kube-system:cephfs-provisioner" cannot create resource "secrets" in API group "" in the namespace "x"
event.go:221] Event(v1.ObjectReference{Kind:"PersistentVolumeClaim", Namespace:"x", Name:"y", UID:"z", APIVersion:"v1", ResourceVersion:"t", FieldPath:""}): type: 'Warning' reason: 'ProvisioningFailed' failed to provision volume with StorageClass "shared": failed to create secret
```

### DIFF
--- a/ceph/cephfs/deploy/rbac/clusterrole.yaml
+++ b/ceph/cephfs/deploy/rbac/clusterrole.yaml
@@ -5,7 +5,7 @@ metadata:
   namespace: cephfs
 rules:
   - apiGroups: [""]
-    resources: ["persistentvolumes"]
+    resources: ["persistentvolumes","secrets"]
     verbs: ["get", "list", "watch", "create", "delete"]
   - apiGroups: [""]
     resources: ["persistentvolumeclaims"]


Bruceforce:
What would be the better fix for this? To allow access to secrets like proposed here or to set
```
        env:
        - name: PROVISIONER_SECRET_NAMESPACE
          value: cephfs
```
in https://github.com/kubernetes-incubator/external-storage/blob/master/ceph/cephfs/deploy/rbac/deployment.yaml

and add a sed command to replace this in https://github.com/kubernetes-incubator/external-storage/blob/master/ceph/cephfs/deploy/README.md

wongma7:
it is better to set  PROVISIONER_SECRET_NAMESPACE for security reasons (don't want to give anything clusterwide secret permissions). However this isn't documented anywhere is it?

Bruceforce:
@wongma7 

No it isn't unfortunately. I had to look it up here https://github.com/kubernetes-incubator/external-storage/blob/master/ceph/cephfs/cephfs-provisioner.go#L48

wongma7:
yes, a docs PR would be much appreciated if you don't mind! thanks

Bruceforce:
@wongma7 done ( #1135 ).

Issue #992, opened 2018-09-18
Ceph RBD resize volume


unteem:
Hi
I'm trying to resize a volume provisioned with ceph rbd provisioner.

Here is the error I get:
`Error expanding volume "indie-host/minio-indie-host" of plugin kubernetes.io/rbd : rbd info failed, error: executable file not found in $PATH`

Is it possible to resize a volume when using ceph rbd provisioner ?
kubernetes version: 1.10.3

wongma7:
yes, once the volume is created by the provisioner it is like any rbd volume. That error sounds like the rbd executable is not on the node
/area ceph/rbd

JrCs:
I have this error:
`Warning  VolumeResizeFailed     13s (x4 over 2m)  volume_expand                                                                       Error expanding volume "kube-system/claim1" of plugin kubernetes.io/rbd : failed to get admin secret from ["kube-system"/"ceph-admin-secret"]: failed to get secret from ["kube-system"/"ceph-admin-secret"]
`
The kube-system/ceph-admin-secret secret exist and it is used to create the volume.

What rules to ClusterRole I need to change to allow the expansion of volume ?

This is the current ClusterRole i use:
```
kind: ClusterRole
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  name: rbd-provisioner
  namespace: kube-system
rules:
  - apiGroups: [""]
    resources: ["persistentvolumes"]
    verbs: ["get", "list", "watch", "create", "delete"]
  - apiGroups: [""]
    resources: ["persistentvolumeclaims"]
    verbs: ["get", "list", "watch", "update"]
  - apiGroups: ["storage.k8s.io"]
    resources: ["storageclasses"]
    verbs: ["get", "list", "watch"]
  - apiGroups: [""]
    resources: ["events"]
    verbs: ["list", "watch", "create", "update", "patch"]
  - apiGroups: [""]
    resources: ["services"]
    verbs: ["list", "get"]
  - apiGroups: [""]
    resources: ["secrets"]
    verbs: ["get", "create", "delete"]
  - apiGroups: [""]
    resources: ["endpoints"]
    verbs: ["get", "list", "watch", "create", "update", "patch"]

```

msau42:
I do not believe the external provisioner handles resizing.  It is done by in-tree volume plugin.

ainiml:
[...]

Issue #622, opened 2018-02-25
Not all permissions granted in rbd RBAC deploy


bruce2409:
When using the rbac deployment as stock, you get the following error:

"failed to get admin secret from [\"kube-system\"/\"ceph-admin-secret\"]: secrets \"ceph-admin-secret\" is forbidden: User \"system:serviceaccount:default:rbd-provisioner\" cannot get secrets in the namespace \"kube-system\"".

To fix this I had to edit deploy/rbac/clusterrole.yaml to add:

```yaml
  - apiGroups: [""]
    resources: ["secrets"]
    verbs: ["get","list"]
```

cofyc:
hi, @bruce2409 

It's because rbac files on master branch is for latest rbd-provisioner. The latest rbd-provisioner does not require cluster-level secrets permissions.

jamstar:
where is the latest rdb-provisioner available at?  the latest listed in the deployment file is from 8 months ago.

wongma7:
should be fixed, see #766

PR #1133, opened 2019-02-12
Fix secret type in rbd readme

Paxa:
For rbd, secret type should be `kubernetes.io/rbd` otherwise kube-controller-manager will not find it.
And key should be `key` not `secret`

Example yaml is already correct, but readme was misleading

official secret example:
https://github.com/kubernetes/examples/blob/master/staging/volumes/rbd/secret/ceph-secret.yaml

Related to https://github.com/kubernetes-incubator/external-storage/issues/622 and https://github.com/kubernetes-incubator/external-storage/issues/992

### DIFF
--- a/ceph/rbd/README.md
+++ b/ceph/rbd/README.md
@@ -31,8 +31,8 @@ See https://kubernetes.io/.
 * Create a Ceph admin secret
 
 ```bash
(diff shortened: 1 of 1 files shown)


Paxa:
I already signed cla, should I wait or re-submit PR?

cofyc:
/lgtm
/approve
Thanks!

PR #1147, opened 2019-03-11
Add options to skip custom pvc path creation and to keep created pvc …

midN:
EFS Provisioner is nice, but i wish it had 2 extra options for specific use cases:
1. Do not create Custom PVC-named paths EFS storage
2. Do not delete created/shared paths in EFS storage upon wiping out the PVC

Reasoning:
1. There are cases where you would want to share same EFS path between either multiple namespaces or even clusters or even non-kube resources.
In current state, you cannot do that as the EFS Provisioner creates hardcoded path with unique PVC id.
Example use case is EFS for an App running in cluster A and Jenkins CI/CD process running in cluster B.
I want to be able to mount same EFS volume and PATH to a Jenkins CI/CD run to modify/read/backup some files.
With given change, providing "SKIP_PVC_NAMED_PATH" env variable set to "true" will skip adding unique PVC id to provided path.
You can now mount ROOT EFS path to your PVC.

2. Linked to change/request number 1. Current Provisioner with no ability to control it, deletes created Path in EFS if the PVC get's deleted.
There are cases where i want the folder to persist for debugging purposes.
In other cases, when i mount the PVC to custom/shared path, i don't want it get wiped out once the PVC is poof, gone.

### DIFF
--- a/aws/efs/cmd/efs-provisioner/efs-provisioner.go
+++ b/aws/efs/cmd/efs-provisioner/efs-provisioner.go
@@ -40,17 +40,21 @@ import (
 )
 
 const (
-	provisionerNameKey = "PROVISIONER_NAME"
-	fileSystemIDKey    = "FILE_SYSTEM_ID"
-	awsRegionKey       = "AWS_REGION"
-	dnsNameKey         = "DNS_NAME"
+	provisionerNameKey  = "PROVISIONER_NAME"
+	fileSystemIDKey     = "FILE_SYSTEM_ID"
+	awsRegionKey        = "AWS_REGION"
+	dnsNameKey          = "DNS_NAME"
+	skipPvcNamedPathKey = "SKIP_PVC_NAMED_PATH"
+	keepPvPathKey       = "KEEP_PV_PATH"
 )
 
 type efsProvisioner struct {
-	dnsName    string
-	mountpoint string
-	source     string
-	allocator  gidallocator.Allocator
+	dnsName          string
+	mountpoint       string
+	source           string
+	skipPvcNamedPath string
+	keepPvPath       string
+	allocator        gidallocator.Allocator
 }
 
 // NewEFSProvisioner creates an AWS EFS volume provisioner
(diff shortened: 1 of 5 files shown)


midN:
CLA pls, kthxbye

midN:
CLA pls?

midN:
@wongma7 I don't know what to do at this point.
Linuxfoundation is broken to pieces, email doesn't work, redirects to JIRA, JIRA is dead.

https://jira.linuxfoundation.org/browse/CL-95

Opened 3 days ago, not a single hearbeat.
I can't sign the commit, seems like i'll keep using private fork.

midN:
I have no interest in getting that commited anymore.
9 days, no help on getting the broken CLA fixed.

I'll fork and use custom fork, <3 opensource

amitaymolko:
Very unfortunate this can't be merged in, I would very much like this feature

KevW7:
This is exactly what I'm after too! Will this ever find it's way into the main branch?

midN:
@KevW7 I doubt, looks like doing open source contributions to Google projects is fake news 🤷‍♂

wongma7:
For problem #2 of this PR, you can set the StorageClass reclaimPolicy to Retain instead of Delete.

karngyan:
Hey, @midN  thanks for the PR and https://hub.docker.com/r/m1dn/efs-provisioner.
🤘

mkuendig:
@wongma7 is there a chance that this is considered? We would love this too. Need it for a large nfs installation with kubernetes.

tstjep:
We would love to see this functionality too as it would enable us to create predictable a path structure on the base nfs share.

Current structure:
/{nfsroot}/{namespace}-{pvc-name}-{pv-uuid}
resulting in following path on the nfs share:
/my_nfsroot/mynamespace-mypvc-pvc-a4de6bb2-aad6-48e5-9a5b-60a0f9a7e59c

@wongma7: How would you suggest recreating the PV/PVC pairs that you had provisioned with the efs-provisioner on a fresh new cluster? Because of the uuids I don't see how you would easily automate the manual creation of the pv on the new cluster without exporting these paths somehow to an external location (e.g. the nfs root?)

Also: What's the added value of the uuid part apart from using it in multi-tenant clusters? If I have full control over my k8s deployment I could enforce the uniqueness of these paths on my own?

Thanks for consider this :)

PR #1164, opened 2019-04-25
[efs-provisioner] Update K8s config files

luarx:
Main changes:
- Update PVC **deprecated annotation**
- Update deployment **deprecated apiVersion**
- Fix **omitted namespace for a k8s role**

### DIFF
--- a/aws/efs/deploy/rbac.yaml
+++ b/aws/efs/deploy/rbac.yaml
@@ -34,6 +34,7 @@ kind: Role
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: leader-locking-efs-provisioner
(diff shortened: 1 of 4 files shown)


aws/efs/deploy/rbac.yaml
```diff
@@ -34,6 +34,7 @@ kind: Role
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: leader-locking-efs-provisioner
+  namespace: default
```
wongma7: put the same comment here?

# replace with namespace where provisioner is deployed

wongma7:
/lgtm
needs CLA though

PR #1169, opened 2019-04-30
Allow setting of glusterblock provisioner name using env

### LINKED ISSUE #1168: Unable to change the name of the glusterblock provisioner by setting env PROVISIONER_NAME
It should be possible to start the glusterblock provisioner with the name of user's choice by setting only the env PROVISIONER_NAME. 

Currently, it is does not work.

raghavendra-talur:
Fixes #1168

Signed-off-by: Raghavendra Talur <rtalur@redhat.com>

### DIFF
--- a/gluster/block/cmd/glusterblock-provisioner/glusterblock-provisioner.go
+++ b/gluster/block/cmd/glusterblock-provisioner/glusterblock-provisioner.go
@@ -905,13 +905,11 @@ func main() {
 	provName := provisionerName
 	provEnvName := os.Getenv(provisionerNameKey)
 
-	// Precedence is given for ProvisionerNameKey
-	if provEnvName != "" && *id != "" {
-		provName = provEnvName
-	}
-
-	if provEnvName == "" && *id != "" {
+	// cmd args override env
+	if *id != "" {
 		provName = *id
+	} else if provEnvName != "" {
+		provName = provEnvName
 	}
 
 	clientset, err := kubernetes.NewForConfig(config)


humblec:
As discussed in detail, I think we need to decide on below thing first.  

- Who get the priority if both are set ( ENV or ID)

Based on that we need to change the condition.

phlogistonjohn:
> As discussed in detail in the subjected bz, I think we need to decide on below thing first.
> 
>     * Who get the priority if both are set ( ENV or ID)
> 

Agreed.
> 
> Thats a must thing. current code is done for ENV. afaict. ENV get preference on other settings in general in linux.
> 

The current code's behavior is too confusing to use as a baseline IMO. I'll take your word that's what was intended though.


> Based on that assumption as mentioned in the bugzilla this is my solution.
> #1170
> 
> @phlogistonjohn @raghavendra-talur

I disagree about the precedence though. Funny enough it was a comment of yours elsewhere that made me realize that. When an env var and a cli option both can be used to set the same internal value, I believe it is more typical for the cli option to have precedence. For example, take the virsh command line. It can take the server to connect to via the cli option --connect or an env var. But if both are specified it uses the value passed with --connect. To take something a little closer to home, heketi-cli can read the server url from a var or the --server option. If both are set, then --server's value is used.
(These also makes sense as the "nearest" or most specific to the process being run "wins")

humblec:
>
The current code's behavior is too confusing to use as a baseline IMO. I'll take your word that's what was intended though.
>

It is, so its clearly documented in the current code. :)

>
I disagree about the precedence though. Funny enough it was a comment of yours elsewhere that made me realize that.
>

Oh.. I dont know where/when it happened :)

> When an env var and a cli option both can be used to set the same internal value, I believe it is more typical for the cli option to have precedence. For example, take the virsh command line....
>

The reason for me to keep https://github.com/kubernetes-incubator/external-storage/pull/1170 with preference of ENV is this @phlogistonjohn .. Its kind of a assurance for maintaining backward compatibility.  Since start if both set the preference was given for `ENV`. If we change that, compatibility is broken which I dont want to happen. Thats why I kept https://github.com/kubernetes-incubator/external-storage/pull/1170 as an improvement to handle/respect only `ENV` set scenario. Hopefully it handle all the scenarios.

raghavendra-talur:
> > The current code's behavior is too confusing to use as a baseline IMO. I'll take your word that's what was intended though.
> 
> It is, so its clearly documented in the current code. :)
> 
> > I disagree about the precedence though. Funny enough it was a comment of yours elsewhere that made me realize that.
> 
> Oh.. I dont know where/when it happened :)
> 
> > When an env var and a cli option both can be used to set the same internal value, I believe it is more typical for the cli option to have precedence. For example, take the virsh command line....
> 
> The reason for me to keep #1170 with preference of ENV is this @phlogistonjohn .. Its kind of a assurance for maintaining backward compatibility. Since start if both set, the preference was given for `ENV`. If we change that, compatibility is broken which I dont want to happen. Thats why I kept #1170 as an improvement to handle/respect only `ENV` set scenario. Hopefully it handle all the scenarios.

Can we please have a table with required behavior in either this patch or #1170 ?
Based on the logic illustrated by @humblec if would be :

| ENV set | CLI set | Name of the provisioner|
| ------------- | ------------- | ------------- |
| No | No | Default |
| Yes | No | ENV |
| No | Yes | CLI |
| Yes | Yes | ENV |

phlogistonjohn:
I'm less concerned about backwards compat. here as I doubt many users have tried setting the name of the provisioner or we might have uncovered this issue earlier.
As for @raghavendra-talur  suggestion of a table I suggest:

ENV set | CLI set | Name of the provisioner|
| ------------- | ------------- | ------------- |
| No | No | Default |
| Yes | No | ENV |
| No | Yes | CLI |
| Yes | Yes | CLI |
as this is more typical of other tools.

raghavendra-talur:
I agree with @phlogistonjohn on the preference of command args over ENV because that is the norm in most of the *nix projects. It is very confusing to not to be able to override ENV with a command line.

humblec:
@wongma7 I think you mistakenly placed lgtm here, so removing the same. :) 
 That said, as discussed, to maintain backward compatibility we have to work on solution proposed in 1170.

wongma7:
it was not a mistake. I don't see a conclusion to your discussions so please decide which PR you want to merge

humblec:
Just to make sure  everyone on same page. The lgtm was removed to avoid auto merging. That said, if this solution (auto)merged there is a backward compatibility break which I was trying to avoid, especially when there was no consensus on the solution between proposed solutions in subjected PRs.

PR #1170, opened 2019-05-01
Make env also used while naming provisioner.

### LINKED ISSUE #1168: Unable to change the name of the glusterblock provisioner by setting env PROVISIONER_NAME

humblec:
Fixes #1168

Signed-off-by: hchiramm <hchiramm@redhat.com>

### DIFF
--- a/gluster/block/cmd/glusterblock-provisioner/glusterblock-provisioner.go
+++ b/gluster/block/cmd/glusterblock-provisioner/glusterblock-provisioner.go
@@ -905,12 +905,10 @@ func main() {
 	provName := provisionerName
 	provEnvName := os.Getenv(provisionerNameKey)
 
-	// Precedence is given for ProvisionerNameKey
-	if provEnvName != "" && *id != "" {
+	// Precedence is given for ProvisionerNameKey if both are set
+	if provEnvName != "" && *id != "" || provEnvName != "" {
 		provName = provEnvName
-	}
-
-	if provEnvName == "" && *id != "" {
+	} else if *id != "" {
 		provName = *id
 	}
 


gluster/block/cmd/glusterblock-provisioner/glusterblock-provisioner.go
```diff
@@ -905,12 +905,10 @@ func main() {
 	provName := provisionerName
 	provEnvName := os.Getenv(provisionerNameKey)
 
-	// Precedence is given for ProvisionerNameKey
-	if provEnvName != "" && *id != "" {
+	// Precedence is given for ProvisionerNameKey if both are set
+	if provEnvName != "" && *id != "" || provEnvName != "" {
```
raghavendra-talur: The logic currently reads like this:

```
if (a && b) || a {
  use ENV
} else if b {
  use cli
}
```

If we want to keep logic as in the table 

| ENV set | CLI set | Name of the provisioner|
| ------------- | ------------- | ------------- |
| No | No | Default |
| Yes | No | ENV |
| No | Yes | CLI |
| Yes | Yes | ENV |

Then I suggest we simplify the code as
```
if a {
  use ENV
} else if b {
  use cli
}
```

humblec:
>Then I suggest we simplify the code as
>
@raghavendra-talur  Thats for sure, I deliberately kept that part `( a && b)` for better code reading, not really for logic. but, I can remove it and ONLY keep source code comment if you prefer.  Please let me know your choice.

@phlogistonjohn  @raghavendra-talur  Coming back to the priority, I am still not sure or cant `100%` confirm what should get priority between `cli` and `env` , may be there are different opinions and dont have any writen rule. However I am really worried  about backward compatibility which is a must thing assured in kubernetes. So, lets give preference to `ENV` to make sure we are not breaking.

humblec:
@raghavendra-talur @phlogistonjohn any comments or thoughts ?

humblec:
@raghavendra-talur if we are agreed on preference part, we can change https://github.com/kubernetes-incubator/external-storage/pull/1169 or we can take this PR in. I am fine with any.

humblec:
@wongma7 PTAL.

humblec:
@wongma7 can you please merge this PR ?

PR #1179, opened 2019-06-07
Remove ServiceAccount from deployment yaml files

santhosh-tekuri:
ServiceAccount is already defined in rbac.yaml


wongma7:
/lgtm
thank you

PR #1180, opened 2019-06-12
[rbd-provisioner] Rename `usePVName` as `deterministicNames`

hswong3i:
Follow up for #1109, rename `usePVName` as `deterministicNames`, synchronize the style as cephfs-provisioner.

Moreover, also move the `deterministicNames` as parameters of StorageClass.


wongma7:
this would be a breaking change. It might be okay to have usePVName as an arg AND as a storageclass parameter (with one overriding the other, documented), but replacing usePVName would disrupt current users

PR #1181, opened 2019-06-12
[rbd-provisioner] Code Lint from Tab to Spaces

hswong3i:
Minor code lint, convert the use of tab into spaces x4.


wongma7:
gofmt forces us to use tabs, sorry :/

PR #1182, opened 2019-06-14
Snapshots: add deprecation notice

tsmetana:
We should inform potential users that there is a better solution for volume snapshots now...


wongma7:
/lgtm
whole repo needs one tbh, but for a lot of stuff the CSI replacement is not ready yet.

PR #1190, opened 2019-07-05
[nfs-provisioner] Add error check for endpoints list when validating …

uzuku:
Currently endpoints listing failure is marked as invalid service, which may be confusing when tracking problem.
Return the endpoints listing err as soon as it is captured.


wongma7:
/lgtm
/approve

Thanks, apologies for the months-long delay in review for a nice and simple fix.

PR #1196, opened 2019-07-30
Add ability to build nfs-provisioner for ARM

cryptk:
This PR adds the ability to build nfs-provisioner for ARMv7 based platforms.  The arm build uses an ubuntu base image rather than the fedora image for a few reasons:

- the fedora image was unable to interact with it's package repositories properly in a qemu container (required for cross-building the container)
- Ubuntu 19.04 has an updated nfs-ganesha that is newer than 2.7.1.  There is a comment in the x86_64 Dockerfile referencing some compatibility issues with versions previous to 2.7.1.  This removes the requirement to build nfs-ganesha from source which greatly reduces the time required to build the container.

An image built from this is available at https://hub.docker.com/r/cryptk/nfs-provisioner-arm if you would like to test it (or, ideally, you can use this PR to build your own!)


cryptk:
I have signed the CLA

cryptk:
/unassign @jsafrane

Found a bug with running it in an Ubuntu container... give me a bit to work out a solution

cryptk:
/assign @jsafrane

Sorry for the notification spam... Found an issue caused by a missing mtab file in the ubuntu images... should be resolved now.

cryptk:
@jsafrane I am willing to rebase this, but considering that I submitted it 90 days ago with zero feedback, I am curious if you guys are even interested in having support for building an ARM container in the project.  Can I get some clarification on if this is a feature that you actually want before I bother rebasing?  I would hate to rebase it, remove the stale label, and then just have the PR sit for another 90 days and rot out again.

wongma7:
@cryptk sorry for the long wait. @jsafrane doesn't really (have time to) actively review PRs here, please tag me if you rebase and I'll make sure to look at it.

This is certainly of interest. We already have nfs-client-provisioner and nfs-client-provisioner-arm so there is precedent.

cryptk:
Made some additional changes so that anything that builds this the "old way" will still get the same behavior... just x86_64 binaries/images/etc.  Getting a complete local arm docker container is as simple as running `make container-arm` (assuming you have a functional docker and go toolchain of course).

cryptk:
@wongma7 the only thing that I can think of that would be needed after this would be to update the RELEASE.md file to add/update the instructions for building and pushing an ARM based container.  This may be as simple as changing the instructions to run `make push-all` rather than `make push`.  I figured I would leave handling how the image gets released to the owners since you will be the ones running the release process.

For what it's worth, I've been running this on a k3s install on a RPi4 for several months now with no issues!

wongma7:
/lgtm
/approve

Thanks. 

Yes, in terms of releasing it there is some manual work involved. And it's complicated by the x64 image being in the quay.io/kubernetes_incubator registry which should be deprecated. The ARM one should be in quay.io/external_storage like everything else

PR #1208, opened 2019-08-16
Update nfs to lib 4 and client-go w GetReference fix

wongma7:
step 1 of resolving failing alpha suite https://github.com/kubernetes/kubernetes/issues/81403

Thanks to go modules + submodules, we can now update provisioners 1 by 1 :)


wongma7:
/cc @jsafrane 
I'll probably need to self-lgtm.

msau42:
You're not showing up as an approver?

PR #1217, opened 2019-08-30
Fix NFS e2e to build using build/run.sh

wongma7:


wongma7:
self lgtm to unblock testing

PR #1226, opened 2019-09-25
Add iSCSI FlexVol support.

anusha-ragunathan:
Background:
The external targetd provisioner allocates targetd based LUNs. These LUNs are eventually mounted into pods by the kubelet. The in-tree iSCSI support in the kubelet allows for mounting the iSCSI LUNs into pods.

Note that Kubelet's iSCSI in-tree support (discovery, attach, mount) is available on Linux workers only. Although the Windows Kubelet does not provide in-tree iSCSI support, mounting of iSCSI volumes can be accomplished using Windows Flexvolume plugins. https://github.com/microsoft/K8s-Storage-Plugins

Proposed change:
LUNs provisioned by the targetd provisioner can be mounted by Windows iSCSI FlexVolume plugins by creating FlexPersistentVolumeSource for PV Source. This allows Windows pods to use:
    - targetd external provisioner for provisioning.
    - windows flexvol plugins for mounting.

Sample yamls:
    1. iscsi/targetd/kubernetes/iscsi-provisioner-d-win.yaml
        note that deploying this on a mixed cluster of Windows and Linux workers
        requires to qualify that the provisioner runs on Linux only.
        So set nodeSelector to be "beta.kubernetes.io/os: linux"
        TODO: provisioner image is set to aragunathan/iscsi-controller:latest.
        Once this PR merges and official images are built, I'll send a follow on
        PR to update yaml with official image.
    2. iscsi/targetd/kubernetes/iscsi-provisioner-flexvol-class.yaml
  	note "fsType: ntfs" for Windows mounts/workloads.
    3. iscsi/targetd/kubernetes/iscsi-provisioner-pvc-win.yaml
	note that os: "windows" annotation should be added for PVCs.

    Testing:
    Tested on a kubernetes 1.15.0 cluster.
    - Setup targetd server

    - Setup flexvol scripts on Windows workers per https://github.com/microsoft/K8s-Storage-Plugins
    - Using powershell, get the windows worker's iSCSI IQN:
      (Get-InitiatorPort).NodeAddress
      This will be added to the storage class yaml

    - Setup targetd account and secret
    - Deployed storage class: iscsi-provisioner-flexvol-class.yaml
    - Deployed PVC: iscsi-provisioner-pvc-win.yaml
    - Created a Pod to use the claim and observed LUNs mounted successfully.

Signed-off-by: Anusha Ragunathan <anusha.ragunathan@docker.com>

### DIFF
--- a/iscsi/targetd/kubernetes/iscsi-provisioner-pvc-win.yaml
+++ b/iscsi/targetd/kubernetes/iscsi-provisioner-pvc-win.yaml
@@ -3,9 +3,9 @@ apiVersion: v1
 metadata:
   name: myclaim
   annotations:
-    volume.beta.kubernetes.io/storage-class: "iscsi-targetd-vg-targetd"
     os: "windows"
 spec:
+  storageClassName: "iscsi-targetd-vg-targetd"
   accessModes:
     - ReadWriteOnce
   resources:
--- a/iscsi/targetd/kubernetes/iscsi-provisioner-pvc-win.yaml
+++ b/iscsi/targetd/kubernetes/iscsi-provisioner-pvc-win.yaml
@@ -1,0 +1,14 @@
+kind: PersistentVolumeClaim
+apiVersion: v1
+metadata:
+  name: myclaim
+  annotations:
+    volume.beta.kubernetes.io/storage-class: "iscsi-targetd-vg-targetd"
+    os: "windows"
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 100Mi
+
--- a/iscsi/targetd/provisioner/iscsi-provisioner.go
+++ b/iscsi/targetd/provisioner/iscsi-provisioner.go
@@ -108,6 +108,40 @@ func (p *iscsiProvisioner) getAccessModes() []v1.PersistentVolumeAccessMode {
 	}
 }
 
+func createPVSource(options controller.VolumeOptions, portals []string, lun int32) v1.PersistentVolumeSource {
+	if options.PVC.Annotations["os"] == "windows" {
+		return v1.PersistentVolumeSource{
+			FlexVolume: &v1.FlexPersistentVolumeSource{
+				Driver: "microsoft.com/iscsi.cmd",
+				FSType: getFsType(options.Parameters["fsType"]),
+				Options: map[string]string{
+					"chapAuthDiscovery": options.Parameters["chapAuthDiscovery"],
+					"chapAuthSession":   options.Parameters["chapAuthSession"],
+					"iqn":               options.Parameters["iqn"],
+					"lun":               strconv.FormatInt(int64(lun), 10),
+					"targetPortal":      options.Parameters["targetPortal"],
+				},
+				ReadOnly: getReadOnly(options.Parameters["readonly"]),
+			},
+		}
+	}
+
+	return v1.PersistentVolumeSource{
+		ISCSI: &v1.ISCSIPersistentVolumeSource{
+			TargetPortal:      options.Parameters["targetPortal"],
+			Portals:           portals,
+			IQN:               options.Parameters["iqn"],
+			ISCSIInterface:    options.Parameters["iscsiInterface"],
+			Lun:               lun,
+			ReadOnly:          getReadOnly(options.Parameters["readonly"]),
+			FSType:            getFsType(options.Parameters["fsType"]),
+			DiscoveryCHAPAuth: getBool(options.Parameters["chapAuthDiscovery"]),
+			SessionCHAPAuth:   getBool(options.Parameters["chapAuthSession"]),
+			SecretRef:         getSecretRef(getBool(options.Parameters["chapAuthDiscovery"]), getBool(options.Parameters["chapAuthSession"]), &v1.SecretReference{Name: viper.GetString("provisioner-name") + "-chap-secret"}),
+		},
+	}
+}
+
 // Provision creates a storage asset and returns a PV object representing it.
 func (p *iscsiProvisioner) Provision(options controller.VolumeOptions) (*v1.PersistentVolume, error) {
 	if !util.AccessModesContainedInAll(p.getAccessModes(), options.PVC.Spec.AccessModes) {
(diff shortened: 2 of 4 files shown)


anusha-ragunathan:
cc @ddebroy

iscsi/targetd/provisioner/iscsi-provisioner.go
```diff
@@ -108,6 +108,40 @@ func (p *iscsiProvisioner) getAccessModes() []v1.PersistentVolumeAccessMode {
 	}
 }
 
+func createPVSource(options controller.VolumeOptions, portals []string, lun int32) v1.PersistentVolumeSource {
+	if options.PVC.Annotations["os"] == "windows" {
```
ddebroy: I think this merits some discussion: 
1. For multi-OS support from a single provisioner, is the recommendation to use an annotation on the PVC? If so maybe the annotation can be standardized with appropriate prefixes. Advantage of this is you get to keep a single storage class and service PVCs from that.
2. An alternative could be to have a storage class per OS and pass the OS a parameter in the storage class that the external provisioner processes.
Thoughts? @jsafrane @jingxu97

ddebroy (commented):
Overall looks good. One question that I think needs to be discussed a bit.

anusha-ragunathan:
Thanks for the review @ddebroy 

Agree that the annotation can be made from the PVC or the storage class. I found that embedding the OS specifics as part of the PVC to be more logical. The user wants a claim specific to Windows workloads; she/he adds the "os: windows" annotation. The storage class remains OS-agnostic. The claim is used in the workload and it seems like the more natural object to hold this information.

pjh:
Should the PR title be updated to something like "Add Windows support for iSCSI FlexVol"? I had to read a couple paragraphs into the PR description before I realized that iSCSI FlexVolume support already exists on Linux but not on Windows.

anusha-ragunathan:
Thanks for the review, @pjh. iSCSI FlexVolume support does not exist for Linux. Its not needed since Kubernetes provides in-tree support for what we are trying to achieve using FlexVols here.

iscsi/targetd/kubernetes/iscsi-provisioner-pvc-win.yaml
```diff
@@ -0,0 +1,14 @@
+kind: PersistentVolumeClaim
+apiVersion: v1
+metadata:
+  name: myclaim
+  annotations:
+    volume.beta.kubernetes.io/storage-class: "iscsi-targetd-vg-targetd"
```
jingxu97: why put storageclass inside annotations?

iscsi/targetd/provisioner/iscsi-provisioner.go
```diff
@@ -108,6 +108,40 @@ func (p *iscsiProvisioner) getAccessModes() []v1.PersistentVolumeAccessMode {
...
+		return v1.PersistentVolumeSource{
+			FlexVolume: &v1.FlexPersistentVolumeSource{
+				Driver: "microsoft.com/iscsi.cmd",
```
msau42: I'm slightly wary of putting a dependency on a vendor-specific driver in this package that is supposed to be vendor-neutral.  Would it be better to put this under the same repo where the flex driver is located at?

wongma7: yes, feel free to fork. The targetd provisioner doesn't have an active maintainer so if you want to repurpose it to support also iscsi on windows workers, & have free reign to make changes to it, we can link to the fork from here.

ddebroy: Sounds good. @KnicKnic will it make sense to host the Linux targetd provisioner fork (with the above PR) in https://github.com/microsoft/ ?

anusha-ragunathan: Instead of forking, how about make the flexvol plugin `Driver` field configurable from the storage class? Simpler.

iscsi/targetd/kubernetes/iscsi-provisioner-pvc-win.yaml
anusha-ragunathan: Thanks for your review @jingxu97 
The pattern is from https://github.com/kubernetes-incubator/external-storage/blob/master/iscsi/targetd/kubernetes/iscsi-provisioner-pvc.yaml. It could be from a sample from an earlier version of k8s. Let me fix that.

iscsi/targetd/provisioner/iscsi-provisioner.go
SubodhBhargava: > Sounds good. @KnicKnic will it make sense to host the Linux targetd provisioner fork (with the above PR) in https://github.com/microsoft/ ?

@ddebroy - I work with @KnicKnic. That sounds good; we can host it at microsoft/k8s-storage-plugins

ddebroy: @anusha-ragunathan the Driver field may be made configurable but the [options](https://github.com/kubernetes-incubator/external-storage/pull/1226/files#diff-d5f705cb57d27aa7dee80b63d3cb25efR117) keys are pretty tightly coupled with the implementation of the flexvolume driver.

iscsi/targetd/provisioner/iscsi-provisioner.go
jsafrane: If the cluster is heterogenous (i.e. Windows and Linux nodes mixed together), I would expect that WaitForFirstConsumer is used and the provisioner creates a PV that can be used from the node where corresponding pod is scheduled (i.e. the provisioner checks Node `kubernetes.io/os` label.

With Immediate mode, the provisioner can use a config option or StorageClass parameter to determine what to provision. Or simply roll a dice. Annotations in PVC make the PVC dependent on the platform and that's not what we want.

In all cases (both Immediate & WaitForFirstConsumer provisioning), the provisioned PV should have node affinity that selects nodes with the right OS (e.g. `kubernetes.io/os: linux`)

jingxu97:
@anusha-ragunathan are you still working on this PR?

anusha-ragunathan:
@jingxu97 : Thanks for the ping. I havent had the time to work on this. I'll go ahead and close it.

Issue #1243, opened 2019-10-24
Gluster Block Provisioner is incompatible with the current Ansible Playbooks for OKD 3.11


cgruver:
Changes were made to the Ansible Role for installing Gluster in OKD 3.11 and presumably OCP 3.11, that enabled the deployment of multiple Gluster provisioners.  The provisioner name is appended with the namespace that it is deployed in.  However, the current code in glusterblock-provision.go is not properly leveraging the environment variable that is set in the deployment descriptor that Ansible creates.

There also seems to be a redundant call to klog.InitFlags() that causes the app to panic when it starts.

I am submitting a pull request with a fix.  The code can be found at https://github.com/cgruver/external-storage/blob/master/gluster/block/cmd/glusterblock-provisioner/glusterblock-provisioner.go

This issue was also reported at [Can't bind PVC to StorageClass using gluster block
](https://github.com/openshift/origin/issues/23881)

PR #1244, opened 2019-10-24
Correct the handling of environment variable that sets the Provisioner Name for Gluster Block Provisioning

cgruver:
This fix is to correct the issue reported in #1243

[Gluster Block Provisioner is incompatible with the current Ansible Playbooks for OKD 3.11](https://github.com/kubernetes-incubator/external-storage/issues/1243)

I replaced the declared constant for provisionerName with a variable, which takes the value of an environment variable if it is set.  Otherwise it defaults to the value that was set in the constant which I replaced.

I am not an experienced GO developer, so please review to make sure I haven't done something silly.

This fix has been tested in a newly built OKD 3.11 cluster using the playbooks provided by the openshift-ansible project.


cgruver:
Appologies, the last commit was made from host that did not have my git config set up...  I will try to correct the author info in that last commit.

PR #1245, opened 2019-10-24
This change is to correct issue #1243.

cgruver:
This fix is to correct the issue reported in #1243

Gluster Block Provisioner is incompatible with the current Ansible Playbooks for OKD 3.11

I replaced the declared constant for provisionerName with a variable, which takes the value of an environment variable if it is set. Otherwise it defaults to the value that was set in the constant which I replaced.

I am not an experienced GO developer, so please review to make sure I haven't done something silly.

This fix has been tested in a newly built OKD 3.11 cluster using the playbooks provided by the openshift-ansible project.


cgruver:
CLA completed.

gluster/block/cmd/glusterblock-provisioner/glusterblock-provisioner.go
```diff
@@ -861,13 +860,14 @@ func GetSecretForPV(restSecretNamespace, restSecretName, volumePluginName string
...
-	id         = flag.String("id", "", "Unique provisioner identity")
+	master          = flag.String("master", "", "Master URL")
+	kubeconfig      = flag.String("kubeconfig", "", "Absolute path to the kubeconfig")
+	id              = flag.String("id", "", "Unique provisioner identity")
+	provisionerName = "gluster.org/glusterblock"
 )
 
 func main() {
-	klog.InitFlags(nil)
+	//klog.InitFlags(nil)
```
wongma7: this looks suspect. Was it crashing until you removed this line?

cgruver: Yes.  The app would crash out on startup.

I'm not experienced with GO, but did some digging on the generated error and the app behavior.  What I found seemed to indicate that klog.InitFlags, or something similar was already being called higher up in the bootstrap process for the application.  So, I removed it.  Looking back through the code, it looked like klog had recently been added to the app as a replacement for glog.

wongma7: OK, thanks, you're right to remove it. glog->klog migration happened in this git repo and I was not paying attention, unfortunate that it broke.

PR #1252, opened 2019-11-07
nfs-client add parameters to control target dirs.

iineva:
For some reason we need target dir static.
This PR is implement this feature.

### DIFF
--- a/nfs-client/README.md
+++ b/nfs-client/README.md
@@ -111,6 +111,7 @@ provisioner: fuseim.pri/ifs # or choose another name, must match deployment's en
 parameters:
   archiveOnDelete: "false" # When set to "false" your PVs will not be archived
                            # by the provisioner upon deletion of the PVC.
+  archiveDoNotMove: "false" # When set to "true" and archiveOnDelete=true archived will keep directory name.
 ```
 
 **Step 5: Finally, test your environment!**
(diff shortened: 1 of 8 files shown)


wongma7:
Nice, I think something like this has been requested a number of times and I've just told people to create PVs manually. But this is an acceptable middle-ground, it saves them time...

Awaiting CLA

iineva:
> Nice, I think something like this has been requested a number of times and I've just told people to create PVs manually. But this is an acceptable middle-ground, it saves them time...
> 
> Awaiting CLA

You can use this image`ineva/nfs-client-provisioner:latest`  to  replace `quay.io/external_storage/nfs-client-provisioner` to user this features.

iineva:
CLA done!

yashbhutwala:
ping on status of this 😄

wongma7:
I think archiveDoNotMove makes things quite confusing. archiveOnDelete was already a bit confusing because "archiving" a directory instead of actually deleting it is non-standard behavior. Basically there are 2 flags and 3 possibilities on PVC deletion:
1. Delete directory archiveOnDelete false
2. "Archive" directory archiveOnDelete true and archiveDoNotMove false
3. Keep directory archiveOnDelete true and archiveDoNotMove true

Could we introduce a new parameter that makes this clear and deprecate archiveOnDelete? Like deletionBehavior that accepts (delete, archive, keep)?

PR #1263, opened 2019-12-09
fix on openshift  Permission denied

helingjun:


cofyc:
I suggest you use [ceph-csi](https://github.com/ceph/ceph-csi) instead.

helingjun:
add cla

humblec:
@helingjun as @cofyc suggested its better to use csi. We are no longer planning to maintain these provisioners so the suggestion.

helingjun:
Think you

helingjun:
I signed it

PR #1285, opened 2020-02-20
Fix cmake version for CHECK_SYMBOL_EXISTS

wongma7:


wongma7:
merge to fix travis CI nfs e2e tests failing because ganesha won't compile with new cmake

PR #1306, opened 2020-04-21
added rbd install instructions and ceph admin secret consistency

### LINKED ISSUE #1256: CEPH RBD Provisioner Creates PV that fails to attach
I have setup a CEPH cluster. Independent of that I have followed processes found online to setup RBD provisioning.

The process I've followed is found here:
https://medium.com/velotio-perspectives/an-innovators-guide-to-kubernetes-storage-using-ceph-a4b919f4e469

This has largely worked. I have a "fast-rbd" storage class as follows:
image

I've created the various resources required for provisioning:
image

I've created the secrets to access ceph properly.
image

I am able to create PVC's that successfully bind to the PV created by the provisioner

image

However, the pod that I create fails to attach to the volume:
image

Looking at the pod in more detail I see the following
image

Googling the warning: 

> fail to check rbd image status with (executable file not found in $path)

I find various hits online including this [one](https://github.com/kubernetes-incubator/external-storage/issues/1002)

It would seem that others have struggled with this. I dont fully understand the resolution described here. I am looking for guidance relative to getting past this problem

I believe the problem is that the running container does not include "rbd" in its path. Its not clear to me how this is properly resolved.

Guidance/Advice would be appreciated.
Dave

kifeo:
- added rbd install instructions to fix #1256
- added consistency between cephfs and rbd to use the same ceph admin secret

### DIFF
--- a/ceph/rbd/README.md
+++ b/ceph/rbd/README.md
@@ -32,7 +32,7 @@ See https://kubernetes.io/.
 
 ```bash
 ceph auth get client.admin 2>&1 |grep "key = " |awk '{print  $3'} |xargs echo -n > /tmp/key
-kubectl create secret generic ceph-admin-secret --from-file=/tmp/key --namespace=kube-system --type=kubernetes.io/rbd
+kubectl create secret generic ceph-secret-admin --from-file=/tmp/key --namespace=kube-system --type=kubernetes.io/rbd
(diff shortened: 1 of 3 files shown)


cofyc:
/lgtm
/approve
Thanks!

Issue #1275, opened 2020-01-28
NFS-Client Provisioner work across all namespaces ?


yacota:
I came across this issue https://github.com/kubernetes-incubator/external-storage/issues/1210 and was wondering whether this could be possible with nfs-client-provisioner

I mean, is it possible to have a **single** namespace with one pod/provisioner per each nfs mount point in use and just declare a persistent volume claim in another namespace that use these storageclassnames ?

For the time being I have 2 namespaces that are mounting the same nfs export, both are working fine (both use 'default' serviceAccount in each namespace), but I'd like to share storargeclassname/provisioner between them(is this possible?)

StorageClasses
```
$  kubectl get storageclass
AME                       PROVISIONER                    AGE
gp2 (default)             kubernetes.io/aws-ebs          362d
ssditto-nfs-live-client   ssditto-nfs-live-provisioner   3h35m
xfiles-nfs-live-client    xfiles-nfs-live-provisioner    170m
```

Pods per namespace xfiles
```
$  kubectl get pods -n xfiles
NAME                                                  READY   STATUS    RESTARTS   AGE
xfiles-669c7bd646-jlhbk                               1/1     Running   0          166m
xfiles-nfs-client-provisioner-live-78c9f59fd5-9jb8z   1/1     Running   0          137m
```

Pods per namespace ssditto
```
$  kubectl get pods -n ssditto
NAME                                                  READY   STATUS    RESTARTS   AGE
sditto-69dc99bb7-6fdfx                                1/1     Running   0          26m
sditto-nfs-client-provisioner-live-85484fdddb-rzjlb   1/1     Running   0          3h30m
```

Persitent volumes
```
$ kubectl get pv
NAME                                       CAPACITY   ACCESS MODES   RECLAIM POLICY   STATUS   CLAIM                                             STORAGECLASS   REASON   AGE
pv-ssditto-nfs-client-provisioner-live     10Mi       RWO            Delete           Bound    ssditto/pvc-ssditto-nfs-client-provisioner-live                           3h29m
pv-xfiles-nfs-client-provisioner-live      10Mi       RWO            Delete           Bound    xfiles/pvc-xfiles-nfs-client-provisioner-live                             165m
```

Persistent volume claim for xfiles
```
$ kubectl get pvc -n xfiles
NAME                                     STATUS   VOLUME                                  CAPACITY   ACCESS MODES   STORAGECLASS   AGE
[...]

Issue #1210, opened 2019-08-16
Is there a way to have efs-provisioner work across all namespaces?


richstokes:
Seems that it only works if you create pods in the same namespace as the pvc. Is there a way that other name spaces can share the same EFS server?  

Looks like you can do it if you create a new PVC foreach namespace with `storageClassName: "aws-efs"` set, but was wondering if there is a simpler way that will allow all namespaces by default?

Thanks

martin2176:
I have not attempted it, but standard rbac should allow you use any resource across namespaces.
In your case, the service account which runs the pod (default service account in most cases) should be given admin role in the namespace which has pvc. via a rolebinding.
If you dont want to give admin role, a more restrictive privilege can be given .

therc:
> I know that generally PVCs are namespace resources and NS2 should not have access to the PVC from NS1. But there need to be a way to use the same storage folder from different Namespaces. Maybe by creating a PVC in each Namespace but with an uniq identifier so that the `nfs-client-provisioner` could match these PVC and they use the same folder.

> In my case i use the "nfs-client-provisioner" not the "aws-efs" but the idea should be the same.

I need something similar. I think there's a way to do this with a few small changes and you might not even need an unique identifier.

You create PVC1 in NS1 as before. If you're on AWS, you use efs-provisioner for that, presumably. Then you run nfs-client-provisioner in NS1, mounting PVC1.

It doesn't do it at the moment, but nfs-client-provisioner can find out the server and the path for PVC1, by looking at /proc/mounts.

These would be the changes needed:
1. allow NFS_SERVER to be set to "auto" (or a new AUTO_NFS_SERVER=true)
1. allow NFS_PATH to be set to "clone" (or a new CLONE_NFS_PATH=true)

In the former case, the server field will be populated automatically with the value discovered from /proc/mounts.

In the latter, the provisioner no longer creates a new directory on the server, but reuses the path from /proc/mount. It should also **set the reclaim policy to Retain**. (And it could propagate whether the mount is R/W or R/O!)

I don't think efs-provisioner will delete your EFS server once the last reference to it goes away, but it will remove recursively the PVC1 directory tree if the claim gets deleted and `retainPolicy == Delete|Recycle`, so your clones might see data disappear if you're not careful.

The above shouldn't require a lot of code, which I can help with. The real work is mostly a matter of agreeing on how to trigger this mode, how to test it and what kind of guarantees to offer (or not to offer).

therc:
[...]

PR #1318, opened 2020-05-10
Prototype for nfs-client alias mode

therc:
This might be a solution for #1210 and #1275
 
In combination with an existing NFS claim in namespace X, allow the administrator or the users to create new claims for the same NFS tree in namespaces A, B, C, etc.

This change tries as much as possible NOT to disrupt existing setups.

A few things still left:

1. Is "alias" the right term? Previous choices I tossed out: clone (easy to mix with real volume cloning), proxy (there's no middle-man NFS server), passthrough (too nebulous).
1. Is NFS_SERVER=--alias the best way to trigger this?
1. Should the auto-detection of the server host:path happen always, even when not in alias mode?
1. Should the parent's ReadOnly field be propagated to the newly provisioned PV?
1. Should this be made more generic so that one deployment can expose N different claims/trees to the other namespaces? See comment below for more details.

### DIFF
--- a/nfs-client/README.md
+++ b/nfs-client/README.md
@@ -22,7 +22,7 @@ $ helm install stable/nfs-client-provisioner --set nfs.server=x.x.x.x --set nfs.
 
 **Step 1: Get connection information for your NFS server**. Make sure your NFS server is accessible from your Kubernetes cluster and get the information you need to connect to it. At a minimum you will need its hostname.
 
-**Step 2: Get the NFS-Client Provisioner files**. To setup the provisioner you will download a set of YAML files, edit them to add your NFS server's connection information and then apply each with the ``kubectl`` / ``oc`` command. 
+**Step 2: Get the NFS-Client Provisioner files**. To setup the provisioner you will download a set of YAML files, edit them to add your NFS server's connection information and then apply each with the ``kubectl`` / ``oc`` command.
 
 Get all of the files in the [deploy](https://github.com/kubernetes-incubator/external-storage/tree/master/nfs-client/deploy) directory of this repository. These instructions assume that you have cloned the [external-storage](https://github.com/kubernetes-incubator/external-storage) repository and have a bash-shell open in the ``nfs-client`` directory.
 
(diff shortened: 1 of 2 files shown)


therc:
"Clone mode" is very confusing, I think "passthrough" is clearer and more descriptive.

Naming aside, this could be made more generic so that you don't need to run a million clone provisioners if you have more than one NFS tree to make available to your namespaces.

Let's say you run the nfs-client deployment for provisioner `foo.com/shared` bound to three volume claims that I want to expose to any number of other namespaces: data, code, docs. They're mounted under `/persistentvolumes/`. There is a storage class `shared-volumes` associated with the provisioner. Now I create claim `code` in namespace `developer1` under that class. The provisioner will lookup the server:path for `/persistentvolume/code` and stick those in a new volume `developer1-code-BLABLABLA`.

If others agree that the generic, multi-claim approach is better, I could change the code so that the mode activates when NFS_SERVER is set to `-passthrough` (hostnames can't start with a dash, so there's no collision with a real host). Then NFS_PATH would be the directory under which to look for mounted claims. In the example above, it would be set to `/persistentvolume`.

therc:
Ping?

kmova:
@therc - We are in the middle of migrating the NFS client code into a new home under  https://github.com/kubernetes-sigs/nfs-subdir-external-provisioner. The migration PR is up for review at the moment. https://github.com/kubernetes-sigs/nfs-subdir-external-provisioner/pull/2

Sorry to have kept mum on this PR for a while, can I circle back with you once the above migration PR is merged and take your help to shift these changes into the new repo?

therc:
Sounds good, @kmova... thank you!

nikhita:
As @kmova mentioned, we are in the process of archiving this repo and continuing work in https://github.com/kubernetes-sigs/nfs-subdir-external-provisioner. Please reopen a PR in the new repo once https://github.com/kubernetes-sigs/nfs-subdir-external-provisioner/pull/2 has merged. Thanks for your patience! :) 

Going to close this PR in order to be able to archive this repo.

PR #1333, opened 2020-06-16
fix bug: gluster file params does not take effect

lowang-bh:
It always use the default volumetype due to the parsed config is not used

### DIFF
--- a/ceph/cephfs/cephfs-provisioner.go
+++ b/ceph/cephfs/cephfs-provisioner.go
@@ -45,6 +45,7 @@ const (
 	provisionCmd       = "/usr/local/bin/cephfs_provisioner"
 	provisionerIDAnn   = "cephFSProvisionerIdentity"
 	cephShareAnn       = "cephShare"
+	cephVolumeGroup    = "CEPH_VOLUME_GROUP"
 	provisionerNameKey = "PROVISIONER_NAME"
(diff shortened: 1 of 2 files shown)


lowang-bh:
I signed it

lowang-bh:
I signed it

humblec:
@lowang-bh thanks for the PR. One thing to note here is, this repo is under deprecation, so may be getting this PR in may not add much value. The PR looks good to me though

lowang-bh:
> @lowang-bh thanks for the PR. One thing to note here is, this repo is under deprecation, so may be getting this PR in may not add much value. The PR looks good to me though

Yes, it is to be deprecate. But there is  also someone using this external-storage, so I think it is useful to fix the issue so that others can use it directly. So can you approve it?

nikhita:
@lowang-bh Thanks for your PR!

We are in the process of archiving this repo and are no longer accepting new PRs. Please see https://github.com/kubernetes/org/issues/1563 for more details. 

> But there is also someone using this external-storage, so I think it is useful to fix the issue so that others can use it directly. 

If you need specific code to be still maintained, please reach out to sig-storage on their mailing list. We've done that for certain code paths like https://github.com/kubernetes/org/issues/1743 and https://github.com/kubernetes/org/issues/1744

PR #1334, opened 2020-06-20
[MOD]update to new volume controller api.

ThePolarNight:
We find that cephfs use old volume api struct.


lowang-bh:
not only cephfs, all volume controller used old api;
you change the api but vendor is the old one. should upate all also the vendor

ThePolarNight:
I use CSI plugin Ceph-CSI(https://github.com/ceph/ceph-csi) instead.

PR #1336, opened 2020-06-26
support china efs dns names

stoehdoi:
**EFS in now GA in AWS China.** However, EFS DNS names in respective regions end with `.cn` which is currently not supported. This PR adds support for china AWS regions.


nikhita:
Thanks for your PR!

We are in the process of archiving this repo and are no longer accepting new PRs. Please see https://github.com/kubernetes/org/issues/1563 for more details. Thanks again!

Issue #1345, opened 2020-07-20
nfs-provisioner vs ganesha fsid=152.152 reserved for internal purposes 


rafal-prasal:
ganesha reserved exportID 152 for internal purposes. when generating id for filesystem nfs-provisioner should avoid it.

https://github.com/nfs-ganesha/nfs-ganesha/issues/615

of course having 152 volume claims is an outstanding achievement, but we somehow reached that.

rafal-prasal:
this has to be added to initial setup of ganesha

```
EXPORT {
    Export_Id=65536;
    Path="/";
    Pseudo="/";
    Access=MDONLY_RO;
    Protocols=4;
    SecType=none,sys,krb5,krb5i,krb5p; # actually just has to list all the SecTypes used by your exports
    Filesystem_Id=65536.65536; # some value that doesn't conflict
    FSAL {
        Name=PSEUDO;
    }
}
```

ffilz:
Actually nfs-ganesha doesn't reserve export_id 152, but rather, the provisioner is specifying Filesystem_Id as export_id:export_id and nfs-ganesha uses Filesystem_Id 152:152 for the NFS v4 Pseudo Filesystem.

When using FSAL_VFS you should not need to specify Filesystem_Id for exports unless exporting an underlying filesystem that does not have something stable nfs-ganesha can use. nfs-ganesha is capable of using the UUID off the filesystem or the device ID the filesystem is mounted on (which may not be stable, I'm not sure if that's actually a problem for Kubernetes though).

nikhita:
Thanks for reporting the issue!

This repo is no longer being maintained and we are in the process of archiving this repo. Please see https://github.com/kubernetes/org/issues/1563 for more details.

If your issue relates to nfs provisioners, please create a new issue in https://github.com/kubernetes-sigs/nfs-ganesha-server-and-external-provisioner or https://github.com/kubernetes-sigs/nfs-subdir-external-provisioner.

Going to close this issue in order to archive this repo. Apologies for the churn and thanks for your patience! :pray:

PR #1346, opened 2020-07-20
#1345 nfs-provisioner: reserving id 152 because it is internally used by ganesha

rafal-prasal:
reservation for newly generated export ids


rafal-prasal:
it can be handled in a better, less "internal" way

